PR #6, opened 2015-08-30
building the initial bundle

dbu:
fix #1

feedback highly welcome @sagikazarmark @egeloen

my idea is to get this running when i return from holidays in 2 weeks and then contribute the bundle to the php-http organization. then we can work on porting more features from ivory into the new structure.

PR #8, opened 2015-10-17
building the initial bundle

### LINKED ISSUE #7: Httplug
Shouldn't this be renamed to httplug? Also, I think you could move it to the organization.

dbu:
continued from #6
fix #7

once merged, we can move this repo to the php-http organisation.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace Http\HttplugBundle\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+
+/**
+ * This class contains the configuration information for the bundle
+ *
+ * This information is solely responsible for how the different configuration
+ * sections are normalized, and merged.
+ *
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class Configuration implements ConfigurationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder();
+        $rootNode = $treeBuilder->root('httplug');
+
+        $rootNode
+            ->validate()
+                ->ifTrue(function ($v) {
+                    return !empty($v['classes']['client'])
+                        || !empty($v['classes']['message_factory'])
+                        || !empty($v['classes']['uri_factory'])
+                    ;
+                })
+                ->then(function ($v) {
+                    foreach ($v['classes'] as $key => $class) {
+                        if (null !== $class && !class_exists($class)) {
+                            throw new InvalidConfigurationException(sprintf(
+                                'Class %s specified for httplug.classes.%s does not exist.',
+                                $class,
+                                $key
+                            ));
+                        }
+                    }
+
+                    return $v;
+                })
+            ->end()
+            ->children()
+                ->arrayNode('main_alias')
+                    ->addDefaultsIfNotSet()
+                    ->info('Configure which service the main alias point to.')
+                    ->children()
+                        ->scalarNode('client')->defaultValue('httplug.client.default')->end()
+                        ->scalarNode('message_factory')->defaultValue('httplug.message_factory.default')->end()
+                        ->scalarNode('uri_factory')->defaultValue('httplug.uri_factory.default')->end()
+                    ->end()
+                ->end()
+                ->arrayNode('classes')
+                    ->addDefaultsIfNotSet()
+                    ->info('Overwrite a service class instead of using the discovery mechanism.')
+                    ->children()
+                        ->scalarNode('client')->defaultNull()->end()
+                        ->scalarNode('message_factory')->defaultNull()->end()
+                        ->scalarNode('uri_factory')->defaultNull()->end()
+                    ->end()
+                ->end()
+            ->end()
+        ;
+
+        return $treeBuilder;
+    }
+}
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Http\HttplugBundle\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+
+/**
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class HttplugExtension extends Extension
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function load(array $configs, ContainerBuilder $container)
+    {
+        $configuration = $this->getConfiguration($configs, $container);
+        $config = $this->processConfiguration($configuration, $configs);
+
+        $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+
+        $loader->load('discovery.xml');
+        foreach ($config['classes'] as $service => $class) {
+            if (!empty($class)) {
+                $container->removeDefinition(sprintf('httplug.%s.default', $service));
+                $container->register(sprintf('httplug.%s.default', $service), $class);
+            }
+        }
+
+        foreach ($config['main_alias'] as $type => $id) {
+            $container->setAlias(sprintf('httplug.%s', $type), $id);
+        }
+    }
+}
--- a/HttplugBundle.php
+++ b/HttplugBundle.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace Http\HttplugBundle;
+
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+/**
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class HttplugBundle extends Bundle
+{
+}
--- a/README.md
+++ b/README.md
@@ -20,9 +20,54 @@ Via Composer
 $ composer require php-http/httplug-bundle
 ```
 
+Enable the bundle in your kernel:
+ 
+``` php
+<?php
+// app/AppKernel.php
+
+public function registerBundles()
+{
+    $bundles = array(
+        // ...
+        new Http\HttplugBundle\HttplugBundle(),
+    );
+}
+```
 
 ## Usage
 
+The usage documentation is split into two parts. First we explain how to configure the bundle in an application. The second part is for developing reusable Symfony bundles that depend on an HTTP client defined by the Httplug interface.
+
+### Use in applications
+
+This bundle provides 3 services: 
+
+* `httplug.client` a service that provides the `Http\Client\HttpClient`
+* `httplug.message_factory` a service that provides the `Http\Message\MessageFactory`
+* `httplug.uri_factory` a service that provides the `Http\Message\UriFactory`
+
+These services are always an alias to another service. You can specify your own service or leave the default, which is the same name with `.default` appended. The default services in turn use the service discovery mechanism to provide the best available implementation. You can specify a class for each of the default services to use instead of discovery, as long as those classes can be instantiated without arguments.
+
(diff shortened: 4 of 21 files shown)


dbu:
@sagikazarmark can you review again? i think i adressed all your concerns raised in #6. also doing the rename. what should the main namespace be? currently its Http/HttplugBundle. is that ok?

composer.json
```diff
@@ -0,0 +1,36 @@
...
+    "authors": [
+        {
+            "name": "David Buchmann",
+            "email": "mail@davidbu.ch"
+        }
+    ],
+    "require": {
+        "php-http/discovery": "^0.1.1",
+        "php-http/httplug": "^0.1.0",
+        "php-http/httplug-implementation": "^0.1.0",
```
dbu: are these namespaces correct? or what will the virtual package be?

sagikazarmark:
I see `Vendor\Bundle\NamedBundle` regularly. Is it a common pattern? If so , what do you think about it?

Regarding naming: I trust in your choice as you have more experience in it.

composer.json
sagikazarmark: We discussed naming in php-http/httplug#68. I would prefer those two, but I am open to suggestions.

DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+namespace Http\ClientBundle\DependencyInjection;
```
sagikazarmark: Isn't this HttplugBundle?

DependencyInjection/HttplugExtension.php
```diff
@@ -0,0 +1,37 @@
...
+    public function load(array $configs, ContainerBuilder $container)
+    {
+        $configuration = $this->getConfiguration($configs, $container);
+        $config = $this->processConfiguration($configuration, $configs);
+
+        $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
+
+        $loader->load('discovery.xml');
+        foreach ($config['classes'] as $service => $class) {
+            if ($class) {
```
sagikazarmark: What do you check here? Maybe a more explicit check (isset, empty, value check)?

HttplugBundle.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Http\ClientBundle;
+
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+/**
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class HttplugBundle extends Bundle
+{
```
sagikazarmark: Not sure what is the PSR2 standard here: empty line between curly braces necessary or not? @hannesvdvreken you corrected me earlier.

README.md
```diff
@@ -20,9 +20,54 @@ Via Composer
...
 
 ## Usage
 
+The usage documentation is split into two parts. First we explain how to configure the bundle in an application. The second part is for developing reusable Symfony bundles that depend on an HTTP client defined by the Httplug interface.
+
+### Use in applications
+
+This bundle provides 3 services: 
+
+* `httplug.client` a service that provides the `Http\Adapter\HttpAdapter` (TODO: proper interface?)
```
sagikazarmark: Wrong interface?

HttplugBundle.php
hannesvdvreken: No empty line.

Resources/config/discovery.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="httplug.client.default"
+             class="Http\Adapter\HttpAdapter">
```
sagikazarmark: Wrong interface?

dbu:
> I see Vendor\Bundle\NamedBundle regularly. Is it a common pattern? If so
, what do you think about it?

the Bundle bit is optional. what is our vendor name? Http? or PhpHttp? i
looked at the httplug code and there its still Http/Client so i went with
that. do we keep the httplug namespace like that?

sagikazarmark:
Okay, we can skip the Bundle in the namepace.

Honestly, I am not sure about naming. We kept Http\Client for the httplug package. Not sure about that either, but I think the bundle could be HttplugBundle regardless of what the package namespace is.

composer.json
dbu: but depending on php-http/psr-client-implementation should be enough,
right? if people need a client that has the convenience methods, they
can just add that to their dependencies and configure it, but psr-client
is kind of the minimum...

DependencyInjection/HttplugExtension.php
dbu: its the cheapest way to know if the string is non-empty. but i change to
!empty() to make it more explicit.

On 17.10.2015 21:12, Márk Sági-Kazár wrote:
> In DependencyInjection/HttplugExtension.php
> <https://github.com/dbu/HttplugBundle/pull/8#discussion_r42313006>:
> 
>> + */
>> +class HttplugExtension extends Extension
>> +{
>> +    /**
>> +     * {@inheritdoc}
>> +     */
>> +    public function load(array $configs, ContainerBuilder $container)
>> +    {
>> +        $configuration = $this->getConfiguration($configs, $container);
>> +        $config = $this->processConfiguration($configuration, $configs);
>> +
>> +        $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
>> +
>> +        $loader->load('discovery.xml');
>> +        foreach ($config['classes'] as $service => $class) {
>> +            if ($class) {
> 
> What do you check here? Maybe a more explicit check (isset, empty, value
> check)?
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dbu/HttplugBundle/pull/8/files#r42313006>.
> 

-- 
Liip AG // Agile Web Development // T +41 43 500 39 80
CH-8005 Zurich // PGP 0xA581808B // www.liip.ch

README.md
```diff
@@ -20,9 +20,54 @@ Via Composer
...
 
 ## Usage
 
+The usage documentation is split into two parts. First we explain how to configure the bundle in an application. The second part is for developing reusable Symfony bundles that depend on an HTTP client defined by the Httplug interface.
+
+### Use in applications
+
+This bundle provides 3 services: 
+
+* `httplug.client` a service that provides the `Http\Client\HttpClient`
```
dbu: depending on the decision on https://github.com/php-http/httplug/issues/68#issuecomment-149112039 this might need to be changed

Resources/config/discovery.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="httplug.client.default"
+             class="Http\Client\HttpClient">
```
dbu: depending on the decision on https://github.com/php-http/httplug/issues/68#issuecomment-149112039 this might need to be changed

composer.json
dbu: i have to wait for our decision on https://github.com/php-http/httplug/issues/68#issuecomment-149112039 until i can adapt this

Resources/config/discovery.xml
```diff
@@ -0,0 +1,21 @@
...
+            <factory class="Http\Discovery\HttpAdapterDiscovery" method="find"/>
```
dbu: this might change depending on https://github.com/php-http/discovery/issues/11

dbu:
updated. will have to check whether i can install it and all that, then would be ready to transfer to the php-http organisation.

also need to see about tests that make sure we are in sync with all parts of php-http. what is the state of the guzzle adapter?

README.md
dbu: and fine as well.

Resources/config/discovery.xml
dbu: ok, this is fine.

sagikazarmark:
I think guzzle implements the latest-greatest httplug, thanks  to @joelwurtz.

Tests/Resources/Fixtures/config/full.xml
```diff
@@ -0,0 +1,18 @@
...
+<container xmlns="http://symfony.com/schema/dic/services">
+
+    <config xmlns="http://example.org/schema/dic/httplug">
+        <main-alias>
+            <client>my_client</client>
+            <message-factory>my_message_factory</message-factory>
+            <uri-factory>my_uri_factory</uri-factory>
+        </main-alias>
+        <classes>
+            <client>Http\Adapter\Guzzle6HttpAdapter</client>
```
sagikazarmark: Wrong name

Tests/Resources/Fixtures/config/full.yml
```diff
@@ -0,0 +1,9 @@
+httplug:
+    main_alias:
+        client: my_client
+        message_factory: my_message_factory
+        uri_factory: my_uri_factory
+    classes:
+        client: Http\Adapter\Guzzle6HttpAdapter
```
sagikazarmark: Wrong name

sagikazarmark:
Will it be possible to configure multiple clients? Also, can we configure Guzzle Client for example prior to passing it to the adapter? (I guess it has to be done in services somehow)

dbu:
> Will it be possible to configure multiple clients?

this is certainly something we will want to make possible, but right now
the bundle does not help with that.

> Also, can we
> configure Guzzle Client for example prior to passing it to the adapter?
> (I guess it has to be done in services somehow)

the bundle does not know about specific adapters. the idea for that use
case would be that you simply configure guzzle and the adapter as
services in your application and tell the bundle that this service name
is the default.

with httplug being just the contract, the bundle can not really help
with setting up specific clients. where it will really shine however is
when we have the plugin system in place. then we can allow to configure
pipelines.

i also plan to add services for the utils clients (Batch, Methods). but
not in this PR, first get the MVP done.

sagikazarmark:
> that this service name is the default.

Like that, I was thinking about the same.

> the bundle can not really help with setting up specific clients

Actually, I think that specific clients should be configured through their own bundle and then the configured service should be passed to the adapter.

> i also plan to add services for the utils clients (Batch, Methods)

:+1:

README.md
```diff
@@ -20,9 +20,54 @@ Via Composer
...
+If you need to customize the service with decorators, e.g. to add authentication headers, decorate the default service and configure the main alias to your decorating service name.
+
+```yaml
+httplug:
+    main_alias:
+        client: httplug.client.default
+        message_factory: httplug.message_factory.default
+        uri_factory: httplug.uri_factory.default
+    classes:
+        client: ~ # uses discovery if not specified
```
joelwurtz: Hum, maybe we can have a class for each adapter with can inject/send a NodeConfiguration ? So we would now how to configure each adapter.

sagikazarmark: I think it should be done in service configuration by the user.

dbu: the user can specify a class for the client / adapter and we just
instantiate that class without arguments. in a first step, i would
expect the user to define a service for their adapter if they need
specific configuration on their client/adapter.

the next step will be to allow to configure plugins and wrapper clients.
i want to avoid configuring the actual guzzle *client*, but if the
guzzle *adapter* has specific configuration, we can add that at some
point. not sure how much specific configuration the adapters will have
however.

sagikazarmark: > not sure how much specific configuration the adapters will have however.

Shouldn't be any at all.

Is it possible to specify adapter classes as services, so we can configure the underlying client?

For example create a service for Guzzle6Adapter where a Guzzle Client instance is being injected which can be cpnfigured as a service as well.

dbu: > Is it possible to specify adapter classes as services

that seems too complicated to me. either use the factory or simply define a service for the underlying client and another one for the adapter, then configure this bundle to use your service.

i updated the readme to better explain the idea.

sagikazarmark: > simply define a service for the underlying client and another one for the adapter, then configure this bundle to use your service.

That's exactly what I want.

dbu:
i think this is ready for an initial version. note that the bundle does not do much more than providing the autowiring as a symfony service - but with the option to specify a manually created service instead. once plugins are defined, we should work on that. and also provide convenient ways to get the HttpMethodsClient, BatchClient and AsyncClient as services.

any objections if i merge this and move it to the php-http organisation?

sagikazarmark:
Nope, go ahead.

Issue #10, opened 2015-12-24
The scope of this bundle?


Nyholm:
I'm curious what your intention of this bundle is. I know it has been a while since this was updated and your focus are on the libraries. At the moment this bundle creates services for the client and factories. It means that it is only good for dependency injection. Is it any other wins atm?

I had an idea of actually be implementation specific here. The bundle config could allow you to configure clients and make sure to create services for a **configured** client. Example:

```yaml
httplug:
  clients:
    acme: 
      adapter: guzzle6
      timeout: 2
      verify_ssl: false

services:
  linkedin:
    class: Happyr\LinkedIn\LinkedIn
    arguments: ["@httplug.acme"]
```

What do you think? Would you accept a pr implementing this?

dbu:
right now the bundle is indeed minimal. i really like your proposal, a PR
is very  welcome.
the other important next step will be to support configuring plugins and eg
handle the symfony debug toolbar

Nyholm:
Awesome. 
I think I can manage some work the next coming days. 

Naturally, we should have integration with the debug bar. :)

Nyholm:
First PR submitted. Next up is debug toolbar and configurable plugins.

sagikazarmark:
Toolbar (data collector) and plugins has their own issue. Can this be closed?

Nyholm:
Sure. 

FYI. I got the toolbar implementation already. Im just going to fix and merge #16

sagikazarmark:
Oh, nice.

PR #11, opened 2015-12-24
Make sure you can configure clients and register them as services

Nyholm:
This will implement the feature mentioned in #10. 

This will let the user configure HttpClients for a specific implementation. Everything specific to a client, say guzzle5, is in one class. 

At the moment there is no validation of the special config. We just say: "The config you specify here is sent unparsed to Guzzle".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.4
   - 5.5
   - 5.6
   - 7.0
@@ -10,25 +9,22 @@ php:
 env:
   global:
     - TEST_COMMAND="composer test"
+  matrix:
+    - SYMFONY_VERSION=3.0.*
+    - SYMFONY_VERSION=2.8.*
     - SYMFONY_VERSION=2.7.*
 
 matrix:
-  allow_failures:
-    - php: 7.0
   fast_finish: true
-  include:
-    - php: 5.4
-      env:
-        - COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
-        - COVERAGE=true
-        - TEST_COMMAND="composer test-ci"
-    - php: 5.6
-      env: SYMFONY_VERSION=2.8.*
-    - php: 5.6
-      env: SYMFONY_VERSION=3.0.*
+  allow_failures:
+    - php: hhvm
+    - env: SYMFONY_VERSION=3.0.*
+    - php: 5.5
+      env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" && COVERAGE=true && TEST_COMMAND="composer test-ci" && SYMFONY_VERSION=2.7.*
 
 before_install:
   - travis_retry composer self-update
+  - wget https://github.com/puli/cli/releases/download/1.0.0-beta9/puli.phar && chmod +x puli.phar
 
 install:
   - composer require symfony/symfony:${SYMFONY_VERSION} --no-update
@@ -40,4 +36,3 @@ script:
 after_success:
   - if [[ "$COVERAGE" = true ]]; then wget https://scrutinizer-ci.com/ocular.phar; fi
   - if [[ "$COVERAGE" = true ]]; then php ocular.phar code-coverage:upload --format=php-clover build/coverage.xml; fi
-
--- a/ClientFactory/ClientFactoryInterface.php
+++ b/ClientFactory/ClientFactoryInterface.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Http\HttplugBundle\ClientFactory;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+interface ClientFactoryInterface
+{
+    /**
+     * Input an array of configuration to be able to create a HttpClient
+     *
+     * @param array $config
+     *
+     * @return \Http\Client\HttpClient
+     */
+    public function createClient(array $config = array());
+}
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -2,9 +2,11 @@
 
 namespace Http\HttplugBundle\DependencyInjection;
 
+use Http\HttplugBundle\ClientFactory\DummyClient;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 
 /**
@@ -22,6 +24,7 @@ class HttplugExtension extends Extension
 
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
 
+        $loader->load('services.xml');
         $loader->load('discovery.xml');
         foreach ($config['classes'] as $service => $class) {
             if (!empty($class)) {
@@ -32,6 +35,22 @@ class HttplugExtension extends Extension
 
         foreach ($config['main_alias'] as $type => $id) {
             $container->setAlias(sprintf('httplug.%s', $type), $id);
+        }
+
+        // Configure client services
+        $first = isset($config['clients']['default']) ? 'default' : null;
+        foreach ($config['clients'] as $name => $arguments) {
+            if ($first === null) {
+                $first = $name;
+            }
+
+            $def = $container->register('httplug.client.'.$name, DummyClient::class);
+            $def->setFactory([new Reference($arguments['factory']), 'createClient'])
+                ->addArgument($arguments['config']);
+        }
+
+        if ($first !== null) {
+            $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
         }
     }
 }
--- a/README.md
+++ b/README.md
@@ -41,11 +41,14 @@ For information how to write applications with the services provided by this bun
 
 ### Use in Applications
 
+#### Custom services
+
 This bundle provides 3 services: 
 
 * `httplug.client` a service that provides the `Http\Client\HttpClient`
 * `httplug.message_factory` a service that provides the `Http\Message\MessageFactory`
 * `httplug.uri_factory` a service that provides the `Http\Message\UriFactory`
+* `httplug.stream_factory` a service that provides the `Http\Message\StreamFactory`
 
 These services are always an alias to another service. You can specify your own service or leave the default, which is the same name with `.default` appended. The default services in turn use the service discovery mechanism to provide the best available implementation. You can specify a class for each of the default services to use instead of discovery, as long as those classes can be instantiated without arguments.
 
@@ -53,14 +56,47 @@ If you need a more custom setup, define the services in your application configu
 
 ```yaml
 httplug:
-    main_alias:
-        client: httplug.client.default
-        message_factory: httplug.message_factory.default
-        uri_factory: httplug.uri_factory.default
-    classes:
-        client: ~ # uses discovery if not specified
-        message_factory: ~
-        uri_factory: ~
+  main_alias:
+    client: httplug.client.default
+    message_factory: httplug.message_factory.default
+    uri_factory: httplug.uri_factory.default
+    stream_factory: httplug.stream_factory.default
+  classes:
+    # uses discovery if not specified
+    client: ~
+    message_factory: ~ 
+    uri_factory: ~
+    stream_factory: ~
+```
+
+#### Configure your client
+
+You can configure your clients with some good default options. The clients are later registered as services. 
+
+```yaml
+httplug:
+  clients: 
+    my_guzzle5: 
+      factory: 'httplug.factory.guzzle5'
+      config:
+        # These options are given to Guzzle without validation. 
+        base_url: 'http://google.se/'
+        defaults:
+          verify_ssl: false
+          timeout: 4
+          headers:
+            Content-Type: 'application/json'
+    acme: 
+      factory: 'httplug.factory.guzzle6'
+      config:
+        base_url: 'http://google.se/'
+       
+```
+
+```php
+
+$httpClient = $this->container->get('httplug.client.my_guzzle5');
+$httpClient = $this->container->get('httplug.client.acme');
 ```
 
 ### Use for Reusable Bundles
--- a/composer.json
+++ b/composer.json
@@ -14,21 +14,30 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "php-http/discovery": "^0.2.0",
+        "php": ">=5.5",
+        "php-http/discovery": "^0.6.3",
         "php-http/client-implementation": "^1.0",
-        "php-http/message-factory": "^0.2.0",
+        "php-http/message-factory": "^1.0",
+        "php-http/plugins": "dev-master",
         "symfony/framework-bundle": "^2.7|^3.0"
     },
     "require-dev": {
+        "phpunit/phpunit": "^4.4",
+        "php-http/guzzle6-adapter": "^0.3.1",
+        "php-http/message": "^0.2.1",
+        "php-http/client-common": "^0.1.1",
         "symfony/symfony": "^2.7|^3.0",
-        "php-http/guzzle6-adapter": "^0.2",
         "polishsymfonycommunity/symfony-mocker-container": "~1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "0.*"
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7"
     },
     "autoload": {
         "psr-4": {
             "Http\\HttplugBundle\\": ""
         }
+    },
+    "scripts": {
+        "test": "vendor/bin/phpunit",
+        "test-ci": "vendor/bin/phpunit --coverage-text --coverage-clover=build/coverage.xml"
     },
     "extra": {
         "branch-alias": {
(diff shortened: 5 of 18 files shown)


ClientFactory/ClientFactoryInterface.php
```diff
@@ -0,0 +1,27 @@
...
+ * Interface ClientFactoryInterface.
+ */
+interface ClientFactoryInterface
+{
+    /**
+     * Configure and return a client.
+     *
+     * @param array $config
+     *
+     * @return HttpClient
```
dbu: what should happen if this does not work out? i would rather throw an exception than allow returning null

ClientFactory/DummyFactory.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Configure and return a client.
+     *
+     * @param array $config
+     *
+     * @return HttpClient
+     */
+    public function configure(array $config = array())
+    {
+        return;
```
dbu: i would prefer to throw an exception explaining that something went wrong. null is always a pain to debug.

ClientFactory/GenericClientFactory.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+class GenericClientFactory
+{
+    /**
+     * @var ClientFactoryInterface[]
+     */
+    private $clientFactories;
+
+    /**
+     * @param string $adapter
```
dbu: i would call this argument $name

DependencyInjection/HttplugExtension.php
```diff
@@ -33,5 +36,31 @@ public function load(array $configs, ContainerBuilder $container)
         foreach ($config['main_alias'] as $type => $id) {
             $container->setAlias(sprintf('httplug.%s', $type), $id);
         }
+
+        // Configure services
+        foreach ($config['clients'] as $name => $arguments) {
+            $this->verifyDependencies($arguments['adapter']);
+            $def = $container->register('httplug.'.$name, DummyFactory::class);
+            $def->setFactory([new Reference('httplug.client.factory'), 'getClient'])
```
dbu: we can't we use the specific factory directly, but need the GenericClientFactory?

ClientFactory/GenericClientFactory.php
```diff
@@ -0,0 +1,45 @@
...
+namespace Http\HttplugBundle\ClientFactory;
+
+use Http\Client\HttpClient;
+
+/**
+ * This is a client factory that will help you instantiate any client.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class GenericClientFactory
```
dbu: why not use a service name pattern within the DI extension to use the correct factory service directly? does this level of indirection gain us anything additional?

ClientFactory/DummyFactory.php
Nyholm: This is just a dummy class that is not used. Symfony DI needs it when you register a service event though you use a factory. 

I should throw an exception saying that you are using it wrong.

dbu:
thanks for the pull request! from the bundle user perspective, this looks very good. for the implementation side, i am a bit unsure as it seems we duplicate discovery in some way. then again, its not very heavy so i could live with that.

and @sagikazarmark had some ideas about using puli as another way to find implementations. mark, what is your opinion on this PR? how would puli fit in here?

README.md
```diff
@@ -53,14 +56,48 @@ If you need a more custom setup, define the services in your application configu
...
+
+#### Configure your client
+
+You can configure your clients with some good default options. The clients are later registered as services. 
+
+```yaml
+httplug:
+  clients: 
+    my_guzzle5: 
+      adapter: guzzle5
```
dbu: we could treat this as a shortcut for a service name and put httplug.client. in front of it. then people could also specify full service names instead and have their own factories.

DependencyInjection/HttplugExtension.php
Nyholm: I we use the specific factory direct we need to keep a list that maps name=>factory class. Because at this point we do not know any of the DI tags. That would mean if a user would like to add a factory e need to update/extend this bundle. 

Using a generic factory is mor SOLID.

ClientFactory/DummyFactory.php
dbu: yeah, that is what i meant. in some weird case, people might also manage
to have the compiler pass not run or something.

Nyholm: Good point!

ClientFactory/ClientFactoryInterface.php
Nyholm: Done. We throw an `ClientWasNotCreated` exception.

Nyholm:
To address comments about the generic factory (https://github.com/php-http/HttplugBundle/pull/11#discussion_r48436783, https://github.com/php-http/HttplugBundle/pull/11#discussion_r48436696, https://github.com/php-http/HttplugBundle/pull/11#discussion_r48436679):

I want to have a generic factory to which you could add instances of a `ClientFactoryInterface` by simply add a `httplug.client_factory` tag to the service definition. This means that we do not need a list of `adapter name`=>´factory class`. And it makes the bundle easily extendable if someone would like to add their own class. 

```yaml
httplug:
  clients: 
    my_guzzle5: 
      adapter: my_special_client
```

```php
class SpecialFactory implements ClientFactoryInterface
{
    public function createClient(array $config = array())
    {
        return new SpecialClient($config);
    }

    public function getName()
    {
        return 'my_special_client';
    }
}
```
```xml
<service id="acme.factory.mySpecialFactory" class="Acme\SpecialFactory" public="false">
       <tag name="httplug.client_factory" />
</service>
```

We could get around using a generic factory and achieve the same goal by doing this in a compiler pass. The drawback of this is we need to store a lot of configuration in the container's parameters. That approach seams also more "magic" and dirty IMO.

ClientFactory/ClientFactoryInterface.php
sagikazarmark: If something, then rather ClientNotFound

Nyholm: My intention was to include the scenario where a clients configuration was wrong and we could not create a client because of it.

sagikazarmark: Then InvalidConfiguration or something like that, but ClientWasNotCreated does not tell too much about the "WHY" in my opinion.

Nyholm: Good point, I'll change to `InvalidConfiguration`

sagikazarmark:
Puli provides a repository and a discovery layer. With puli we can find the installed classes, but so far, I am not sure we can instantiate/configure them automatically. That said, would be good if puli could replace the current discovery layer.

On the other hand, there is an upcoming DefinitionInterface which could work with Puli nicely:
http://www.thecodingmachine.com/interoperable-php-packages-with-definition-interop-and-puli/

Although it is far in the future.

Client Factory is also something I was thinking about (I think we actually had that interface in HTTPlug earlier).

I am also a bit unsure about this. Mainly because I can't really see the difference between creating configuration, passing it to a client factory and creating a client service directly with configuration. You need config and a registered service both ways. What is the advantage then?

Nyholm:
> Mainly because I can't really see the difference between creating configuration, passing it to a client factory and creating a client service directly with configuration. You need config and a registered service both ways. What is the advantage then?

For the Guzzle clients, there is none. You are correct. You can do something like this in the Extension class:
```php
$guzzleClient = $container->register('httplug.guzzle.client', \GuzzleHttp\Client::class);
$guzzleClient->addArgument($config);
$httpClient = $container->register('httplug.guzzle', \Http\Adapter\Guzzle5HttpAdapter::class);
$httpClient->addArgument(new Reference('httplug.guzzle.client'));
```

But how would you do it for Buzz? Their `Browser` need a `Client`, say that we choose their `Curl` client then we have to configure options with a `setOption($option, $value)` method. That would be a pain to do in the Extension class. It would also violate the single responsibility principle and the open/close principle. This is where a factory is beneficial. 

This implementation is SOLID. We could event put the implementation factories (`Guzzle5Factory`, `Guzzle6Factory` and `BuzzFactory`) in a separate package if we would like to. That would make the Httplug bundle apply with the common reuse principle. But that's a decision we do make later.

README.md
dbu: if this would be the full factory service name instead of a name for the meta-factory, we could get rid of the the service tags and the meta factory that registers factories by short name (aka is a mini-container for httplug factories)

dbu:
i think i can agree to have factories that are tailored to the symfony DI use case in this bundle. its small utitlity code and we can configure several clients which seems a good thing, and difficult to do otherwise. also, having the client configuration in symfony config rather than in some puli resource file (if that is even possible) feels more natural for symfony.

i am still unsure about the generic factory, however. tried to explain my idea in https://github.com/php-http/HttplugBundle/pull/11#discussion_r48656231

Nyholm:
> i am still unsure about the generic factory, however. tried to explain my idea in #11 (comment)

You are correct with your comment. The generic factory is not needed if we specify the factory service id instead of the factory name. That solution is easier, I should update. Thank you!

Nyholm:
There you go.

dbu:
cool! i think this is now ready. can you squash the commits please?

there seems to be an issue with the configuration test using an outdated namespace for the guzzle adapter (see https://travis-ci.org/php-http/HttplugBundle/jobs/99706320) - can you maybe adjust that as well?

Nyholm:
Sure, no problem. I squashed the commits. 

I've aslo updated the old namespace for Guzzle6 and updated the discovery package to 0.6.0. I still get an error though. 

```
RuntimeException: Puli Factory is not available
```

sagikazarmark:
Puli cli must be added at least as development dependency to make tests passing.

Nyholm:
> Puli cli must be added at least as development dependency to make tests passing.

Done. Thank you. 

But shouldn't that be added as a dependency for the discovery? Could you use discovery without Puli/cli?

sagikazarmark:
No, you can't. But you can install it either as composer dependency in your project or the executable in your prefix. It is under discussion if we should make it a dependency in discovery.

Nyholm:
@sagikazarmark as always: Thank you for a clarifying answer.

.travis.yml
```diff
@@ -10,14 +9,12 @@ php:
 env:
   global:
     - TEST_COMMAND="composer test"
-    - SYMFONY_VERSION=2.7.*
+    - SYMFONY_VERSION=2.8.*
```
dbu: i agree that we should change this, but maybe to 3.0 right away. we need to adjust the symfony version for the matrix lines below as well so we still test 2.7, 2.8 and 3.0

dbu:
we seem to run into a conflict with symfony 3 and discovery / puli: https://travis-ci.org/php-http/HttplugBundle/jobs/99844044

and we need to increase the lowest version of discovery that we allow to make the lowest version tests pass again: https://travis-ci.org/php-http/HttplugBundle/jobs/99844042

.travis.yml
Nyholm: Should I also use PHP7 as standard?

sagikazarmark:
Hm, I'll check why puli fails with symfony 3.0.

Nyholm:
I'm on it.. It is because puli/cli requires symfony/filesystem:~2.3

Im about to see if I can send a PR.

Nyholm:
I've submitted a PR: https://github.com/puli/cli/pull/30

There will be no more conflict with Symfony 3 when the Puli/cli PR is merged.

Nyholm:
I've also submitted https://github.com/padraic/phar-updater/pull/16

I think the best solution is not to require `puli/cli`. Let's use the puli.phar instead. 

The issue we are facing now is that `php-http/discovery:0.6` requires `puli/composer-plugin:1.0.0-beta8` which requires `symfony/process: ^2.3`. 

The `puli/composer-plugin:dev-master` has the proper dependency of `symfony/process: ^2.3|^3.0`.

dbu:
trying to make discovery more flexible: https://github.com/php-http/discovery/pull/37

dbu:
https://github.com/php-http/discovery/pull/37 is merged. are there viable releases already? at least discovery needs to be @ dev for now.

sagikazarmark:
We can tag 0.6.1 discovery.

Nyholm:
Let me know when you have tagged 0.6.1 so I can update

sagikazarmark:
Done

sagikazarmark:
Before merging this: can we make sure that the Guzzle5 Adapter follows the newly accepted naming conventions (Http\Adapter\Guzzle5\Client namespace)

Nyholm:
> Before merging this: can we make sure that the Guzzle5 Adapter follows the newly accepted naming conventions (Http\Adapter\Guzzle5\Client namespace)

Of course

Nyholm:
There are more puli related issues. 

I *think* that `webmozart/key-value-store` as a missing dependency on `webmozart/json`. Aslo the `GeneratedPuliFactory` is not compatible with `webmozart/key-value-store`. 

Ping @webmozart
```php
class GeneratedPuliFactory
{
    /**
     * Creates the resource repository.
     *
     * @return ResourceRepository The created resource repository.
     */
    public function createRepository()
    {
        if (!interface_exists('Puli\Repository\Api\ResourceRepository')) {
            throw new RuntimeException('Please install puli/repository to create ResourceRepository instances.');
        }

        $store = new JsonFileStore(__DIR__.'/path-mappings.json', true); // <-- This line
        $repo = new PathMappingRepository($store, __DIR__.'/..');

        return $repo;
    }
// ...
}
```

```php

class JsonFileStore implements SortableStore, CountableStore
{
    // ---
    public function __construct($path, $flags = 0)
    {
        Assert::string($path, 'The path must be a string. Got: %s');
        Assert::notEmpty($path, 'The path must not be empty.');
        Assert::integer($flags, 'The flags must be an integer. Got: %s'); // <-- $flag will be a boolean
    // ...
    }
}

**Comment by sagikazarmark** [comment]
Yes, there is an incompatibility between puli versions. Not sure if it is fixed or not, but can't really do anything about it, until a new beta version is tagged.

**Comment by webmozart** [comment]
If you use the latest beta releases of all the packages (not dev), everything *should* work. The dependency on webmozart/json is optional, but it probably makes sense to include it in puli/composer-plugin since json is the default backend.

**Comment by sagikazarmark** [comment]
@webmozart Everything, except Symfony 3, right? (In case of Puli CLI)

**Comment by webmozart** [comment]
@sagikazarmark Yes exactly. Symfony 3 support was only added in master. Btw. if you need chat support, join https://gitter.im/puli/issues, there's always people hanging around.

File: composer.json [hunk 8]
```diff
@@ -14,22 +14,31 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "php-http/discovery": "^0.2.0",
+        "php": ">=5.5",
+        "php-http/discovery": "^0.6.1",
```
dbu: this would now need to be 0.6.3.

but we have the same discussion as in https://github.com/FriendsOfSymfony/FOSHttpCache/issues/268 - maybe we should not depend on discovery and explain to add a client and either be explicit or additionally add discovery. and adjust our code to nicely report when discovery is missing and nothing was configured.

okay, but that is a bit out of scope for this PR, if you can upgrade to 0.6.3 for now, that should hopefully fix the build.

composer.json
```diff
@@ -14,22 +14,31 @@
...
         "php-http/client-implementation": "^1.0",
-        "php-http/message-factory": "^0.2.0",
+        "php-http/message-factory": "^1.0",
+        "php-http/plugins": "dev-master",
```
dbu: do we have a tag that we can use here?

composer.json
```diff
@@ -14,22 +14,31 @@
...
         "symfony/framework-bundle": "^2.7|^3.0"
     },
     "require-dev": {
+        "phpunit/phpunit": "^4.4",
+        "php-http/guzzle6-adapter": "^0.3.1",
+        "php-http/message": "^0.2",
+        "php-http/client-common": "^0.1",
```
dbu: are these versions still ok?

composer.json
Nyholm: No, there are not tags in the plugins repo

composer.json
sagikazarmark: I think our discover layer can just be replaced by puli discovery. The symfony bundle provides a service for it. See how it is used in our discovery.

composer.json
sagikazarmark: They seem to be OK.

.travis.yml
dbu: yeah lets do that!

Nyholm:
I can't get it to work with Symfony3 and our discovery since it depends on Puli/cli. I think it is okey to allow Symfony3 to fail at this point until the issues discussed are resolved. (The bundle is not even on packagist yet.)

There is also an issue with the lowest stability test. It seams that puli cant find a `Http\Client\HttpClient`. I am using all the lastest versions of php-http/* and I cant figure out why it fails. Any suggestions?

dbu:
i am okay to skip symfony 3 until the chain around puli / phar etc is
cleaned up.

about lowest version: did you open the composer install line on
travis-ci to see what exact versions it installed? maybe the difference
is visible there?

sagikazarmark:
There is a warning at the end of composer install command:

`Warning: Could not load Puli packages: RuntimeException: Function has already been defined`

Nyholm:
That is a dependency error that is out of scope for this PR and bundle. I guess that will be solved when we can rely on stable versions of Puli. I'm adding the "prefer-lowest" test to *allow_failures* as well.

sagikazarmark:
> That is a dependency error that is out of scope for this PR and bundle.

Wow, this is the best "I don't know what causes the error" explanation I have ever heard. :stuck_out_tongue_closed_eyes:

Nyholm:
haha =)

dbu:
thanks tobias. created #13 and  #14 about the build failures.

Nyholm:
Thank you for merging.

Issue #15, opened 2016-01-05
Configure plugins clients with plugins


Nyholm:
Can I get some feedback on this approach to configure the plugins?

```yaml
my_redirect_plugin:
  class: Http\Client\Plugin\RedirectPlugin 
my_logger_plugin:
  class: Http\Client\Plugin\LoggerPlugin 
  arguments: ["@monolog"]
my_cache_plugin:
  class: Http\Client\Plugin\CachePlugin 
  arguments: ["@cache", "@stream_factory"]

httplug:
  clients: 
    my_guzzle5: 
      factory: 'httplug.factory.guzzle5'
      plugins: ['@my_logger_plugin', '@my_cache_plugin']
      config:
        defaults:
          verify_ssl: false
          timeout: 4
    acme: 
      factory: 'httplug.factory.guzzle6'
      plugins: ['@my_logger_plugin', '@my_redirect_plugin']
      config:
        base_uri: 'http://google.se/'
```

This would allow you to use your own plugins, set the order etc. It would also be easy to inject a JournalPlugin to use internally for the debug toolbar. 

We could go with an approach where we name the plugins etc etc but it is a simplier implementation and easier to use (no magic) if we reference them as a service.

dbu:
this looks simple and clean enough, i like it!

maybe we can provide default services for the plugins provided by php-http/plugin, if needed with options to configure things on them. but as with client factories, you would just use the full service name php_http.plugin.redirect (or similar) to add the plugin to a client, no shortcut that adds magic.

Nyholm:
Closed by #16

PR #16, opened 2016-01-05
Plugins

Nyholm:
Added plugins with the configuration feature discussed in #15.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -90,6 +90,10 @@ class Configuration implements ConfigurationInterface
                         ->cannotBeEmpty()
                         ->info('The service id of a factory to use when creating the adapter.')
                     ->end()
+                    ->arrayNode('plugins')
+                        ->info('A list of service ids of plugins. The order is important.')
+                        ->prototype('scalar')->end()
+                    ->end()
                     ->variableNode('config')->end()
                 ->end()
             ->end();
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -25,6 +25,7 @@ class HttplugExtension extends Extension
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
 
         $loader->load('services.xml');
+        $loader->load('plugins.xml');
         $loader->load('discovery.xml');
         foreach ($config['classes'] as $service => $class) {
             if (!empty($class)) {
@@ -36,8 +37,19 @@ class HttplugExtension extends Extension
         foreach ($config['main_alias'] as $type => $id) {
             $container->setAlias(sprintf('httplug.%s', $type), $id);
         }
+        $this->configureClients($container, $config);
 
-        // Configure client services
+
+    }
+
+    /**
+     * Configure client services
+     *
+     * @param ContainerBuilder $container
+     * @param array $config
+     */
+    protected function configureClients(ContainerBuilder $container, array $config)
+    {
         $first = isset($config['clients']['default']) ? 'default' : null;
         foreach ($config['clients'] as $name => $arguments) {
             if ($first === null) {
@@ -45,10 +57,22 @@ class HttplugExtension extends Extension
             }
 
             $def = $container->register('httplug.client.'.$name, DummyClient::class);
-            $def->setFactory([new Reference($arguments['factory']), 'createClient'])
-                ->addArgument($arguments['config']);
+
+            if (empty($arguments['plugins'])) {
+                $def->setFactory([new Reference($arguments['factory']), 'createClient'])
+                    ->addArgument($arguments['config']);
+            } else {
+                $def->setFactory('Http\HttplugBundle\ClientFactory\PluginClientFactory::createPluginClient')
+                    ->addArgument(array_map(function($id) {
+                        return new Reference($id);
+                    }, $arguments['plugins']))
+                    ->addArgument(new Reference($arguments['factory']))
+                    ->addArgument($arguments['config']);
+            }
+
         }
 
+        // Alias the first client to httplug.client.default
         if ($first !== null) {
             $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
         }
--- a/README.md
+++ b/README.md
@@ -43,19 +43,26 @@ For information how to write applications with the services provided by this bun
 
 #### Custom services
 
-This bundle provides 3 services: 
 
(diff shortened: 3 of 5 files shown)


DependencyInjection/HttplugExtension.php
```diff
@@ -45,10 +57,22 @@ public function load(array $configs, ContainerBuilder $container)
             }
 
             $def = $container->register('httplug.client.'.$name, DummyClient::class);
-            $def->setFactory([new Reference($arguments['factory']), 'createClient'])
-                ->addArgument($arguments['config']);
+
+            if (empty($arguments['plugins'])) {
```
Nyholm: Do we want this? An alternative could be to send everything to PluginClientFactory so every client is an instance of PluginClient.

DependencyInjection/Configuration.php
```diff
@@ -90,6 +90,10 @@ protected function configureClients(ArrayNodeDefinition $root)
                         ->cannotBeEmpty()
                         ->info('The service id of a factory to use when creating the adapter.')
                     ->end()
+                    ->arrayNode('plugins')
+                        ->info('A list of service ids of plugins. The order is imortant.')
```
dbu: /s/imortant/important/

DependencyInjection/HttplugExtension.php
dbu: i would keep it. there is no way to add plugins later at runtime, so it would only be relevant for people that typehint the concrete class instead of the interface, which would be so wrong that we should not care about helping them :-)

README.md
```diff
@@ -99,6 +99,27 @@ $httpClient = $this->container->get('httplug.client.my_guzzle5');
...
 ```
 
+#### Plugins
+
+You can configure the clients with plugins. 
+
+```yaml
+// services.yml
+my_cache_plugin:
+  class: Http\Client\Plugin\CachePlugin
```
dbu: i think we should use a fictional plugin class name here, as we provide the standard plugins as services.

Resources/config/plugins.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="httplug.plugin.cache" class="Http\Client\Plugin\CachePlugin" public="false">
+            <argument />
```
dbu: where is this completed? do we need configuration for this?

Resources/config/plugins.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="httplug.plugin.cache" class="Http\Client\Plugin\CachePlugin" public="false">
```
dbu: agree that the plugins should be private (= removed from the container after it is built). we should list the services that are available in the documentation, and also remind that they are private and can thus not be manually fetched from the container.

dbu:
:+1: thanks tobias!

Resources/config/plugins.xml
Nyholm: Hm, no it is not completed. I missed that. I have excluded a couple of plugin that require dependencies. The cache plugin must have a PSR6 cache pool and there is no way for me to give a good default. 

I think the best solution is to allow some sort of configuration like: 

```yaml
httplug: 
  plugins: 
    cache:
      service: 'acme.cache'
      config: 
        respect_headers: true
    logger:
      formatter: 'acme.formatter'
```

If you don't mind, could I do that in a separate PR?

Nyholm:
Thank you for reviewing this so quickly. I've updated according to your comments.

Resources/config/plugins.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="httplug.plugin.content_length" class="Http\Client\Plugin\ContentLengthPlugin" public="false" />
+        <service id="httplug.plugin.decoder" class="Http\Client\Plugin\DecoderPlugin" public="false">
+            <argument />
```
dbu: there is an undefined argument here

Resources/config/plugins.xml
```diff
@@ -0,0 +1,19 @@
...
+        </service>
+        <service id="httplug.plugin.error" class="Http\Client\Plugin\ErrorPlugin" public="false" />
+        <service id="httplug.plugin.logger" class="Http\Client\Plugin\LoggerPlugin" public="false">
+            <argument type="service" id="logger"/>
+            <argument />
```
dbu: and here as well

dbu:
i agree solving plugin configuration in a separate PR. but right now there are two plugin configs with empty <argument/> tags, i think that will cause problems, no? if we can omit the params we should, otherwise we should remove those plugins for now.

Resources/config/plugins.xml
Nyholm: Isn't that okey since the decoder has a default value there?

Nyholm:
We can omit those. 

Ignore my comment in: https://github.com/php-http/HttplugBundle/pull/16#discussion_r48978328

Resources/config/plugins.xml
dbu: not sure. do the tests cover this? if we have a fuctional test that manages
to load an actual client service with plugins set on it, we would see. atm
we lack a bit on testing afaik

Nyholm: We do not have tests that cover this. You are correct that the bundle lacks tests.

dbu:
alright, i think this is ready to merge. @Nyholm can you please squash the commits? feel free to merge, things look good for me.

Nyholm:
Done. I've squashed my commints

dbu:
great, thanks a lot!

Nyholm:
Thank you for merging.

Issue #3, opened 2015-08-24
DataCollector for the symfony toolbar


dbu:
see also https://github.com/egeloen/IvoryHttpAdapterBundle/blob/master/DataCollector/IvoryHttpAdapterDataCollector.php and https://github.com/misd-service-development/guzzle-bundle/blob/master/DataCollector/GuzzleDataCollector.php

i think this will also need to use some tags to discover additional clients that are defined as service.

joelwurtz:
Can be done with a plugin i think, or a decorator if we don't want to depend on the plugin system

dbu:
i think using the plugin system is good. my idea for the bundle is that it helps with plugins anyways.

sagikazarmark:
With the History plugin this could easily be done. Another possibility is using AOP to catch all the HttpClient calls. It would be more generic and it would not require to add a plugin.

http://go.aopphp.com/

dbu:
aopphp sounds interesting, but could have a rather massive impact on the
rest of the application. i think the History plugin is exactly what we want.

we need to have a system to active plugins into clients anyways, so this
should be no big deal to use it.

joelwurtz:
I was thinking some sort of StandalonePluginClient, which transform only one Plugin into an HttpClient / HttpAsyncClient (like the DataCollector thing will only add this with the HistoryPlugin to the actual HttpClient when in debug mode / toolbar activated)

sagikazarmark:
What if the http client itself is already a plugin client? Also, I think it is not generic: you might be able to create an http client service without httplug bundle.

Although I agree that AOP could have an impact on the application itself, not sure if the history plugin is generic enough. That said, I agree it would be much nicer.

What if the AOP solution is placed in a separate bundle? Like HttplugDebugBundle? That would only be enabled in development mode, so that the impact on the production application would be zero. Does that make any sense?

dbu:
i will try to get a first thing working at all, first :-)

the aop debug thing should probably not be limited to symfony. but i am
not sure, feels like going very far. the approach of the guzzle bundle
is to tag every client service as guzzle client, so that plugins can be
added in debug env.

sagikazarmark:
Ok, I think this AOP thing is not top prio. It is rather something when you give a new game to the kid: he/she wants to play with it all the time. :stuck_out_tongue:

sagikazarmark:
Is this solved?

dbu:
fixed in #19

Issue #17, opened 2016-01-06
Configure default plugins


Nyholm:
There are some plugins in `php-http/plugins` that got dependencies. We should allow the user to configure those plugins. 

```yaml
httplug: 
  plugins: 
    cache:
      service: 'acme.cache'
      config: 
        respect_headers: true
    logger:
      formatter: 'acme.formatter'
```

Reference: https://github.com/php-http/HttplugBundle/pull/16#discussion_r48946387

dbu:
agree. we can use the `->canBeEnabled()` construct in configuration to know for each plugin whether it should be loaded. as soon as any other option is configured, enabled becomes true automatically. the DI extension would pass parameters (retry limit, redirect limit, cache service and so on) and remove plugins that are disabled from the container.

PR #19, opened 2016-01-08
Added web profiler and toolbar

Nyholm:
As mentioned in #3. 

This PR implements a web debug toolbar. It uses the HistoryPlugin.

### DIFF
--- a/Collector/MessageJournal.php
+++ b/Collector/MessageJournal.php
@@ -37,7 +37,7 @@ class MessageJournal extends DataCollector implements Journal
     public function addSuccess(RequestInterface $request, ResponseInterface $response)
     {
         $this->data['success'][] = [
-            'request'  => $this->formatter->formatRequest($request),
+            'request' => $this->formatter->formatRequest($request),
             'response' => $this->formatter->formatResponse($response),
         ];
     }
@@ -56,7 +56,7 @@ class MessageJournal extends DataCollector implements Journal
         }
 
         $this->data['failure'][] = [
-            'request'  => $this->formatter->formatRequest($request),
+            'request' => $this->formatter->formatRequest($request),
             'response' => $formattedResponse,
         ];
     }
--- a/Collector/MessageJournal.php
+++ b/Collector/MessageJournal.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace Http\HttplugBundle\Collector;
+
+use Http\Client\Exception;
+use Http\Client\Plugin\Journal;
+use Http\Message\Formatter;
+use Http\Message\Formatter\SimpleFormatter;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class MessageJournal extends DataCollector implements Journal
+{
+    /**
+     * @var Formatter
+     */
+    private $formatter;
+
+    /**
+     * @param Formatter $formatter
+     */
+    public function __construct(Formatter $formatter = null)
+    {
+        $this->formatter = $formatter ?: new SimpleFormatter();
+        $this->data = ['success' => [], 'failure' => []];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addSuccess(RequestInterface $request, ResponseInterface $response)
+    {
+        $this->data['success'][] = [
+            'request'  => $this->formatter->formatRequest($request),
+            'response' => $this->formatter->formatResponse($response),
+        ];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addFailure(RequestInterface $request, Exception $exception)
+    {
+        if ($exception instanceof Exception\HttpException) {
+            $formattedResponse = $this->formatter->formatResponse($exception->getResponse());
+        } elseif ($exception instanceof Exception\TransferException) {
+            $formattedResponse = $exception->getMessage();
+        } else {
+            $formattedResponse = sprintf('Unexpected exception of type "%s"', get_class($exception));
+        }
+
+        $this->data['failure'][] = [
+            'request'  => $this->formatter->formatRequest($request),
+            'response' => $formattedResponse,
+        ];
+    }
+
+    /**
+     * Get the successful request-resonse pairs.
+     *
+     * @return array
+     */
+    public function getSucessfulRequests()
+    {
+        return $this->data['success'];
+    }
+
+    /**
+     * Get the failed request-resonse pairs.
+     *
+     * @return array
+     */
+    public function getFailedRequests()
+    {
+        return $this->data['failure'];
+    }
+
+    /**
+     * Get the total number of request made.
+     *
+     * @return int
+     */
+    public function getTotalRequests()
+    {
+        return count($this->data['success']) + count($this->data['failure']);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function collect(Request $request, Response $response, \Exception $exception = null)
+    {
+        // We do not need to collect any data form the Symfony Request and Response
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'httplug';
+    }
+}
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -34,8 +34,7 @@ class Configuration implements ConfigurationInterface
                     return !empty($v['classes']['client'])
                         || !empty($v['classes']['message_factory'])
                         || !empty($v['classes']['uri_factory'])
-                        || !empty($v['classes']['stream_factory'])
-                    ;
+                        || !empty($v['classes']['stream_factory']);
                 })
                 ->then(function ($v) {
                     foreach ($v['classes'] as $key => $class) {
@@ -72,8 +71,19 @@ class Configuration implements ConfigurationInterface
                         ->scalarNode('stream_factory')->defaultNull()->end()
                     ->end()
                 ->end()
-            ->end()
-        ;
+                ->arrayNode('toolbar')
+                    ->addDefaultsIfNotSet()
+                    ->info('Extend the debug profiler with inforation about requests.')
+                    ->children()
+                        ->enumNode('enabled')
+                            ->info('If "auto" (default), the toolbar is activated when kernel.debug is true. You can force the toolbar on and off by changing this option.')
+                            ->values([true, false, 'auto'])
+                            ->defaultValue('auto')
+                        ->end()
+                        ->scalarNode('formatter')->defaultNull()->end()
+                    ->end()
+                ->end()
+            ->end();
 
         return $treeBuilder;
     }
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -3,8 +3,8 @@
 namespace Http\HttplugBundle\DependencyInjection;
 
 use Http\HttplugBundle\ClientFactory\DummyClient;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
@@ -27,6 +27,18 @@ class HttplugExtension extends Extension
         $loader->load('services.xml');
         $loader->load('plugins.xml');
         $loader->load('discovery.xml');
+
+        $enabled = is_bool($config['toolbar']['enabled']) ? $config['toolbar']['enabled'] : $container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug');
+        if ($enabled) {
+            $loader->load('data-collector.xml');
+            $config['_inject_collector_plugin'] = true;
+
+            if (!empty($config['toolbar']['formatter'])) {
+                $container->getDefinition('httplug.collector.message_journal')
+                    ->replaceArgument(0, new Reference($config['toolbar']['formatter']));
+            }
+        }
+
         foreach ($config['classes'] as $service => $class) {
             if (!empty($class)) {
                 $container->removeDefinition(sprintf('httplug.%s.default', $service));
@@ -52,6 +64,10 @@ class HttplugExtension extends Extension
         foreach ($config['clients'] as $name => $arguments) {
             if ($first === null) {
                 $first = $name;
+            }
+
+            if (isset($config['_inject_collector_plugin'])) {
+                array_unshift($arguments['plugins'], 'httplug.collector.history_plugin');
             }
 
             $def = $container->register('httplug.client.'.$name, DummyClient::class);
--- a/README.md
+++ b/README.md
@@ -51,9 +51,10 @@ For information how to write applications with the services provided by this bun
 | httplug.stream_factory | Service* that provides the `Http\Message\StreamFactory`
 | httplug.client.[name] | This is your Httpclient that you have configured. With the configuration below the name would be `acme_client`.
 | httplug.client | This is the first client configured or a client named `default`.
-| httplug.plugin.content_length <br> httplug.plugin.decoder<br> httplug.plugin.error<br> httplug.plugin.logger<br> httplug.plugin.redirect<br> httplug.plugin.retry | These are build in plugins that lives in the `php-http/plugins` package. These servcies are not public and may only be used when configure HttpClients or services.
+| httplug.plugin.content_length <br> httplug.plugin.decoder<br> httplug.plugin.error<br> httplug.plugin.logger<br> httplug.plugin.redirect<br> httplug.plugin.retry | These are built in plugins that live in the `php-http/plugins` package. These servcies are not public and may only be used when configure HttpClients or services. 
 
 \* *These services are always an alias to another service. You can specify your own service or leave the default, which is the same name with `.default` appended. The default services in turn use the service discovery mechanism to provide the best available implementation. You can specify a class for each of the default services to use instead of discovery, as long as those classes can be instantiated without arguments.*
+
 
 If you need a more custom setup, define the services in your application configuration and specify your service in the `main_alias` section. For example, to add authentication headers, you could define a service that decorates the service `httplug.client.default` with a plugin that injects the authentication headers into the request and configure `httplug.main_alias.client` to the name of your service.
 
--- a/Resources/config/data-collector.xml
+++ b/Resources/config/data-collector.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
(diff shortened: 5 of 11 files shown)


Collector/MessageJournal.php
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public function getTotalRequests()
+    {
+        return count($this->data['success'])+count($this->data['failure']);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function collect(Request $request, Response $response, \Exception $exception = null)
```
dbu: is this intentionally empty? if so, please add a comment why it is there and empty.

DependencyInjection/HttplugExtension.php
```diff
@@ -56,6 +61,10 @@ protected function configureClients(ContainerBuilder $container, array $config)
                 $first = $name;
             }
 
+            if (isset($config['_inject_journal_plugin'])) {
+                array_unshift($arguments['plugins'], 'httplug.plugin.history');
```
dbu: what if somebody was already adding the history plugin? should we array_search ('httpplug.plugin.history') to avoid that?

README.md
```diff
@@ -51,14 +51,16 @@ For information how to write applications with the services provided by this bun
 | httplug.stream_factory | Service* that provides the `Http\Message\StreamFactory`
 | httplug.client.[name] | This is your Httpclient that you have configured. With the configuration below the name would be `acme_client`.
 | httplug.client | This is the first client configured or a client named `default`.
-| httplug.plugin.content_length <br> httplug.plugin.decoder<br> httplug.plugin.error<br> httplug.plugin.logger<br> httplug.plugin.redirect<br> httplug.plugin.retry | These are build in plugins that lives in the `php-http/plugins` package. These servcies are not public and may only be used when configure HttpClients or services. 
+| httplug.plugin.content_length <br> httplug.plugin.decoder<br> httplug.plugin.error<br> httplug.plugin.history<br> httplug.plugin.logger<br> httplug.plugin.redirect<br> httplug.plugin.retry | These are build in plugins that lives in the `php-http/plugins` package. These servcies are not public and may only be used when configure HttpClients or services.<br><br>*The History plugin is only available when the kernel runs on debug mode and `httplug.toolbar: true`. It gets automatically injected to all clients.*
```
dbu: s/build in/built in/

dbu: s/that lives/that live/  (plural)

DependencyInjection/HttplugExtension.php
dbu: ah, read the doc now. maybe throw an InvalidConfigurationException if its explicitly specified? then again, why not allow users to activate the history plugin and do their own thing, even in prod mode if they feel its a good idea?

dbu:
very cool! i am impressed how simple this looks!

DependencyInjection/HttplugExtension.php
Nyholm: We cant do a search/replace. People should not use "our" namespace when defining services so they would create a service like `my_history_plugin`. So it is impossible to do a search because we dont know what to search for... 

If someone uses our service, then the history will be logged twice. The readme says it is added automatically. We should not bother. 

What we should care about is that what if they are not happy with our formatter? But I think that can be solved by #17.

Collector/MessageJournal.php
Nyholm: Okey, thank you. 

A upcoming PR maybe want to log which Symfony Request that made which Httplug Request. But it should be empty for now.

Nyholm:
> very cool! i am impressed how simple this looks!

Thank you. So am I. I barley wrote any code just some clue to make `Httplug\*` classes talk with Symfony.

DependencyInjection/HttplugExtension.php
dbu: ah, got it. maybe we should use a more specific service name for this. its not "the" history plugin, but a history plugin linked to the message journal.

README.md
```diff
@@ -51,14 +51,16 @@ For information how to write applications with the services provided by this bun
 | httplug.stream_factory | Service* that provides the `Http\Message\StreamFactory`
 | httplug.client.[name] | This is your Httpclient that you have configured. With the configuration below the name would be `acme_client`.
 | httplug.client | This is the first client configured or a client named `default`.
-| httplug.plugin.content_length <br> httplug.plugin.decoder<br> httplug.plugin.error<br> httplug.plugin.logger<br> httplug.plugin.redirect<br> httplug.plugin.retry | These are build in plugins that lives in the `php-http/plugins` package. These servcies are not public and may only be used when configure HttpClients or services. 
+| httplug.plugin.content_length <br> httplug.plugin.decoder<br> httplug.plugin.error<br> httplug.plugin.history<br> httplug.plugin.logger<br> httplug.plugin.redirect<br> httplug.plugin.retry | These are built in plugins that lives in the `php-http/plugins` package. These servcies are not public and may only be used when configure HttpClients or services.<br><br>*The History plugin is only available when the kernel runs on debug mode and `httplug.toolbar: true`. It gets automatically injected to all clients.*
```
dbu: s/that lives/that live/

README.md
```diff
@@ -51,14 +51,16 @@ For information how to write applications with the services provided by this bun
...
```
dbu: lets make an explicit note on the history plugin. that plugin needs to be set up with a journal, so you need to define your own service for it.

Resources/config/data-collector.xml
```diff
@@ -0,0 +1,16 @@
...
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="httplug.message_journal" class="Http\HttplugBundle\Collector\MessageJournal" public="false">
+            <tag name="data_collector" template="HttplugBundle::webprofiler.html.twig" priority="200"
+                 id="httplug"/>
+        </service>
+
+        <service id="httplug.plugin.history" class="Http\Client\Plugin\HistoryPlugin" public="false">
```
dbu: lets use a more specific service name here to prevent confusion with generally useable plugins.

DependencyInjection/Configuration.php
```diff
@@ -72,6 +72,7 @@ public function getConfigTreeBuilder()
                         ->scalarNode('stream_factory')->defaultNull()->end()
                     ->end()
                 ->end()
+                ->booleanNode('toolbar')->defaultTrue()->end()
```
dbu: i would make toolbar an arrayNode with canBeEnabled and a message_formatter option to use a custom formatter. its not a plugin configuration for me. the fact that the history plugin is used is an implementation detail. if i configure my own services with history plugin instances, i need my own configuration for formatters anyways.

the other plugins can be configured and then used as they are in any client the user wants to use them.

DependencyInjection/HttplugExtension.php
Nyholm: I'll make this service "internal" instead.. Im using service id `httplug.collector.history_plugin`

README.md
Nyholm: I've made it "internal". Configuration for plugins should be solved by #17

DependencyInjection/Configuration.php
Nyholm: Agreed. But I would like `canBeDisabled` instead... It should be enabled by default.

dbu: oh, you are right, yes. maybe add a description() saying that the toolbar is only active if the debug toolbar itself is active.

Collector/MessageJournal.php
dbu: :+1:

DependencyInjection/HttplugExtension.php
```diff
@@ -27,6 +27,13 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('services.xml');
         $loader->load('plugins.xml');
         $loader->load('discovery.xml');
+        if ($config['toolbar']['enabled'] && $container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug')) {
```
dbu: is this state of the art to detect whether the debug toolbar is active? what if somebody has it active in prod env or deactivated in dev env? i guess in the latter case, we just waste some cpu cycles in dev mode which is no big deal. the other might be a problem. or is debug toolbar hard coupled to kernel.debug?

DependencyInjection/Configuration.php
```diff
@@ -72,8 +71,14 @@ public function getConfigTreeBuilder()
                         ->scalarNode('stream_factory')->defaultNull()->end()
                     ->end()
                 ->end()
-            ->end()
-        ;
+                ->arrayNode('toolbar')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('formatter')->defaultValue('httplug.collector.message_journal')->end()
```
dbu: erm, i think this default value is wrong. i think it should defaultNull and if set it must be a service of a class that implements Http\Message\Formatter - or am i confused?

DependencyInjection/HttplugExtension.php
Nyholm: You are correct. I should redo this. 
What I really want to know is if the service `profiler` exits and if its configuration for `toolbar` is true. 

I'll investigate some more.

DependencyInjection/Configuration.php
Nyholm: No, sorry. Im confused.. You are correct.

DependencyInjection/HttplugExtension.php
dbu: maybe look at what other bundles that plug into the toolbar do

Nyholm: > or is debug toolbar hard coupled to kernel.debug?

No it is not. To get the toolbar you need `framework.profiler` enabled and of course the WebProfilerBundle. There is nothing that ties it to kernel.debug. So this is technically wrong, but it works for most cases. Others are doing the same.

https://github.com/egeloen/IvoryHttpAdapterBundle/blob/master/DependencyInjection/IvoryHttpAdapterExtension.php#L62

I could use a compiler pass and check for the `web_profiler.debug_toolbar.mode` parameter if it is enabled or disabled. But I think that will complicate how I inject the HistoryPlugin because that would also be done in the compiler pass. 

Question: Are we happy with postponing support for the use case where you want the debug Httplug request via the toolbar with `kernel.debug: false`?

DependencyInjection/Configuration.php
Nyholm: Thank you. I've updated the PR

DependencyInjection/HttplugExtension.php
Nyholm: > maybe look at what other bundles that plug into the toolbar do

I've looked at a few. It is common that you inject a logger or a collector to a decorating class. That injection is done in a compiler pass. Some bundles (like [GeocoderBundle](https://github.com/geocoder-php/BazingaGeocoderBundle) do always inject the logger, no matter debug or environment.

sagikazarmark: It's always a good idea to inject logger. Fingers Crossed handler can
decide then when and what to log.

Tobias Nyholm <notifications@github.com> ezt írta (2016. január 8., péntek):

Nyholm: > It's always a good idea to inject logger. Fingers Crossed handler candecide then when and what to log.

Yes, I meant more of a DataCollector to collect data for the profiler. That is injected all the time

DependencyInjection/Configuration.php
```diff
@@ -34,8 +34,7 @@ public function getConfigTreeBuilder()
                     return !empty($v['classes']['client'])
                         || !empty($v['classes']['message_factory'])
                         || !empty($v['classes']['uri_factory'])
-                        || !empty($v['classes']['stream_factory'])
-                    ;
+                        || !empty($v['classes']['stream_factory']);
```
dbu: is this styleci? my idea was that adding another check only does a diff on one line, like trailing comma in arrays.

DependencyInjection/Configuration.php
```diff
@@ -72,8 +71,15 @@ public function getConfigTreeBuilder()
...
-        ;
+                ->arrayNode('toolbar')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                    ->info('Extend the debug profiler with inforation about requests. This could only be active when the kernel is in debug mode.')
+                    ->children()
+                        ->scalarNode('formatter')->defaultNull()->end()
+                    ->end()
+                ->end()
+            ->end();
```
dbu: same logic here, if we add another section, there will be an unnecessary diff on this line.

DependencyInjection/Configuration.php
```diff
@@ -72,8 +71,15 @@ public function getConfigTreeBuilder()
                         ->scalarNode('stream_factory')->defaultNull()->end()
                     ->end()
                 ->end()
-            ->end()
-        ;
+                ->arrayNode('toolbar')
+                    ->canBeDisabled()
```
dbu: regarding kernel.debug: could we make enabled a simple field and default it to the value kernel.debug? then people could just explicitly enable/disable it if they explicitly enabled the toolbar despite kernel.debug being false. if kernel.debug is not available here, we could have a choice of `true`, `false` and `"auto"`, with auto meaning the DI extension classs looks at kernel.debug.

DependencyInjection/HttplugExtension.php
```diff
@@ -27,6 +27,17 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('services.xml');
         $loader->load('plugins.xml');
         $loader->load('discovery.xml');
+
+        if ($config['toolbar']['enabled'] && $s = $container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug')) {
```
dbu: so here you would do something like:

$enabled = is_boolean($config['toolbar']['enabled']) ? $config['toolbar']['enabled'] : $container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug');
if ($enabled) {
...

DependencyInjection/Configuration.php
Nyholm: Yeah, this is styleci

DependencyInjection/Configuration.php
Nyholm: Sounds like a plan

DependencyInjection/HttplugExtension.php
Nyholm: Thanks! I've updated the PR.

README.md
```diff
@@ -51,15 +51,16 @@ For information how to write applications with the services provided by this bun
...
 
 \* *These services are always an alias to another service. You can specify your own service or leave the default, which is the same name with `.default` appended. The default services in turn use the service discovery mechanism to provide the best available implementation. You can specify a class for each of the default services to use instead of discovery, as long as those classes can be instantiated without arguments.*
 
+
 If you need a more custom setup, define the services in your application configuration and specify your service in the `main_alias` section. For example, to add authentication headers, you could define a service that decorates the service `httplug.client.default` with a plugin that injects the authentication headers into the request and configure `httplug.main_alias.client` to the name of your service.
 
 ```yaml
 httplug:
-  clients:
+  clients:
```
sagikazarmark: Is this a trailing space?

Resources/views/webprofiler.html.twig
```diff
@@ -0,0 +1,75 @@
...
+{% block menu %}
+    {# This left-hand menu appears when using the full-screen profiler. #}
+    <span class="label {{ collector.totalRequests == 0 ? 'disabled' }}">
+        <span class="icon">{{ include('@WebProfiler/Icon/ajax.svg') }}</span>
+        <strong>Httplug</strong>
+    </span>
+{% endblock %}
+
+{% block panel %}
+    <h2>Httplug</h2>
```
sagikazarmark: HTTPlug :smile:

Tests/Resources/Fixtures/config/full.php
```diff
@@ -1,16 +1,16 @@
...
-        'uri_factory' => 'my_uri_factory',
-        'stream_factory' => 'my_stream_factory',
-    ),
-    'classes' => array(
-        'client' => 'Http\Adapter\Guzzle6\Client',
+        'uri_factory'     => 'my_uri_factory',
+        'stream_factory'  => 'my_stream_factory',
+    ],
+    'classes' => [
+        'client'          => 'Http\Adapter\Guzzle6\Client',
```
sagikazarmark: Is this styleci as well? AFAIK symfony preset unaligns double arrows, doesn't it?

dbu: Tests is ignored by styleci i think. not sure if we really want to ignore them. consistent CS in the tests sounds like a valuable thing to have.

DependencyInjection/Configuration.php
```diff
@@ -72,8 +71,15 @@ public function getConfigTreeBuilder()
                         ->scalarNode('stream_factory')->defaultNull()->end()
                     ->end()
                 ->end()
-            ->end()
-        ;
+                ->arrayNode('toolbar')
+                    ->addDefaultsIfNotSet()
+                    ->info('Extend the debug profiler with inforation about requests. This could only be active when the kernel is in debug mode.')
```
dbu: lets remove the second phrase

Tests/Resources/Fixtures/config/full.php
sagikazarmark: Ah, I see. Well, the reason why I ignored tests by default was that spec tests are snake cased and function names does not contain visibility. I agree it would make sense here, so let's just remove the test paths from exclusions.

DependencyInjection/Configuration.php
```diff
@@ -72,8 +71,15 @@ public function getConfigTreeBuilder()
...
+                    ->children()
+                        ->enumNode('enabled')->values([true, false, 'auto'])->defaultValue('auto')->end()
```
dbu: and add an info() here saying that by default, the toolbar is active when kernel.debug is true, but you can force it to on or off with this option.

Tests/Resources/Fixtures/config/full.php
Nyholm: Yes, That is styleci. 
Ignoring sounds like a good idea.

Nyholm:
Thank you for your feedback. I've updated the PR.

sagikazarmark:
Sorry, CS fixes probably introduced a few conflicts.

Nyholm:
No worries, I've updated the PR.

sagikazarmark:
Cool. How far is it? I am eager to try it in a project. :smile:

Nyholm:
I'm done with the implementation, I've received a lot feedback and made some improvements. I have addressed all the issues that were brought up. Now Im waiting for @dbu to give this PR the final thumbs up.

sagikazarmark:
Cool.

dbu:
thanks a lot for this one, good job!

sagikazarmark:
:+1:

Nyholm:
Thank you for merging. I'm also very exited to try this in a project =)

DependencyInjection/HttplugExtension.php
```diff
@@ -54,6 +66,10 @@ protected function configureClients(ContainerBuilder $container, array $config)
                 $first = $name;
             }
 
+            if (isset($config['_inject_collector_plugin'])) {
+                array_unshift($arguments['plugins'], 'httplug.collector.history_plugin');
```
dbu: actually, adds the plugin first. does that mean we log before anything else? that will mean that if we have the redirect plugin active, we will not see that a redirect happened. then again, if redirect was activated, we might not care about it.
but it also means we will not log the request exactly as it was sent, with base url or credentials or whatever was added by the plugins. do we maybe need a flag to decide whether to add to the beginning or to the end?

Nyholm: If you have the CachePlugin and put the HistoryPlugin last you will not see anything... But that might be okey, since no request was actually made.

dbu: true. hm. it would be cool to see whether requests where served from
cache or the actual backend. we could try to have two history plugins
and see whether a request goes to the client or not, and also see
whether there is a redirect to serve one answer.

dbu: created #26 for this idea. i guess what we have now is ok, but the other thing sure would be cool.

PR #21, opened 2016-01-08
Update package files

sagikazarmark:


.styleci.yml
```diff
@@ -0,0 +1,11 @@
+preset: symfony
+
+finder:
+    exclude:
+        - "spec"
+        - "tests"
+    path:
+        - "src"
```
dbu: we follow the symfony bundle best practices here, so these paths are different. the code path is ., exclude is Tests. there is no spec folder.

sagikazarmark: spec is there so that we can use the same config in all projects (without thinking about being tested with phpunit or phpspec)

Tests folder: you are right

.styleci.yml
```diff
@@ -0,0 +1,11 @@
+preset: symfony
+
+finder:
+    exclude:
+        - "spec"
+        - "Tests"
+    path:
+        - "src"
```
dbu: source code is in root, not in src

.styleci.yml
```diff
@@ -0,0 +1,11 @@
+preset: symfony
+
+finder:
+    exclude:
+        - "spec"
```
dbu: there is no spec folder. but a Resources folder

.scrutinizer.yml
```diff
@@ -1,11 +1,8 @@
 filter:
```
dbu: do we need both scrutinizer AND styleci? it looks like scrutinizer is not active in thsi repository

sagikazarmark: Scrutinizer does some basic code checks, but I don't insist on it.

dbu: i have nothing against having it. but the PR does not show a scruinizer line

sagikazarmark: I just removed the style checking. Will enable scrutinizer.

dbu:
thanks. did not see any scrutinizer output, though. did you activate it?

sagikazarmark:
Just did

PR #30, opened 2016-01-14
fix dependencies

dbu:
(i hope)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^4.4",
-        "php-http/guzzle6-adapter": "^0.3.1",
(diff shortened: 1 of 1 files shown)


dbu:
i have the impression that `^0.3.1` is not as flexible as we probably think. seems that does not allow 0.4.0. maybe that is the semantics with 0.x versions?

sagikazarmark:
Yes.

^0.3 is translated to >=0.3,<0.4 (BUT: ^1.1 = >=1.1,<2.0)
~0.3 is translated to >=0.3,<1.0

sagikazarmark:
Reasoning: BC break is allowed between MJZ versions, so you might not want to update to a new minor version.

dbu:
makes sense.

okay, and now it installs again, but one of the tests fails. will investigate and update this PR.

PR #32, opened 2016-01-15
fix paths for scrutinizer

dbu:


dbu:
@sagikazarmark any idea why scrutinizer fails here? even if some of the tests fail scrutinizer should still work i think.

sagikazarmark:
@dbu Travis fails and scrutinizer waits for external code coverage. Feel free to merge this, problem will be solved once tests are back to green.

dbu:
but the install does not fail on travis now, its only that one of the
tests failed. does that make scrutinizer think it should not look at
coverage?

sagikazarmark:
Coverage is uploaded by the travis script in the after_success hook. Otherwise coverage is not generated at all. If no coverage is received by scrutinizer in a given time (300 seconds by default I think) then it fails.

dbu:
ah, makes sense. lets hope puli can be fixed soon

PR #35, opened 2016-01-15
Make sure we have a default config for the clients variable node

Nyholm:
If you configure the bundle with: 

```yaml
httplug:
  clients:
    guzzle:
```

You would get an undefined array key error because we have no default value for the `config` node. This PR fix that issue.


sagikazarmark:
:+1:

Issue #26, opened 2016-01-11
detailed reporting in toolbar


dbu:
Right now, we add the reporting at the begin of the plugin chain. This means we don't see the exact request as it was sent to the server, and we dont see whether a request is answered from cache or not, and we will miss redirects.

We could use a second plugin at the end of the chain, and correlate the requests to determine: request going to actual HTTP client (and thus to real server) / request leading to redirect / request served from cache.

See also https://github.com/php-http/HttplugBundle/pull/19/files#r49300763

Another thing is that we should identify which client was used, in the case where there is more than one client that reports requests. (follow up of #36)

sagikazarmark:
This is something that would only make sense here, so the plugin itself could be in this repository, isn't it?

Nyholm:
I do not think we need a new plugin class. We just need a new HistoryPlugin service and add that last in the chain. 

We could maybe do something cool when displaying this.. Like a git diff to show what was changed by the plugin chain.

sagikazarmark:
Diff sounds cool, but unsure about the usefulness in this case...but I am not against it.

Also, see php-http/httplug#105 for diff stuff

Nyholm:
I suggest we add two collectors per client. One at the beginning of the plugin queue and one at the end. We want to do it per client because we want to show that client X sent this request and client Y sent that request.

sagikazarmark:
Why would we need two collectors.

We want to add something to the beginning and to the end of a queue where the order matters. It sounds familiat, doesn't it? (Plugin chain)

Maybe we can somehow pass the client to the journal as well to identify the sender?

joelwurtz:
Or it could be a PluginClientDebug whichs extends the PluginClient one to replace the plugin chain and log the total stack trace of a request (go in this plugin, then in this one, then in this client, ...)

Nyholm:
I can't see how we could pass an identifier to the journal/history plugin to be used every second time. That's why I suggested two collectors. 

I like @joelwurtz idea with a PluginClientDebug. That would make a lot of things easier to extract. Like which plugins are used with which clients. It could also be easier to inject the data collector and also a stopwatch(?).

sagikazarmark:
I also like the idea. But keep in mind that this debug client must not have any more features than debug. I mean, the behavior should be transparent: logging should also be done in prod mode in the "ordinary" plugin client with the configured logger. (BTW: is it possible right now to configure multiple logger plugins with different loggers (channels)?)

[...]

PR #36, opened 2016-01-15
Add history plugin to the auto discovery client

Nyholm:
If you do not configure httplug at all it will use the auto discovery client. That client should also have the history plugin injected if you use the toolbar. 

Im not 100% confident in this PR. We will as always have two identifiers for the default client: `httplug.client` and `httplug.client.default`. This PR makes `httplug.client` a plugin client but not the other. 

Is this okey? The implications would be that you do not see the toolbar when using no configuration and the `httplug.client.default` service id.

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -2,6 +2,7 @@
 
 namespace Http\HttplugBundle\DependencyInjection;
 
+use Http\Client\Plugin\PluginClient;
 use Http\HttplugBundle\ClientFactory\DummyClient;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -88,6 +89,11 @@ class HttplugExtension extends Extension
         // Alias the first client to httplug.client.default
         if ($first !== null) {
             $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
+        } elseif (isset($config['_inject_collector_plugin'])) {
+            // No client was configured. Make sure to inject history plugin to the auto discovery client.
+            $container->register('httplug.client', PluginClient::class)
+                ->addArgument(new Reference('httplug.client.default'))
+                ->addArgument([new Reference('httplug.collector.history_plugin')]);
         }
     }
 }


dbu:
i guess this is not wrong per se, as the .default is just an alias. but i wonder if we should not handle multiple clients and show multiple icons with tooltip per client service name? or log all requests to one common collection and allow to filter by client service name?

Nyholm:
> i guess this is not wrong per se, as the .default is just an alias. 

No, the `httplug.client.default` is the service. `httplug.client` is the alias. `httplug.client.default` will by the default be the auto discovery client and `httplug.client` will be the first configured client or the configured client named `default`. 

> but i wonder if we should not handle multiple clients and show multiple icons with tooltip per client service name? or log all requests to one common collection and allow to filter by client service name?

I do not think we should have multiple toolbar icons. But I do think we should requests and responses per client on the profiler page and in the toolbar pop up. 

I can write another PR about that.

sagikazarmark:
> multiple clients and show multiple icons

Not sure if multiple icons is a good idea. My toolbar is big enough already :stuck_out_tongue_closed_eyes: 

I agree with @Nyholm, it makes sense on the profiler page. (Don't know how doctrine does this with multiple entity managers, maybe we could look at that)

DependencyInjection/HttplugExtension.php
```diff
@@ -88,6 +89,11 @@ protected function configureClients(ContainerBuilder $container, array $config)
         // Alias the first client to httplug.client.default
         if ($first !== null) {
             $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
+        } elseif (isset($config['_inject_collector_plugin'])) {
+            // No client was configured. Make sure to inject history plugin to the auto discovery client.
+            $container->register('httplug.client', PluginClient::class)
```
dbu: was `httplug.client` not available at all before? or did this happen somewhere else? i am a bit confused by this.

dbu:
right, makes more sense to have one report and allow to filter by client. i guess that overlaps somewhat with #26 as well. we don't need to fix this in this PR, lets just make the incremental improvement of not missing out on the auto discovery client.

DependencyInjection/HttplugExtension.php
Nyholm: `httplug.client` was an alias before this line. This register it as a service.

Nyholm:
Yeah, I agree that there is some overlapping.

dbu:
thanks. i added a note to #26 in the description.

PR #37, opened 2016-01-17
Configure plugins

### LINKED ISSUE #17: Configure default plugins

Nyholm:
This PR fixes #17 

```yaml
httplug: 
  plugins: 
    cache:
      cache_pool: 'acme.cache'
      config: 
        respect_headers: true
    logger:
      formatter: 'acme.formatter'
```

This make sure that you can configure all our 9 configurable plugins. Some plugins (like `RetryPlugin`) has only optional configuration. They are enabled by default. They others are disabled by default. 

Also, the `LoggerPlugin` is treated special. We inject the `logger` service automatically and it is enabled by default. This is done because of ease of use.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -15,6 +15,7 @@ use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
  * sections are normalized, and merged.
  *
  * @author David Buchmann <mail@davidbu.ch>
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
 class Configuration implements ConfigurationInterface
 {
@@ -27,6 +28,7 @@ class Configuration implements ConfigurationInterface
         $rootNode = $treeBuilder->root('httplug');
 
         $this->configureClients($rootNode);
+        $this->configurePlugins($rootNode);
 
         $rootNode
             ->validate()
@@ -107,5 +109,130 @@ class Configuration implements ConfigurationInterface
                     ->variableNode('config')->defaultValue([])->end()
                 ->end()
             ->end();
+    }
+
+    /**
+     * @param ArrayNodeDefinition $root
+     */
+    protected function configurePlugins(ArrayNodeDefinition $root)
+    {
+        $root->children()
+            ->arrayNode('plugins')
+                ->addDefaultsIfNotSet()
+                ->children()
+
+                    ->arrayNode('authentication')
+                    ->canBeEnabled()
+                        ->children()
+                            ->scalarNode('authentication')
+                                ->info('This must be a service id to a service implementing Http\Message\Authentication')
+                                ->isRequired()
+                                ->cannotBeEmpty()
+                            ->end()
+                        ->end()
+                    ->end() // End authentication plugin
+
+                    ->arrayNode('cache')
+                    ->canBeEnabled()
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('cache_pool')
+                                ->info('This must be a service id to a service implementing Psr\Cache\CacheItemPoolInterface')
+                                ->isRequired()
+                                ->cannotBeEmpty()
+                            ->end()
+                            ->scalarNode('stream_factory')
+                                ->info('This must be a service id to a service implementing Http\Message\StreamFactory')
+                                ->defaultValue('httplug.stream_factory')
+                                ->cannotBeEmpty()
+                            ->end()
+                            ->arrayNode('config')
+                                ->addDefaultsIfNotSet()
+                                ->children()
+                                    ->scalarNode('default_ttl')->defaultNull()->end()
+                                    ->scalarNode('respect_cache_headers')->defaultTrue()->end()
+                                ->end()
+                            ->end()
+                        ->end()
+                    ->end() // End cache plugin
+
+                    ->arrayNode('cookie')
+                    ->canBeEnabled()
+                        ->children()
+                            ->scalarNode('cookie_jar')
+                                ->info('This must be a service id to a service implementing Http\Message\CookieJar')
+                                ->isRequired()
+                                ->cannotBeEmpty()
+                            ->end()
+                        ->end()
+                    ->end() // End cookie plugin
+
+                    ->arrayNode('decoder')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('use_content_encoding')->defaultTrue()->end()
+                        ->end()
+                    ->end() // End decoder plugin
+
+                    ->arrayNode('history')
+                    ->canBeEnabled()
+                        ->children()
+                            ->scalarNode('journal')
+                                ->info('This must be a service id to a service implementing Http\Client\Plugin\Journal')
+                                ->isRequired()
+                                ->cannotBeEmpty()
+                            ->end()
+                        ->end()
+                    ->end() // End history plugin
+
+                    ->arrayNode('logger')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('logger')
+                                ->info('This must be a service id to a service implementing Psr\Log\LoggerInterface')
+                                ->defaultValue('logger')
+                                ->cannotBeEmpty()
+                            ->end()
+                            ->scalarNode('formatter')
+                                ->info('This must be a service id to a service implementing Http\Message\Formatter')
+                                ->defaultNull()
+                            ->end()
+                        ->end()
+                    ->end() // End logger plugin
+
+                    ->arrayNode('redirect')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('preserve_header')->defaultTrue()->end()
+                            ->scalarNode('use_default_for_multiple')->defaultTrue()->end()
+                        ->end()
+                    ->end() // End redirect plugin
+
+                    ->arrayNode('retry')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('retry')->defaultValue(1)->end()
+                        ->end()
+                    ->end() // End retry plugin
+
+                    ->arrayNode('stopwatch')
+                    ->canBeDisabled()
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('stopwatch')
+                                ->info('This must be a service id to a service extending Symfony\Component\Stopwatch\Stopwatch')
+                                ->defaultValue('debug.stopwatch')
+                                ->cannotBeEmpty()
+                            ->end()
+                        ->end()
+                    ->end() // End stopwatch plugin
+
+                ->end()
+            ->end()
+        ->end();
     }
 }
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -6,12 +6,14 @@ use Http\Client\Plugin\PluginClient;
 use Http\HttplugBundle\ClientFactory\DummyClient;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 
 /**
  * @author David Buchmann <mail@davidbu.ch>
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
 class HttplugExtension extends Extension
 {
@@ -50,6 +52,7 @@ class HttplugExtension extends Extension
         foreach ($config['main_alias'] as $type => $id) {
             $container->setAlias(sprintf('httplug.%s', $type), $id);
         }
+        $this->configurePlugins($container, $config['plugins']);
         $this->configureClients($container, $config);
     }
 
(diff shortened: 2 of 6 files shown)


Nyholm:
> Also, the LoggerPlugin is treated special. We inject the logger service automatically and it is enabled by default. This is done because of ease of use.

I'm having second thoughts about this. It would mean that a Symfony installation that does not have Monolog would get an error like "httplug.plugin.logger has a dependency on a service 'logger'. No such service found". 

I guess that it would be less "magic" if we removed it and let the user configure the logger like any other plugin with dependencies.

joelwurtz:
Maybe we can also enabled the stopwatch plugin based on the kernel.debug value by default ?

Nyholm:
> Maybe we can also enabled the stopwatch plugin based on the kernel.debug value by default ?

Same problem there as with the logger. What if the stopwatch service does not exist?

Nyholm:
I've updated the PR. The logger is not treated any special

joelwurtz:
> Same problem there as with the logger. What if the stopwatch service does not exist?

I don't think this is too much of a problem, we can just throw an exception if service is not available. IMO if someone have desactived the stopwatch and log service in his debug mode / application he will know how to change the configuration :)

sagikazarmark:
What if we only enable them automatically if:
- the user haven't enabled it manually
- the relevant service is available

Nyholm:
hm, sure. I can give the logging plugin and the stopwatch plugin its dependencies in a compiler pass (if the dependency services exists). That would make sense.

DependencyInjection/Configuration.php
```diff
@@ -108,4 +109,95 @@ protected function configureClients(ArrayNodeDefinition $root)
...
+    {
+        $root->children()
+            ->arrayNode('plugins')
+                ->addDefaultsIfNotSet()
+                ->children()
+
+                    ->arrayNode('authentication')
+                    ->canBeEnabled()
+                        ->children()
+                            ->scalarNode('authentication')->isRequired()->cannotBeEmpty()->end()
```
dbu: can we add an info to this? this is the name of the authentication service that implements some php-http interface?

DependencyInjection/Configuration.php
```diff
@@ -108,4 +109,95 @@ protected function configureClients(ArrayNodeDefinition $root)
...
+                                    ->scalarNode('respect_cache_headers')->defaultTrue()->end()
+                                ->end()
+                            ->end()
+                        ->end()
+                    ->end() // End cache plugin
+
+                    ->arrayNode('cookie')
+                    ->canBeEnabled()
+                        ->children()
+                            ->scalarNode('cookie_jar')->isRequired()->cannotBeEmpty()->end()
```
dbu: same here, what service is this? could we provide a default service as well here?

DependencyInjection/Configuration.php
```diff
@@ -108,4 +109,95 @@ protected function configureClients(ArrayNodeDefinition $root)
...
+                    ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('use_content_encoding')->defaultTrue()->end()
+                        ->end()
+                    ->end() // End decoder plugin
+
+                    ->arrayNode('history')
+                    ->canBeEnabled()
+                        ->children()
+                            ->scalarNode('journal')->isRequired()->cannotBeEmpty()->end()
```
dbu: does this make sense? why not let the user define both the journal and the history plugin service and enable the plugin themselves?

DependencyInjection/HttplugExtension.php
```diff
@@ -90,4 +92,66 @@ protected function configureClients(ContainerBuilder $container, array $config)
...
+     * @param ContainerBuilder $container
+     * @param array            $config
+     */
+    protected function configurePlugins(ContainerBuilder $container, array $config)
+    {
+        foreach ($config as $name => $pluginConfig) {
+            $pluginId = 'httplug.plugin.'.$name;
+            if ($pluginConfig['enabled']) {
+                $def = $container->getDefinition($pluginId);
+                $this->configurePluginByName($name, $def, $pluginConfig);
```
dbu: or configurePlugin$name() and split the code below into a method per case?

dbu:
cool! i like the idea with the compiler pass. we could say: if explicitly disabled, we remove the service definition in the di class. if explicitly enabled, we somehow tell the compiler pass to explode if stopwatch/logger are not available. 

the default behaviour should be that if the requirements are met, we enable those plugin, otherwise we automatically disable them. would need some container parameter i fear. and injecting the optoinal plugins to the client services in the compiler pass. hm. which will lead to priorisation challenges.

DependencyInjection/Configuration.php
Nyholm: Yeah, This is too much of a replacement for writing a few lines of service configuration yourself. But what does the user expect? That there is config options for all but 2 plugins? 

I think it somehow makes sens to do like this because it is coherent.

DependencyInjection/HttplugExtension.php
Nyholm: I was considering that approach. Switch statements are code smell but dynamic functions like that is worse. Also, you cant type hint dynamic functions in PHP7

Nyholm:
> the default behaviour should be that if the requirements are met, we enable those plugin, otherwise we automatically disable them. would need some container parameter i fear. and injecting the optoinal plugins to the client services in the compiler pass. hm. which will lead to priorisation challenges.

I might be able to do it without ana container parameters. I'll give it some thoughts after work.

DependencyInjection/HttplugExtension.php
dbu: okay. lets keep this then. its very internal either way so no big deal.

DependencyInjection/HttplugExtension.php
```diff
@@ -90,4 +92,66 @@ protected function configureClients(ContainerBuilder $container, array $config)
             $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
         }
     }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param array            $config
+     */
+    protected function configurePlugins(ContainerBuilder $container, array $config)
```
dbu: why not private?

DependencyInjection/Configuration.php
dbu: the debug toolbar history is auto-injected into every client we know of, when we determine that the toolbar is wanted. what is the use for this plugin? i would have to add it manually, right? i don't see the use of having it in the configuration like this, and would prefer dropping it and explaining in the documentation why its not in there.

i think this would just clutter the configuration with things that won't be used much, if at all.

dbu: what other plugin are you talking about btw? i think all the others make a lot of sense.

DependencyInjection/HttplugExtension.php
Nyholm: Fixed!

DependencyInjection/Configuration.php
Nyholm: Added info() on all nodes that require a service.

Nyholm:
@dbu mentioned an issue in https://github.com/php-http/HttplugBundle/pull/37#discussion_r50282114

> does this make sense? why not let the user define both the journal and the history plugin service and enable the plugin themselves?

> the debug toolbar history is auto-injected into every client we know of, when we determine that the toolbar is wanted. what is the use for this plugin? i would have to add it manually, right? i don't see the use of having it in the configuration like this, and would prefer dropping it and explaining in the documentation why its not in there.

> i think this would just clutter the configuration with things that won't be used much, if at all.

First of, we do not use `httplug.plugin.history` for the toolbar. We register our own `httplug.collector.history_plugin`. The `httplug.plugin.*` are for the users. 

You do highlight the fact that the following two way of using the cache plugins are exactly the same: 

Option A)
```yaml
// config.yml
httplug:
  plugins:
    cache:
      cache_pool: 'cache'
  clients:
    acme:
      plugins: ['httplug.plugin.cache']
```

Option B) 
```yaml
// services.yml
services: 
  my_cache_plugin:
    class: Http\Client\Plugin\CachePlugin
    arguments: ["@cache"]
```
```yaml
// config.yml
httplug:
  clients:
    acme:
      plugins: ['my_cache_plugin']
```

**B** is the generic approach that will always work and is suggested that you do with your custom made plugins. 

**A** makes much more sense for plugins like decoder, retry and redirect because they only have scalar dependencies. In fact, option **B** would be unnecessarily complicated for such plugins. 

I believe that we should allow these two options for configuring all plugins because it would be weird to support option **A** for only a subset of our plugins.

DependencyInjection/CompilerPass/PluginPass.php
```diff
@@ -0,0 +1,28 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        if (!($container->hasDefinition('logger') || $container->hasAlias('logger'))) {
+            $container->removeDefinition('httplug.plugin.logger');
+        }
+
+        if (!$container->hasDefinition('debug.stopwatch')) {
+            $container->removeDefinition('httplug.plugin.stopwatch');
+        }
+    }
```
Nyholm: With this compiler pass we make sure that the exception messages makes sense. 

Before:
If logger plugin was or wasn't used and `logger` is not defined you would get an exception like "httplug.plugin.logger" requested an non existing service 'logger'". 

Now:
If logger plugin is used and `logger` is missing you will get an exception like "httplug.client.acme has requested an non existing service 'httplug.plugin.logger'"

dbu: hm, i wonder if we could instead detect the problem in the compiler pass and throw the exception "to use the logger plugin, you need to have a logger service configured"

DependencyInjection/CompilerPass/PluginPass.php
```diff
@@ -0,0 +1,28 @@
...
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class PluginPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!($container->hasDefinition('logger') || $container->hasAlias('logger'))) {
```
dbu: this is not really correct. its possible to configure a different logger, at which point this is no longer needed. i think we should use a container parameter to know which logger service must exist.

dbu:
okay, agree on consistency and allowing to configure all standard plugins.

DependencyInjection/CompilerPass/PluginPass.php
Nyholm: That would require us to know what plugins that are being used. We have currently no clue about that in the compiler pass. It could be solved with container parameters or a DebugPluginClient::getPlugins(). 

You could also make an argument that the very few users that have logger/stopwatch disabled probably knows about it and are skilled enough to figure out that the LoggerPlugin requires a logger. But now I am making assumptions based on no facts.

DependencyInjection/CompilerPass/PluginPass.php
Nyholm: Oh, sorry. Yeah you are correct... Hm..

Nyholm: I would like to avoid container parameters if we could. 

We could fetch the `httplug.plugin.logger` definition and look at it's dependencies and validate if they are equal to the default value and then check if they exists. But that does not solve the issue you brought up in https://github.com/php-http/HttplugBundle/pull/37/files#r50537858

Hm, we maybe need to use container parameters anyway.

dbu: while i agree that theoretically its better to avoid container
parameters, i think the value of clear understandable error messages has
a lot of real value.

we might also find a way to communicate the used plugins directly to the
compiler pass (keep a reference in the bundle class and give it to the
di extension so it can update the compiler pass instance?)

Nyholm:
> With this compiler pass we make sure that the exception messages makes sense.

>Before:
If logger plugin was or wasn't used and logger is not defined you would get an exception like "httplug.plugin.logger" requested an non existing service 'logger'".

>Now:
If logger plugin is used and logger is missing you will get an exception like "httplug.client.acme has requested an non existing service 'httplug.plugin.logger'"


Im happy to say that I was wrong.
If you do not have a service with id `logger` and if you are not using or logging plugin no error will show. The logging plugin is private so it will be removed if not used when we build the container. 

If, however, I use the logging plugin (still with no `logger` I will get an exception like this: 

```
[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]              
  The service "httplug.client.acme" has a dependency on a non-existent service "logger". 
```

```yaml
// config.yml
httplug:
  clients:
    acme:
      plugins: ['httplug.plugin.logger']
```

---- 
No compiler pass needed =)

dbu:
looks good! 
@Nyholm can you please adjust the configuration phpunit test? it does not expect the plugins section of the configuration: https://travis-ci.org/php-http/HttplugBundle/jobs/105014582

Nyholm:
@dbu, I have addressed all the comments and feedback. I think this is ready to merge.

sagikazarmark:
Do you think we can tag a new release after this PR? Either 1.0 or a new MJZ. I don't want to rush with the stable if you think there is more work to be done.

Nyholm:
I do not think this is ready for a stable release yet. There are some changes that we want to do on the issue tracker. Also, I have not been running this in production yet. I want to make sure it is no major bugs and that concept/API/config makes sense when adding more third party bundles.

sagikazarmark:
Okay. Although I already use it in production :smile:

dbu:
thanks a lot!

i created #38 for the release discussion

PR #44, opened 2016-02-25
Updated the authentication configuration to support credentials

### LINKED ISSUE #42: Add configuration for autentication
I would like to configure different clients with authentication. 

```yaml
httplug:
  clients:
    guzzle5:
      factory: 'httplug.factory.guzzle5'
      plugins: ['httplug.plugin.logger']
    trans_logger:
      factory: 'httplug.factory.guzzle5'
      authentication: 
        type: 'basic'
        username: 'foo'
        password: 'bar'
    microservice2:
      factory: 'httplug.factory.guzzle5'
      plugins: ['httplug.plugin.logger', 'my_cache_plugin']
      authentication: 
        type: 'wsse'
        username: 'baz'
        password: 'biz'
```

What do you think? Is this something we should support? Do you like the config example?

Nyholm:
This PR fixes #42 

```yaml
// config.yml
httplug:
    plugins:
        authentication:
            my_basic:
                type: 'basic'
                username: 'foo'
                password: 'bar'
            my_wsse:
                type: 'wsse'
                username: 'foo'
                password: 'bar'
            my_brearer:
                type: 'bearer'
                token: 'foo'

    clients:
        acme:
            plugins: ['httplug.plugin.authentication.my_wsse']
```

This PR will break BC because change of the configuration in `httplug.plugins.authentication`.


DependencyInjection/Configuration.php
```diff
@@ -120,14 +120,18 @@ protected function configurePlugins(ArrayNodeDefinition $root)
...
+                        ->prototype('array')
+                            ->children()
+                                ->enumNode('type')
+                                    ->values(['basic', 'bearer', 'wsse'])
+                                    ->isRequired()
+                                    ->cannotBeEmpty()
+                                ->end()
+                                ->scalarNode('username')->end()
+                                ->scalarNode('password')->end()
+                                ->scalarNode('token')->end()
```
dbu: we could go all fancy with custom validation to have the right required parameters based on the type. but we don't need that now.

i wonder if we should add type `service` that accepts a scalarNode `authentication_service` to keep supporting custom authentication services?

DependencyInjection/HttplugExtension.php
```diff
@@ -162,4 +168,53 @@ private function configurePluginByName($name, Definition $definition, array $con
...
+     */
+    private function configureAuthentication(ContainerBuilder $container, array $config)
+    {
+        foreach ($config as $name => $values) {
+            $authServiceKey = sprintf('httplug.plugin.authentication.%s.auth', $name);
+            if ($values['type'] === 'bearer') {
+                $this->validateAuthenticationConfiguration($values, 'bearer', $name, ['token']);
+                $container->register($authServiceKey, Bearer::class)
+                    ->addArgument($values['token']);
+            } elseif ($values['type'] === 'basic') {
```
dbu: i would use a switch statement for this.

README.md
```diff
@@ -134,6 +134,31 @@ httpug:
...
+                type: 'basic'
+                username: 'foo'
+                password: 'bar'
+            my_wsse:
+                type: 'wsse'
+                username: 'foo'
+                password: 'bar'
+            my_brearer:
+                type: 'bearer'
+                token: 'foo'
```
dbu: can we use something else than "foo" here to make it visually different? maybe 'hash' ?

DependencyInjection/HttplugExtension.php
```diff
@@ -162,4 +168,53 @@ private function configurePluginByName($name, Definition $definition, array $con
...
+
+    /**
+     * Throw a Logic error with a descriptive error if some configuration is missing.
+     *
+     * @param array  $config
+     * @param string $authType
+     * @param string $type
+     * @param array  $names
+     */
+    private function validateAuthenticationConfiguration(array $config, $type, $authType, array $names)
```
dbu: ok, as we need the validation anyways, could we move it to the Configuration class? an example of how this looks would be https://github.com/doctrine/DoctrinePHPCRBundle/blob/master/DependencyInjection/Configuration.php#L133-L152

dbu:
perfect, i think this configuration looks as simple and straightforward as it can be!

DependencyInjection/HttplugExtension.php
Nyholm: Great! Thank you for this example. I will remove the `validateAuthenticationConfiguration` and use something like this instead.

DependencyInjection/Configuration.php
Nyholm: > i wonder if we should add type service that accepts a scalarNode authentication_service to keep supporting custom authentication services?

Sure!

sagikazarmark: Not sure. People can still set up authentication plugins, can't they?

Nyholm: If we add type `service` then you could use the authentication plugin just as before. It just require you to do a minor config change. 

Before: 
```yaml
// config.yml
httplug:
    plugins:
        authentication:
             authentication: 'my_authentication_service'
    clients:
        my_client:
            plugins: ['httplug.plugin.authentication']
```

After:
```yaml
// config.yml
httplug:
    plugins:
        authentication:
            acme:
                type: 'service'
                service: 'my_authentication_service'
    clients:
        my_client:
            plugins: ['httplug.plugin.authentication.acme']
```

Nyholm:
Thank you for the feedback. I've addressed your comments.

DependencyInjection/Configuration.php
```diff
@@ -235,4 +225,62 @@ protected function configurePlugins(ArrayNodeDefinition $root)
...
+                    ->always()
+                    ->then(function ($config) {
+                        switch ($config['type']) {
+                            case 'basic':
+                                if (empty($config['username']) || empty($config['password'])) {
+                                    throw new InvalidConfigurationException('Authentication "basic" requires both "username" and "password".');
+                                }
+                                break;
+                            case 'bearer':
+                                if (empty($config['bearer'])) {
```
dbu: s/bearer/token/

DependencyInjection/Configuration.php
```diff
@@ -235,4 +225,62 @@ protected function configurePlugins(ArrayNodeDefinition $root)
...
+            ->useAttributeAsKey('name')
+            ->prototype('array')
+                ->validate()
+                    ->always()
+                    ->then(function ($config) {
+                        switch ($config['type']) {
+                            case 'basic':
+                                if (empty($config['username']) || empty($config['password'])) {
+                                    throw new InvalidConfigurationException('Authentication "basic" requires both "username" and "password".');
+                                }
```
dbu: should we also complain about fields that should *not* be set? maybe too strict though.

Nyholm: Good question. I believe it is not a required feature but maybe a *nice to have*. 
Maybe we should do something more sophisticated then. Do you know if it is possible to conditionally add nodes?

Like "If $config['type'] == 'basic', add node 'username' and 'password'"

dbu: the nodes happen before the configuration comes. so i think we would need to check `if (!empty(...)`

Nyholm: > the nodes happen before the configuration comes. so i think we would need to check if (!empty(...)

Sorry, not too sure I understand what you mean.

dbu: sorry. what i meant is: the code creating the configuration nodes is executed without knowledge of the concrete configuration. only the callback of the `validate()` can know the actual configuration, so i don't think we can define the nodes conditionally.

dbu: created #48 to show how i propose to do this

Nyholm:
I've merged PR #48 and I think this PR is ready for merge into master.

sagikazarmark:
A squash would be a good idea first.

Nyholm:
Of course, Sorry @sagikazarmark. 
They are now squashed.

dbu:
great job, thanks a lot!

Nyholm:
Thank you for the input!

PR #45, opened 2016-02-25
Bugfix, circular reference with client named 'default'

Nyholm:
If we named a client to "default" we would get a circular reference because we are trying to add an alias named `httplug.client.default` for service `httplug.client.default`.


Nyholm:
Thank you!

dbu:
good catch! i was worried for a moment whether calling a later client than the first 'default' but actually $first is set to 'default' if any client is called that.

PR #46, opened 2016-02-25
Added example how to use the bundle without auto discovery

Nyholm:
I've got [some feedback](https://github.com/Happyr/TranslationBundle/pull/23#issuecomment-188829279) that documentation was missing about how to use this bundle without auto discovery. 

This PR adds a short explanation and example configuration.


dbu:
:+1: thanks!

PR #47, opened 2016-02-26
Updated badges

Nyholm:
The badges from Shield does not seam to work. This PR will fix that. 

screen shot 2016-02-26 at 08 57 30

TODO: Enable code coverage in scrutinizer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # HttplugBundle
 
-[![Latest Version](https://img.shields.io/github/release/php-http/HttplugBundle.svg?style=flat-square)](https://github.com/php-http/HttplugBundle/releases)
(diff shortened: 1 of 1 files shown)


sagikazarmark:
I am not a big fan of this. I prefer having the flat style everywhere and shields provides that. It is working for me now perfectly fine.

What we could do is replacing github version with packagist version as Github API seems to be the bottleneck here:

badges/shields#617
badges/shields#628

Nyholm:
Sorry, I did not notice different styles of the badges. 

Shields.io has been working poorly for me for a couple of months and this repo has never shown (what **I** have seen) anything other that the print screen.

If you see the badges as they are supposed to be shown then I hope it is only me, I will investigate further.

sagikazarmark:
I also had this issue a couple of months ago and I found that it is mostly github what causes the problem.

PR #48, opened 2016-02-26
validate type specific options

dbu:
@Nyholm trying to do this validation thing


DependencyInjection/Configuration.php
```diff
@@ -283,4 +275,19 @@ private function addAuthenticationPluiginNode()
 
         return $node;
     }
+
+    public static function validateAuthenticationType($expected, $actual, $authName)
+    {
+        unset($actual['type']);
+        if ($expected == $actual) {
```
Nyholm: Make a comment that you actually intend to do == instead of ===.

Nyholm:
Cool. I like it. =)

DependencyInjection/Configuration.php
```diff
@@ -243,24 +243,16 @@ private function addAuthenticationPluiginNode()
                     ->then(function ($config) {
                         switch ($config['type']) {
                             case 'basic':
-                                if (empty($config['username']) || empty($config['password'])) {
-                                    throw new InvalidConfigurationException('Authentication "basic" requires both "username" and "password".');
-                                }
+                                Configuration::validateAuthenticationType(['username', 'password'], array_keys($config), 'basic');
```
Nyholm: Should be `self::validateAuthenticationType` right?

sagikazarmark: Stupid question: since it is a closure, why can't it be `$this->validateAuthenticationType`?

dbu: i was not sure with closure and scope, to be honest ;-)
will try $this->validate... can that method then even be private?

Nyholm: Using $this in a closure has been supported since php 5.4. I just tried this: https://3v4l.org/gvIgS

dbu:
fixed to use $this, and added the missing tests for the new configuration (and plugin configuration in general)

Nyholm:
Cool. Thanks!

PR #51, opened 2016-02-28
only cache downloaded packages

xabbuh:
Caching the entire Composer cache makes Travis to cache Packagist metadata too which are much more likely to make the cache stale too often (while downloaded package archives could still be cached).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 cache:
     directories:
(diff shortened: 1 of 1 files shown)


sagikazarmark:
Thanks. Can you make a PR against our boilerplate repo as well?

xabbuh:
@sagikazarmark done

sagikazarmark:
Thanks

PR #53, opened 2016-02-28
Make sure travis can run on HHVM

### LINKED ISSUE #14: investigate hhvm build failure
investigate why hhvm builds fail and fix them. we started ignoring the fail in #11

Nyholm:
This PR makes sure to include a different config file for the functional test when running HHVM. The new config file will not use auto discovery. 

This will fix #14


dbu:
awesome!

PR #54, opened 2016-02-29
[POC] how to not be dependent on discovery

Nyholm:
This is just a proof of concept how you could do to not be dependent on discovery. This will make the dependency to puli or puli bundle optional

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -46,12 +46,21 @@ class HttplugExtension extends Extension
             }
         }
 
+        $useDiscovery = false;
         foreach ($config['classes'] as $service => $class) {
             if (!empty($class)) {
                 $container->removeDefinition(sprintf('httplug.%s.default', $service));
                 $container->register(sprintf('httplug.%s.default', $service), $class);
+            } else {
+                // we have to use discovery to find this class
+                $useDiscovery = true;
             }
         }
+
+        if ($useDiscovery) {
+            $this->verifyDiscoveryInstalled($container);
+        }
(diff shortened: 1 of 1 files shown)


dbu:
oh, that makes sense to me!

DependencyInjection/HttplugExtension.php
```diff
@@ -207,4 +216,25 @@ private function configureAuthentication(ContainerBuilder $container, array $con
...
+
+    /**
+     * Verify that Puli is installed
+     *
+     * @param ContainerBuilder $container
+     * @throws \Exception
+     */
+    private function verifyDiscoveryInstalled(ContainerBuilder $container)
+    {
+        $enabledBundles = $container->getParameter('kernel.bundles');
```
dbu: i guess we need to fake some things for the unit tests here. we can then also test the error situation.

sagikazarmark:
For me this is the same as the interface packages with one more downside: this way you have to manually install the Puli composer plugin AND the symfony bundle as well.

While the composer plugin is not necessary if you want to use puli, you can manually configure any kind of client, the symfony bundle is ALWAYS necessary in order to use Puli.

sagikazarmark:
Hm, it seems I was wrong. The symfony-bundle actually requires the composer plugin. In this case, it's enough to suggest and install the symfony bundle.

DependencyInjection/HttplugExtension.php
sagikazarmark: I don't think we need to check against the bundle, we can check against the service as well. I think that is easier to mock.

PR #55, opened 2016-02-29
[WIP] Remove discovery dependency

sagikazarmark:
Includes #54 
Unfortunately I couldn't fix functional tests, although I added the symfony bundle to the test AppKernel.

Any ideas?


sagikazarmark:
Also, I am open for better names, I just picked factory in order to clearly separate from discovery.

DependencyInjection/HttplugExtension.php
```diff
@@ -207,4 +217,20 @@ private function configureAuthentication(ContainerBuilder $container, array $con
...
+    /**
+     * Verifies that Puli bundle is installed.
+     *
+     * @param ContainerBuilder $container
+     *
+     * @throws \Exception
+     */
+    private function verifyPuliInstalled(ContainerBuilder $container)
+    {
+        if (false === $container->has('puli.discovery')) {
```
Nyholm: Since `puli.discovery` is part of another bundle this if statement will always be true. This code has to be moved to a compiler pass, or you could check if the bundle is enabled.

sagikazarmark: I am not sure I understand why it would be true.

Although I am not against moving this to a compiler pass. Can we load a compiler pass from the extension where we check the necessity of discovery usage? Or should we move the whole discovery stuff to a compiler pass?

Nyholm: In a bundle's extension class you can not see services registered in other bundles. That is why `$container->has('puli.discovery')` will always be false. 

I suggest to save the $useDiscovery variable in a container parameter and then handle the rest in a compiler pass. 

If we check if discovery is necessary in a compiler pass we need to inspect `httplug.client.default`, `httplug.message_factory.default` etc to see if they are using the standard class or not. That may be more elegant because we do not use container parameters but I believe it requires more hard coded values.

sagikazarmark: What if we do the following: remove the fallback services and move them to a compiler pass. Then we can check in the compiler if all the services has been defined in the extension or not? (This requires to know in what order are the compilers and extensions are loaded). If any of the services are not registered, register it with the factory (and register the factory as well).

Nyholm: Awesome! That would be better and easier to read/understan. 

> This requires to know in what order are the compilers and extensions are loaded

Different bundles extensions loads independently, order is not important. After **all** extensions are loaded Symfony starts with the compiler passes. (The "last minute change before compiling the container") Our compiler passes loads in the order we specify in our HttplugBundle.php file. Since the `puli.discovery` service is defined in an extension we are safe to verify its existens in our compiler pass.

sagikazarmark: The order is important because of checking what services has been registered, which needs fallback, etc. But I think it is not a problem then.

Nyholm: :+1:

Tests/Unit/DependencyInjection/Compiler/DiscoveryPassTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->assertContainerBuilderHasService('httplug.uri_factory.default', 'Http\Message\UriFactory');
+        $this->assertContainerBuilderHasService('httplug.stream_factory.default', 'Http\Message\StreamFactory');
+    }
+
+    /**
+     * Overridden test as we have dependencies in this compiler pass.
+     *
+     * @test
+     */
+    public function compilation_should_not_fail_with_empty_container()
```
sagikazarmark: Not sure if this is a good idea. What are the symfony best practices for this?

Nyholm: I only know that Symfony best practices does not tell you anything about this. There is not event a cookbook entry about how you test configuration, extension or your compiler passes.

DependencyInjection/Compiler/DiscoveryPass.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Fallback services and classes.
+     *
+     * @var array
+     */
+    private $services = [
+        'client' => 'Http\Client\HttpClient',
+        'message_factory' => 'Http\Message\MessageFactory',
+        'uri_factory' => 'Http\Message\UriFactory',
+        'stream_factory' => 'Http\Message\StreamFactory',
```
Nyholm: Maybe use HttpClient::class etc here?

DependencyInjection/Compiler/DiscoveryPass.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    private function registerFactory(ContainerBuilder $container)
+    {
+        if (false === $container->has('puli.discovery')) {
+            throw new RuntimeException(
+                'You need to install puli/symfony-bundle or add configuration at httplug.classes in order to use this bundle. Refer to http://some.doc'
+            );
+        }
+
+        $definition = $container->register('httplug.factory', 'Http\HttplugBundle\Util\HttplugFactory');
```
Nyholm: Same here... HttplugFactory::class

DependencyInjection/Compiler/DiscoveryPass.php
```diff
@@ -0,0 +1,80 @@
...
+
+        $definition
+            ->addArgument(new Reference('puli.discovery'))
+            ->setPublic(false)
+        ;
+    }
+}
```
Nyholm: I like this solution!

DependencyInjection/Compiler/DiscoveryPass.php
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * @param ContainerBuilder $container
+     *
+     * @throws RuntimeException
+     */
+    private function registerFactory(ContainerBuilder $container)
+    {
+        if (false === $container->has('puli.discovery')) {
+            throw new RuntimeException(
+                'You need to install puli/symfony-bundle or add configuration at httplug.classes in order to use this bundle. Refer to http://some.doc'
```
Nyholm: And of course, before we merge this. We should update http://some.doc to something better ;)

Tests/Unit/DependencyInjection/Compiler/DiscoveryPassTest.php
dbu: well, unit tests should be straightforward. and the functional test is already covered if we do anything where we use the factories.

so the problem here is that if we run this test we get an error, right? do we not install the puli things on travis, would this not work?

DependencyInjection/Compiler/DiscoveryPass.php
```diff
@@ -0,0 +1,85 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        $useDiscovery = false;
+
+        foreach ($this->services as $service => $class) {
+            $serviceId = sprintf('httplug.%s.default', $service);
+
+            if (false === $container->has($serviceId)) {
+                // Register and create factory for the first time
+                if (false === $useDiscovery) {
```
dbu: why this? if we need several factories, will this not prevent the creation of the other factories?

sagikazarmark: The factory service itself is the same.

dbu: ah, ok, got it. but then we could call that variable simple puliFactoryLoaded instead of "useDiscovery"

Nyholm: No, the `$factory` is a "httplug.factory". A service that works as a factory class for our *Factory classes. (MessageFactory, ClientFactory etc.. )

Tests/Unit/DependencyInjection/Compiler/DiscoveryPassTest.php
sagikazarmark: It is not about installation. In this test, a container is built with only this compiler pass. And there is a burnt-in test which tests the thing against an empty container. If there is any httplug service which is not defined nor discovery is installed, the intended behaviour is to get an exception during compile.

DependencyInjection/Compiler/DiscoveryPass.php
Nyholm: wow, you are too quick..

Tests/Unit/DependencyInjection/Compiler/DiscoveryPassTest.php
dbu: is discovery not installed on travis? if its optional, i would add a check and skip the tests if discovery is not available.

sagikazarmark: Discovery is removed in this PR. Do you mean Puli?

Yes, Puli is installed, but this test only tests this specific compiler pass.

.travis.yml
```diff
@@ -34,6 +34,7 @@ before_install:
     - travis_retry composer self-update
 
 install:
+    - wget https://github.com/puli/cli/releases/download/1.0.0-beta10/puli.phar && chmod +x puli.phar
```
Nyholm: Ping @dbu 
We download puli.phar here. We do also require "puli/symfony-bundle"

Tests/Unit/DependencyInjection/Compiler/DiscoveryPassTest.php
dbu: sorry. jetlagged brain in action ;-)

right. so i think we should have one test that makes sure we get a decent error message when the puli bundle is not available and nothing is configured explicitly. and 2 more tests, one for explicit configuration and one for when puli is available.

sagikazarmark: These cases are already tested.

sagikazarmark:
Squashed and rebased, ready for merge IMO. Calling for a final review.

sagikazarmark:
Note to self: the URL in exception must be changed.

Util/HttplugFactory.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+namespace Http\HttplugBundle\Util;
```
Nyholm: I do not like the `Util` namespace. It feels like the old days when you had a folder called `functions`. Maybe do something more descriptive like `Discovery`?

Tests/Resources/app/AppKernel.php
```diff
@@ -10,10 +10,16 @@ class AppKernel extends Kernel
...
     {
-        return [
+        $bundles =  [
             new \Symfony\Bundle\FrameworkBundle\FrameworkBundle(),
             new \Http\HttplugBundle\HttplugBundle(),
         ];
+
+        if (false === defined('HHVM_VERSION')) {
+            array_splice($bundles, 1, 0, [new \Puli\SymfonyBundle\PuliBundle()]);
+        }
```
Nyholm: Why do we need to insert the PuliBundle before HttplugBundle?
If the order is important we must document it.

sagikazarmark: Well, rethinking this: it does not, since all compilers are loaded after all extensions did.

Util/HttplugFactory.php
sagikazarmark: Not sure if we even need that namespace

Nyholm:
Awesome @sagikazarmark 
:+1: for me. 

Im happy to merge when the tests passes.

Util/HttplugFactory.php
Nyholm: :+1:

Tests/Resources/app/AppKernel.php
dbu: > Well, rethinking this: it does not, since all compilers are loaded after
all extensions did.

exactly, one of the principal reasons to use a compiler pass. you can even
in the extension class see which bundles are instantiated, but not their
config

PR #62, opened 2016-03-02
Updated links to the new documentation

### LINKED ISSUE #60: Update link in error message
We introduced an error message in #55 if PuliBundle was not installed. We should update the link so it points directly to the heading in our documentation that explains how to set up the bundle with and without discovery.

Nyholm:
This will fix #60


dbu:
thanks!

Issue #64, opened 2016-03-03
Todos before stable


sagikazarmark:
- [x] Make PluginClientFactory final
- [x] Fix PluginClientFactory constructr docblock (array of Plugins)
- [x] Update changelog
- [x] Update README title: "Symfony integration for HTTPlug"

Nyholm:
:+1:

Nyholm:
We do also need to deal with these: 
* https://github.com/php-http/HttplugBundle/blob/master/HttplugFactory.php#L83
* https://github.com/php-http/HttplugBundle/blob/master/HttplugFactory.php#L83

PR #65, opened 2016-03-03
Cleanup

Nyholm:
This will address some of the changes in #64

### DIFF
--- a/HttplugFactory.php
+++ b/HttplugFactory.php
@@ -37,7 +37,6 @@ final class HttplugFactory
     {
         $class = $this->findOneByType($type);
 
-        // TODO: use doctrine instantiator?
         return new $class();
     }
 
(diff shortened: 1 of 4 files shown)


HttplugFactory.php
```diff
@@ -37,7 +37,6 @@ public function find($type)
     {
         $class = $this->findOneByType($type);
 
-        // TODO: use doctrine instantiator?
```
Nyholm: I do not believe it is a good idea to instanciate a factory class without calling its constructor. Our `DiactorosMessageFactory` will not work unless its constructor is used.

sagikazarmark: Agree, remove that comment

sagikazarmark:
Added a few more things to #64

Nyholm:
Added changelog

Nyholm:
:+1:

sagikazarmark:
I added a few things as well. Can you please squash commits. Call it something like Prepare stable release. Thank you.

Nyholm:
Squashed!

sagikazarmark:
Thanks

PR #66, opened 2016-03-03
Added tests for Httplug factory

Nyholm:
I did also rewrite the `evaluateCondition` function to support associative arrays. 

This will address one issue in #64

### DIFF
--- a/HttplugFactory.php
+++ b/HttplugFactory.php
@@ -79,8 +79,6 @@ final class HttplugFactory
      * @param mixed $condition
      *
      * @return bool
-     *
-     * TODO: review this method
      */
     protected function evaluateCondition($condition)
     {
@@ -95,8 +93,8 @@ final class HttplugFactory
             $evaluatedCondition = true;
 
             // Immediately stop execution if the condition is false
-            for ($i = 0; $i < count($condition) && false !== $evaluatedCondition; ++$i) {
-                $evaluatedCondition &= $this->evaluateCondition($condition[$i]);
+            while (count($condition) > 0 && $evaluatedCondition) {
+                $evaluatedCondition = $this->evaluateCondition(array_shift($condition));
             }
 
             return $evaluatedCondition;
--- a/Tests/Unit/HttplugFactoryTest.php
+++ b/Tests/Unit/HttplugFactoryTest.php
@@ -1,0 +1,123 @@
+<?php
+
+namespace Http\HttplugBundle\Tests\Unit;
+
+use Http\HttplugBundle\ClientFactory\DummyClient;
+use Http\HttplugBundle\Collector\MessageJournal;
+use Http\HttplugBundle\HttplugFactory;
+use Puli\Discovery\Api\Discovery;
+use Puli\Discovery\Binding\ClassBinding;
+use Webmozart\Expression\Expr;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class HttplugFactoryTest extends \PHPUnit_Framework_TestCase
+{
+    public function testEvaluateConditionString()
+    {
+        $existingClass = MessageJournal::class;
+        $factory = $this->getMockedFactory($existingClass);
+        $this->assertInstanceOf(DummyClient::class, $factory->find('type'));
+    }
+
+    /**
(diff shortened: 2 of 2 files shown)


HttplugFactory.php
```diff
@@ -95,11 +93,14 @@ protected function evaluateCondition($condition)
             $evaluatedCondition = true;
 
             // Immediately stop execution if the condition is false
-            for ($i = 0; $i < count($condition) && false !== $evaluatedCondition; ++$i) {
-                $evaluatedCondition &= $this->evaluateCondition($condition[$i]);
```
Nyholm: I can not see any reson why we need `&=`

sagikazarmark: Because `$evaluatedCondition` is checked before the next loop and breaks it if false.

sagikazarmark: I would not remove this as it is a wonderful solution for sparing that extra if you added.

Nyholm: I will only "look nice". You will still do the same number of comparisons. The current solution does not support assoc. arrays. 

Anyhow. I updated the PR so it gets the best of both worlds... What do you think?

sagikazarmark: Same number of comparisons, but less branches IMO. LGTM.

Tests/Unit/HttplugFactoryTest.php
```diff
@@ -0,0 +1,47 @@
...
+use Puli\Discovery\Api\Discovery;
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class HttplugFactoryTest extends \PHPUnit_Framework_TestCase
+{
+    public function testEvaluateCondition()
+    {
+        $method = new \ReflectionMethod(HttplugFactory::class, 'evaluateCondition');
```
sagikazarmark: The correct way would be mocking the Puli discovery, returning a mocked Class Binding and using it for testing here.

Nyholm: And not using Reflection? Would that allow me to isolate the tests for `evaluateCondition`?

sagikazarmark: Yes. You can mock the hasParameter and getParameter methods of the class binding.

Generally you should always test the public API. It is part of the public API that if a class binding has a depends parameter, it gets evaluated "somehow".

So what you really want to test in this case is what separate depends parameters return a valid class binding, which of them end up in an exception.

Nyholm: > Generally you should always test the public API. 
I know that is "generally" the idea. Your suggestion would require a lot of mocking and is very likely to break if the class changes. That's why I chose to use reflection to just test the protected function. 

> So what you really want to test in this case is what separate depends parameters return a valid class binding, which of them end up in an exception.

Haha. No :smile:  I really wanted to only test the `evaluateCondition` and leave the rest of the class untested.

sagikazarmark: > Your suggestion would require a lot of mocking and is very likely to break if the class changes.

What class?

I think it requires an acceptable amount of mocks. Not to mention you can reuse mock builders. (Even better: as of PHPUnit 4.5 you can use prophecy known from phpspec. You could even use phpspec in httplugbundle as well along with phpunit, that would probably be even better)

Nyholm: > What class?

This, HttplugFactory

Okey, 
I'll give it a try and we take it from there.  =)

Nyholm:
Ive updated the PR using Prophecy.

Tests/Unit/HttplugFactoryTest.php
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class HttplugFactoryTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var \Prophecy\Prophet
+     */
+    private $prophet;
```
sagikazarmark: You don't need to manually instantiate a prophet object. Try `$this->prophesize()`

Nyholm: Thank you!

PR #69, opened 2016-03-06
Added Buzz client factory

Nyholm:
I used the option resolver for this factory. That is because the Buzz Client constructor did not accept an array with config. The OptionsResolver lets me avoid lots of `isset($config['foo])`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## UNRELEASED
+
+### Added
+
+- Client factories for Buzz.
+
+
 ## 1.0.0 - 2016-03-04
 
 ### Added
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## dev-master
```
sagikazarmark: Ah, this hurts my eyes. :smile: Call it UNRELEASED please.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
...
+
+### Added
+
+- Client factories for Buzz.
+
```
sagikazarmark: Two lines between releases

CHANGELOG.md
Nyholm: haha. We are different you an I =)

sagikazarmark:
Also squash please. Good to go then.

Nyholm:
Thank you for the review.

PR #72, opened 2016-05-04
Replace plugins package with client-common

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes
| License         | MIT

#### What's in this PR?

Replaces php-http/plugins with php-http/common-client.

#### Why?

The plugins package has been deprecated and moved to client-common.

#### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Guzzle 6 client is now created according to the Httplug specifications with automated minimal behaviour.
   Make sure you configure the Httplug plugins as needed,
   for example if you want to get exceptions for failure HTTP status codes.
+- **[BC] PluginClientFactory returns an instance of `Http\Client\Common\PluginClient`** (see [php-http/client-common#14](https://github.com/php-http/client-common/pull/14))
+- Plugins are loaded from their new packages, fall back to `php-http/plugins` if necessary
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
   Make sure you configure the Httplug plugins as needed,
   for example if you want to get exceptions for failure HTTP status codes.
 - **[BC] PluginClientFactory returns an instance of `Http\Client\Common\PluginClient`** (see [php-http/client-common#14](https://github.com/php-http/client-common/pull/14))
-- Plugins are loaded from their new packages, fall back to `php-http/plugins` if necessary
+- Plugins are loaded from their new packages
 
 
 ### Fixed
--- a/ClientFactory/PluginClientFactory.php
+++ b/ClientFactory/PluginClientFactory.php
@@ -2,8 +2,8 @@
 
 namespace Http\HttplugBundle\ClientFactory;
 
-use Http\Client\Plugin\Plugin;
-use Http\Client\Plugin\PluginClient;
+use Http\Client\Common\Plugin;
+use Http\Client\Common\PluginClient;
 
 /**
  * This factory creates a PluginClient.
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -3,9 +3,6 @@
 namespace Http\HttplugBundle\DependencyInjection;
 
 use Http\Client\Common\Plugin\AuthenticationPlugin;
-use Http\Client\Common\Plugin\CachePlugin;
-use Http\Client\Common\Plugin\LoggerPlugin;
-use Http\Client\Common\Plugin\StopwatchPlugin;
 use Http\Client\Common\PluginClient;
 use Http\HttplugBundle\ClientFactory\DummyClient;
 use Http\Message\Authentication\BasicAuth;
@@ -141,10 +138,6 @@ class HttplugExtension extends Extension
     {
         switch ($name) {
             case 'cache':
-                // To preserve BC, we check the existence of the new plugin class and use it if available
-                if (class_exists(CachePlugin::class)) {
-                    $definition->setClass(CachePlugin::class);
-                }
                 $definition
                     ->replaceArgument(0, new Reference($config['cache_pool']))
                     ->replaceArgument(1, new Reference($config['stream_factory']))
@@ -160,10 +153,6 @@ class HttplugExtension extends Extension
                 $definition->replaceArgument(0, new Reference($config['journal']));
                 break;
             case 'logger':
-                // To preserve BC, we check the existence of the new plugin class and use it if available
-                if (class_exists(LoggerPlugin::class)) {
-                    $definition->setClass(LoggerPlugin::class);
-                }
                 $definition->replaceArgument(0, new Reference($config['logger']));
                 if (!empty($config['formatter'])) {
                     $definition->replaceArgument(1, new Reference($config['formatter']));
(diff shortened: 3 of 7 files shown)


sagikazarmark:
If one of the three separated plugins are not installed, then any configuration that specifically enables those plugins will end up in a BC break. So I think we have to install the plugins package as well, and replace the used class if the package is installed.

composer.json
```diff
@@ -19,6 +19,7 @@
         "php": ">=5.5",
         "php-http/client-implementation": "^1.0",
         "php-http/message-factory": "^1.0.2",
+        "php-http/client-common": "^1.1",
         "php-http/plugins": "^1.0",
```
dbu: if we require both the new and the old plugins now, we don't need the flexible code you introduce above.

but i think to avoid BC breaks, we can indeed depend on both.

DependencyInjection/HttplugExtension.php
```diff
@@ -205,4 +219,25 @@ private function configureAuthentication(ContainerBuilder $container, array $con
...
+     * @param array  $pluginConfig
+     *
+     * @return bool
+     */
+    private function resolvePluginStatus($name, array $pluginConfig)
+    {
+        switch ($name) {
+            case 'logger':
+                return class_exists('Http\Client\Common\Plugin\LoggerPlugin');
+                break;
```
dbu: no need to break after unconditional return

DependencyInjection/HttplugExtension.php
```diff
@@ -205,4 +219,25 @@ private function configureAuthentication(ContainerBuilder $container, array $con
...
+     * Resolve the plugin enabled status if it is 'auto'.
+     *
+     * Returns false if plugin has no auto status allowed.
+     *
+     * @param string $name
+     * @param array  $pluginConfig
+     *
+     * @return bool
+     */
+    private function resolvePluginStatus($name, array $pluginConfig)
```
dbu: why does this only handle logger and no other plugins?

ClientFactory/PluginClientFactory.php
```diff
@@ -2,8 +2,8 @@
 
 namespace Http\HttplugBundle\ClientFactory;
 
-use Http\Client\Plugin\Plugin;
-use Http\Client\Plugin\PluginClient;
+use Http\Client\Common\Plugin;
+use Http\Client\Common\PluginClient;
```
dbu: i think this is technically a BC break - if somebody typehinted the old plugin client, things will fail. but i think thats really an edge case, just mention it in the changelog.

composer.json
dbu: ah sorry, i was confused. what you do is for the plugins that are now separate you use the new one if are available and otherwise keep using the old one. i think that is fine. once we move to 2.0 the separate plugins will only be available if included.

sagikazarmark: Exactly.

DependencyInjection/HttplugExtension.php
sagikazarmark: This method is not necessary ATM. I started to add this auto, true, false enabled config, but I realized that even in auto mode as default, it would be a BC break for already existing configuration. That's why I added back the plugins dependency. When we release 2.0, we can remove the plugins dep, add the proposed auto default value. Until then, this method is not necessary.

ClientFactory/PluginClientFactory.php
sagikazarmark: Hm, you are right. However I think it is a design mistake to rely on the plugin client: one's code should work with any kind of http client.

sagikazarmark: Added to changelog

DependencyInjection/HttplugExtension.php
dbu: then lets move the whole "auto" logic out of this PR. we can create a PR for the 2.0 version with it. and probably simply not merge it yet, as there is no hurry.

sagikazarmark: Already did, check the latest version of the PR.

dbu:
looks good. i suggest we bump the composer alias for master to 1.1-dev so that we release this as a new minor version. together with the other fix i started preparing.

sagikazarmark:
Why not 1.2-dev then?

David Buchmann <notifications@github.com> ezt írta (2016. május 9., hétfő):

sagikazarmark:
Squashed commits. Anything else here @dbu?

DependencyInjection/HttplugExtension.php
```diff
@@ -137,6 +138,9 @@ private function configurePluginByName($name, Definition $definition, array $con
     {
         switch ($name) {
             case 'cache':
+                if (class_exists('Http\Client\Common\Plugin\CachePlugin')) {
```
Nyholm: I would prefer a comment here that says why your are doing this. It seams kind of weird if you do not know that one namespace was deprecated and you are doing this because we want to keep BC. 

Same with the logger and stopwatch.

DependencyInjection/HttplugExtension.php
```diff
@@ -152,6 +156,9 @@ private function configurePluginByName($name, Definition $definition, array $con
                 $definition->replaceArgument(0, new Reference($config['journal']));
                 break;
             case 'logger':
+                if (class_exists('Http\Client\Common\Plugin\LoggerPlugin')) {
```
xabbuh: What about using `LoggerPlugin::class` and add a proper `use` statement (same with the other places)?

DependencyInjection/HttplugExtension.php
sagikazarmark: Great idea, thanks, will do that.

DependencyInjection/HttplugExtension.php
sagikazarmark: I thought we support PHP 5.4, but just checked and we don't, so this is a good idea, thanks.

Nyholm:
👍

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 - Client factories for Buzz.
 
+### Changed
+
+- **[BC] PluginClientFactory returns an instance of `Http\Client\Common\PluginClient`** (see [php-http/client-common#14](https://github.com/php-http/client-common/pull/14))
+- Plugins are loaded from their new packages, fall back to `php-http/plugins` if necessary
+
```
dbu: should we say that version 2 of the bundle will not automatically require the plugins that moved to separate repositories? i am not sure if we want to do that actually - cache, logger and stopwatch all depend on things that symfony depends on anyways, so there is no cost in requiring them in the bundle, right? maybe we even want to explicitly require those new plugins rather than the BC repository.

dbu:
needs a rebase

CHANGELOG.md
sagikazarmark: I am fine with that as well. In that case, we don't need to rely on the old package.

sagikazarmark:
Added a separate commit for plugin package removal.

Blocked by php-http/stopwatch-plugin#1

dbu:
https://github.com/php-http/stopwatch-plugin/releases/tag/1.0.1

i restarted the build, lets see if its fixed with this

sagikazarmark:
Dont forget to set the lowest version for the plugin.

David Buchmann <notifications@github.com> ezt írta (2016. május 19.,
csütörtök):

dbu:
do you mean stopwatch? composer will figure that out i think.

sagikazarmark:
Symfony 3 will not work with 1.0 as the lowest version

David Buchmann <notifications@github.com> ezt írta (2016. május 19.,
csütörtök):

dbu:
composer will simply choose 1.0.1 with symfony 3, as there is no other
resolution to satisfy all constraint.

sagikazarmark:
Ok

sagikazarmark:
Merge and tag 1.1?

dbu:
yup, done. thanks!

PR #77, opened 2016-05-12
Fix Puli issue on PHP 5.6

sagikazarmark:
Fixes an issue introduced in composer 1.0:

https://github.com/puli/issues/issues/190


dbu:
:+1:

PR #79, opened 2016-05-28
Add plugin container parameters

sagikazarmark:


xabbuh:
Is there a special use case you need the service class names to be controlled through container parameters?

sagikazarmark:
It would have been easier when we changed the namespace. It's quite common
that you have to create your own plugin services. I thought it might be
better to have a parameter instead of class names.

Am I wrong?

Christian Flothmann <notifications@github.com> ezt írta (2016. május 29.,
vasárnap):

dbu:
symfony started deprecating this pattern and stopped doing it in symfony 3. i am -1 introducing this here. we hopefully won't change class names again. and we open the door to let people change those paramters to instantiate other plugin classes which is a real antipattern.

sagikazarmark:
Okay, thanks for clarification (again) :smile:

PR #81, opened 2016-06-28
Require discovery 0.9 and do not handle Puli.

### LINKED ISSUE #80: Remove Puli dependency
The new discovery should be used, although I expect some conflicts.

/cc @Nyholm

Nyholm:
I require discovery library, that is now okey since it is not a very light weight package. Since Puli is optional and handled in the discovery package we do not longer need to handle that in the bundle. 

This will fix #80.

Tests are failing, we are waiting on https://github.com/php-http/buzz-adapter/pull/3 and new **patch** versions on buzz-adapter and react-adapter


sagikazarmark:
Something is wrong with the composer.json

sagikazarmark:
Packages released

Nyholm:
Thank you for the quick review and new releases. I've updated the PR

dbu:
looks like we have a problem with prefer-lowest. do you need to up one of the dependency further?

dbu:
thanks for the PR, great job, we can delete some code!

Nyholm:
Yes, It always feels good creating a PR like this were you remove a lot of code and barley adding any. 

Tests are passing now.

sagikazarmark:
Do we really need the discovery as a dependency? Or check for it if the classes are set in the config, and complain only in that case? Like we did with puli?

Nyholm:
That is a good question Mark. As far as I can remember, we did not require Puli by default to give the user a change to use the bundle without Puli. That was because Puli was a to large of dependency. 

Since version 0.9 our discovery layer is very light weight and I think it is okey to require discovery by default. This will give a better developer experience if things work out of the box.

sagikazarmark:
My only concern is that discovery is not stable yet, so version conflicts are more likely to happen.

Nyholm:
Im not that concerned. As long as we are really careful with following semver.

dbu:
i am pro requiring it. i don't expect people to require discovery
independently of the httplug bundle so as long as we don't break
discovery in a bad way we should be fine.

what is missing to make discovery stable, btw?

sagikazarmark:
> what is missing to make discovery stable, btw?

Well, now that Puli is not a requirement anymore, probably nothing. But if Puli breaks BC in the stable version, we have a problem.

Nyholm:
I created an issue for this on the discovery repo. https://github.com/php-http/discovery/issues/62

PR #82, opened 2016-06-29
Allow CURL constant names

### LINKED ISSUE #71: Add support for string curl client params
Maybe we could accept curl client constants as well. Before passing to the client, we could resolve the constant names to actual values. What do you think?

Nyholm:
This will fix #71

### DIFF
--- a/ClientFactory/CurlFactory.php
+++ b/ClientFactory/CurlFactory.php
@@ -40,6 +40,18 @@ class CurlFactory implements ClientFactory
             throw new \LogicException('To use the Curl client you need to install the "php-http/curl-client" package.');
         }
 
+        // Try to resolve curl constant names
+        foreach ($config as $key => $value) {
+            // If the value starts with 'CURL' we assume it is a reference to a constant.
+            if (strpos($key, 'CURL') === 0) {
+                $contantValue = constant($key);
+                if ($contantValue !== null) {
+                    unset($config[$key]);
+                    $config[$contantValue] = $value;
+                }
+            }
+        }
+
         return new Client($this->messageFactory, $this->streamFactory, $config);
     }
 }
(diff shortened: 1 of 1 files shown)


ClientFactory/CurlFactory.php
```diff
@@ -40,6 +40,18 @@ public function createClient(array $config = [])
             throw new \LogicException('To use the Curl client you need to install the "php-http/curl-client" package.');
         }
 
+        // Try to resolve curl constant names
+        foreach ($config as $key => $value) {
+            // If the value starts with 'CURL' we assume it is a reference to a constant.
+            if (strpos($key, 'CURL') === 0) {
+                $contantValue = constant($key);
```
dbu: is there something to check if a constant exists?

and s/contantValue/constantValue ;-)

Nyholm: If the constant does not exist constant() will return null. 

Question: is it true that the only config parameters to the curl client are integers or constants named "CURL*"? If so, we should throw a exception.

dbu: according to http://php.net/manual/en/function.curl-setopt.php it must be int. so either it already is an int here, or its a constant. maybe we could do

```
if (!is_int($key)) {
    $constantValue = constant($key);
    if (null === $constantValue) {
        throw... $key is not an integer or a CURL constant.
    }
    unset...,
    $config[$constantValue] = $value;
}
```

dbu: not sure if we should force the constants to start with CURL - maybe there could be a reason to have some other constant that does not start with CURL.

is `constant()` case sensitive?

sagikazarmark: Invalid constants (integers) will end up in an error upon setopt anyway.

Nyholm: Okey. Then I'm safe to assume that everything that comes in is either an Int or a name of a constant. If the constant is not defined I'll throw an exception. If the Int is invalid CURL will throw an exception

Nyholm:
Thank you for the feedback. I've updated the PR accordingly.

dbu:
awesome, thanks!

Nyholm:
Thank you for merging.

Issue #70, opened 2016-03-08
Add support for clients in client-common


sagikazarmark:
For example methods client

Nyholm:
We got the current clients: 

* [ ] BatchClient
* [x] PluginClient (added with configuration)
* [x] HttpMethodsClient
* [x] FlexibleClient
* [x] EmulatedHttpClient (by FlexibleClient)
* [x] EmulatedHttpAsyncClient (by FlexibleClient)

Should we support BachClient as well?

dbu:
it sounds useful to have the batch client. i think there are use cases to want to send whole batches of requests.

Nyholm:
Okey. I'll try to manage a PR with that as well.

Nyholm:
Ive submitted a PR for the BatchClient as well. PR #103 will close this ticket.

PR #83, opened 2016-06-29
Add config for enable FlexibleHttpClient and HttpMethodsClient

Nyholm:
This will address part of #70 

With config options you wrap your client in FlexibleHttpClient or HttpMethodsClient. Please give input on the config names. I want them to be descriptive but at the same time also reflect the class we use for wrapping. 

A descriptive config name would be "emulate_async_client" but that is very far from the name `FlexibleHttpClient`. I guess you understand my point. Any suggestions are welcome.


DependencyInjection/HttplugExtension.php
```diff
@@ -206,4 +196,55 @@ private function configureAuthentication(ContainerBuilder $container, array $con
...
+                ->addArgument($arguments['config']);
+        }
+
+
+        /*
+         * Decorate the client with clients from client-common
+         */
+
+        if ($arguments['flexible_client'] && $arguments['add_http_methods']) {
+            throw new \LogicException('A http client can\'t be decorated with both FlexibleHttpClient and HttpMethodsClient. Only one of the following options can be true. ("flexible_client", "add_http_methods")');
```
dbu: this should be validated in the Configuration class. you can add a `validate()` section (or was it `validation()`?) on the level that has the two options.

DependencyInjection/Configuration.php
```diff
@@ -102,6 +102,14 @@ protected function configureClients(ArrayNodeDefinition $root)
                         ->cannotBeEmpty()
                         ->info('The service id of a factory to use when creating the adapter.')
                     ->end()
+                    ->booleanNode('flexible_client')
+                        ->defaultFalse()
+                        ->info('If true we will wrap this client in a FlexibleHttpClient which emulates async or sync behavior.')
```
dbu: i would say: 

Set to true to get the client wrapped in a FlexibleHttpClient which emulates...

DependencyInjection/Configuration.php
```diff
@@ -102,6 +102,14 @@ protected function configureClients(ArrayNodeDefinition $root)
...
+                    ->end()
+                    ->booleanNode('add_http_methods')
```
dbu: i think calling this http_methods_client would be fine

dbu:
the name flexible_client seems right to me. i would call the other http_methods_client.

Nyholm:
Thank you for your review. I've update the PR according to your suggestions

dbu:
thanks a lot @Nyholm !
created https://github.com/php-http/documentation/issues/107 for the documentation.

Nyholm:
Thank you for merging.

PR #84, opened 2016-07-03
Improved collector

Nyholm:
This PR is work in progress. I just want to show the progress. Please try it out and give me feedback. I believe the collector and extensions work properly. The work that is left is on the design and display of things. 

Related to #26 

### TODO

* [x] Only show the first request in the stack by default
* [x] Show rest of the stack when you click "Show all"
* [x] Hide the body of the request by default and add a "Show body" button. 
* [ ] Add nice design 
* [x] Maybe put different adapters on different pages. Design it like the translation "Translated"/"Fallback"/"Not translated"


dbu:
great work @Nyholm !

how do you display the information? we will end up with a lot of information if several plugins are active...

Nyholm:
Thank you. 

Im not too sure at the moment to be honest... I'll look at Symfony pages and try to do something similar.

dbu:
okay. maybe for each request show the final version that was sent / received by default, and have little tabs to see the other versions between plugins?

Nyholm:
This PR is ready for a review. Im not sure I can make the design look better.

Nyholm:
Here are some screenshots

### First view
screen shot 2016-07-13 at 21 06 29

### Show a stack
screen shot 2016-07-13 at 21 06 59
### Show body
screen shot 2016-07-13 at 21 07 14

### Show errors by the error plugin
screen shot 2016-07-13 at 21 40 48

### Show errors from the http client
screen shot 2016-07-13 at 21 41 20

Nyholm:
Blocked by https://github.com/php-http/message/issues/45

Collector/ClientDataProvider.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        return $this->requests;
+    }
+
+    /**
+     * @param array $requests
+     *
+     * @return ClientDataProvider
+     */
+    public function setRequests($requests)
```
Nyholm: Maybe setters should be removed here

Nyholm: ...and the getters... They are not used. Only get*Stack is used.

Collector/DebugPlugin.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Http\HttplugBundle\Collector;
+
+use Http\Client\Common\Plugin;
+use Http\Client\Exception;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * A plugin used for log requests and responses passing through each plugin.
```
Nyholm: Typo. Make sure this is understandable.

Collector/Twig/HttpMessageMarkupExtension.php
```diff
@@ -0,0 +1,45 @@
...
+            new \Twig_SimpleFilter('httplug_markup', [$this, 'markup'], ['is_safe' => ['html']]),
+        ];
+    }
+
+    /**
+     * @param string $message http message
+     */
+    public function markup($message)
+    {
+        $safeMessage = htmlentities($message);
```
Nyholm: Is this safe enough?

joelwurtz:
Not sure if it's possible to add this here, but i would love to have an option to show request and response content as sometime you want to debug the content of the request you send or the response you receive.

However this can have memory and latency issue for big request / response (i.e. files), so it would be better to add an option in the config (which default to false) for allowing this feature.

WDYT ?

Nyholm:
request and response content? Do you mean the HTTP message body? If so, you can toggle that in the top right corner. Look at image 3. =)

joelwurtz:
Who didn't see that, is this an option for the debugger ? And also does it show also the request body ?

Nyholm:
I use the new [FullHttpMessageFormatter](https://github.com/php-http/message/blob/master/src/Formatter/FullHttpMessageFormatter.php) to capture the full HTTP message for both requests and responses. 

You may configure the FullHttpMessageFormatter to a different limit on the body. Default is 1000 chars. To change that construction parameter is not supported in this PR. I'll leave that for a future update.

joelwurtz:
Just see the formatter and how the request / response is read.

As far as i have see, this formatter mess up with the stream, as __toString is called and the stream is not rewind so the content is retrieve only in the first stream ?

(I may have miss something again ^^)

Nyholm:
I believe you are correct. I created an issue. 

So streams should always be rewinded after we read them?

joelwurtz:
> So streams should always be rewinded after we read them?

Depends if the stream is rewindable or not, if not we must buffer the stream into memory and replace it in the request and this will have a huge impact on performance IMO

Nyholm:
Who is using streams that you cant rewind (not seekable) in this case?

Guzzle has PumpStream, NoSeekStream, BufferStream and Zend has CallbackStream which never are seekable.

I think we should add allow to opt-out of this if there is a major performance impact.

Collector/Twig/HttpMessageMarkupExtension.php
dbu: i think so, yes

dbu:
This looks really cool! Some of those screenshots could be used to illustrate the plugin concept in the documentation :-)

regarding streams: i suggest we default to not read anything of the body, but instead have a hint in the debug output that it can be enabled. i would usually expect people to reproduce a call with curl if they want to see the response. (i think we discussed showing a curl command line for a request which would make this trivial - but that is out of scope for this PR)

reading the body means messing with streams, which can pose problems or huge slowdowns if there is larger http bodies involved. if its no bid effort, i am ok to keep the option available, but disabled by default.

sagikazarmark:
> i think we discussed showing a curl command line for a request

That would be a cURL formatter in the message package: https://github.com/php-http/message/issues/19

> reading the body means messing with streams, which can pose problems or huge slowdowns if there is larger http bodies involved. if its no bid effort, i am ok to keep the option available, but disabled by default.

This is not always possible. In some cases the stream is read-once. We need to provide an alternative stream in that case with the original content.

dbu:
i would prefer to wait with merging this until we have a solution to not
break requests with non-rewindable streams.

sagikazarmark:
As @Nyholm said copying is not an option, because the request is immutable. So non-rewindable streams should never be read, otherwise users can optionally turn on this feature.

Collector/RequestStackProvider.php
```diff
@@ -68,7 +68,7 @@ public static function createFromCollectedData(array $data)
 
     /**
      * @param array $messages is an array with keys 'failure', 'request' and 'response' which hold requests for each call to
-     *                        sendRequest and for each depth.
+     *                        sendRequest and for each depth
```
sagikazarmark: Not sure if this is a solution. Normally we need that period. See https://github.com/php-http/message/pull/48

sagikazarmark:
@Nyholm it seems symfony changed their CS: https://github.com/symfony/symfony/pull/19198

However, the fixer still seems to be buggy, so I would stick with the full stop everywhere and change it later if necessary. I updated the master branch with the latest styleci config.

Nyholm:
So, No more CS changes in this PR?

Since the merge of https://github.com/php-http/message/pull/47 is there anything left to be done in this PR?

sagikazarmark:
After merging #86 can you please rebase from master and revert https://github.com/php-http/HttplugBundle/pull/84/commits/4c331320df47ace2b95778a4ec43473e03c0cfc8?

sagikazarmark:
UX: Request 0 -> Request #0

Looks better IMHO, WDYT?

Nyholm:
Sure

Collector/DebugPlugin.php
```diff
@@ -0,0 +1,63 @@
...
+use Http\Client\Exception;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * A plugin used for log requests and responses. This plugin is executed between each normal plugin.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class DebugPlugin implements Plugin
```
sagikazarmark: Does it make sense to extend this? If not, can it be final?

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    private $formatter;
+
+    /**
+     * @var PluginJournal
+     */
+    private $journal;
+
+    /**
+     * @param Formatter $formatter
```
sagikazarmark: Missing phpdoc.

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+
+/**
+ * A data collector for the debug plugin.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class DebugPluginCollector extends DataCollector
```
sagikazarmark: Does it make sense to extend this? If not, can it be final?

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+     * @param Formatter $formatter
+     */
+    public function __construct(Formatter $formatter, PluginJournal $journal)
+    {
+        $this->formatter = $formatter;
+        $this->journal = $journal;
+    }
+
+    /**
+     * @param RequestInterface $request
```
sagikazarmark: Missing phpdoc

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    public function addRequest(RequestInterface $request, $clientName, $depth)
+    {
+        $this->data[$clientName]['request'][$depth][] = $this->formatter->formatRequest($request);
+    }
+
+    /**
+     * @param ResponseInterface $response
```
sagikazarmark: Missing phpdoc or inheritdoc

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    public function addResponse(ResponseInterface $response, $clientName, $depth)
+    {
+        $this->data[$clientName]['response'][$depth][] = $this->formatter->formatResponse($response);
+        $this->data[$clientName]['failure'][$depth][] = false;
+    }
+
+    /**
+     * @param Exception $exception
```
sagikazarmark: Missing phpdoc or inheritdoc

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Returns the successful request-resonse pairs.
+     *
+     * @return array
+     */
+    public function getSucessfulRequests()
+    {
+        $count = 0;
+        foreach ($this->data as $client) {
```
sagikazarmark: Missing newline before control structure

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Returns the failed request-resonse pairs.
+     *
+     * @return array
+     */
+    public function getFailedRequests()
+    {
+        $count = 0;
+        foreach ($this->data as $client) {
```
sagikazarmark: Missing newline before control structure

Collector/DebugPluginCollector.php
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'httplug';
+    }
+
```
sagikazarmark: I would say we should add inheritdoc here.

Collector/PluginJournal.php
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        return;
+    }
+
+    /**
+     * @param string $clientName
+     * @param array  $plugins
+     *
+     * @return $this
```
sagikazarmark: Why do we need fluent interface here?

Collector/PluginJournal.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Http\HttplugBundle\Collector;
+
+/**
+ * A object that remembers what plugins are configured for which clients.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class PluginJournal
```
sagikazarmark: If this class should not be extended, can it be final?

Collector/RequestStackProvider.php
```diff
@@ -0,0 +1,141 @@
...
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class RequestStackProvider
+{
+    /**
+     * Array that tell if a request errored or not. true = success, false = failure.
+     *
+     * @var array
+     */
+    private $failure;
```
sagikazarmark: Shouldn't this be plural too?

Collector/RequestStackProvider.php
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * An object that managed collected data for each client. This is used to display data.
+ *
+ * The Request object at $requests[0][2] is the state of the object between the third
+ * and the fourth plugin. The response after that plugin is found in $responses[0][2].
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class RequestStackProvider
```
sagikazarmark: If this class should not be extended, can it be final?

Collector/RequestStackProvider.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    public function getStackIndexKeys()
+    {
+        return array_keys($this->requests);
+    }
+
+    /**
+     * @param int $idx
+     *
+     * @return array responses
```
sagikazarmark: Not sure if we need these descriptions here. It is not verbose enough and can be confusing. Either have little but more verbose description or none IMO. Same for below.

Resources/public/script/httplug.js
```diff
@@ -0,0 +1,132 @@
+
+
+/**
+ * Toggle hide/show on the message body
+ */
+function httplug_toggleBody(el) {
+  var bodies = document.querySelectorAll(".httplug-http-body");
```
sagikazarmark: I think it would be better to be consistent about indentation.

Resources/public/style/httplug.css
```diff
@@ -0,0 +1,24 @@
...
+    font-size: 130%;
+    font-weight: bold;
+}
+.httplug-request-stack {
+
+    display: none;
+}
+.httplug-error {
+    color: red;
+}
```
sagikazarmark: Missing newline

Resources/public/script/httplug.js
```diff
@@ -0,0 +1,132 @@
+
```
sagikazarmark: Whyleading newlines?

composer.json
```diff
@@ -19,13 +19,15 @@
...
+        "php-http/client-common": "^1.2",
         "php-http/cache-plugin": "^1.0",
         "php-http/logger-plugin": "^1.0",
         "php-http/stopwatch-plugin": "^1.0",
+        "php-http/message": "^1.3",
         "symfony/options-resolver": "^2.7|^3.0",
         "symfony/framework-bundle": "^2.7|^3.0",
-        "php-http/discovery": "^0.9"
+        "php-http/discovery": "^0.9",
+        "twig/twig": "^1.18|^2.0"
```
sagikazarmark: Double pipe please: https://getcomposer.org/doc/articles/versions.md#range

sagikazarmark:
As always, I am the style guy :wink:  Can you please take a look at my comments? I would leave the technical review to @dbu and @joelwurtz as I haven't really followed this conversation.

composer.json
Nyholm: I thought a double pipe was just bad sytax. Thank you!

Resources/public/script/httplug.js
Nyholm: Im not sure what you mean. I indent all my JS with 2 spaces. Isn't that consistent? PHP is a completely different language...

sagikazarmark: Well, doesn't really matter after all. But maybe we should add a rule to the editorconfig, otherwise some IDEs might want to indent differently.

sagikazarmark:
Can you rebase from master?

Nyholm:
This PR is now rebase and I've addressed all your comments except for the one with the Javascript indentation.

sagikazarmark:
So is this ready for merging then?

Nyholm:
If @joelwurtz and/or @dbu says it is okey, then it's ready to merge IMHO.

sagikazarmark:
Ok

ClientFactory/PluginClientFactory.php
```diff
@@ -16,11 +16,12 @@
      * @param Plugin[]      $plugins
      * @param ClientFactory $factory
      * @param array         $config
+     * @param array         $pluginClientOptions
```
dbu: can you add a short text that explains the $config and $pluginClientOptions? as its two sets of options, it would help to know what each is to keep them cleanly separated. saying that $pluginClientOptions is forwareded to PluginClient is good enough for me.

dbu:
looks good. mark had some comments about missing `@inheritDoc` or explanations on the methods, can you adress those as well please? and can you please squash the commits? then looks good to merge to me.

Nyholm:
I believe I already have addressed Marks comments about `@inhertdoc`. They are just not marked as "comment on an outdated diff"

Nyholm:
Rebased and ready for being merged

dbu:
great job, thanks!

Nyholm:
Thank you all for the reviews and feedback.

sagikazarmark:
Great job as always @Nyholm. Many thanks.

PR #87, opened 2016-07-16
Discovery bugfixes and more flexible.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

With this PR I wanted to achieve two things: 
* A way to configure which client that should be used with auto discovery
* A way to opt-out of providing a auto discovery client in the toolbar.


DependencyInjection/Configuration.php
```diff
@@ -83,6 +83,8 @@ public function getConfigTreeBuilder()
                             ->defaultValue('auto')
                         ->end()
                         ->scalarNode('formatter')->defaultNull()->end()
+                        ->booleanNode('inject_debugger_for_discovered_http_client')->defaultTrue()->end()
```
sagikazarmark: I am not sure if we want this opt-in by default.

Nyholm: Why not? All the other clients are opt-in by default.

sagikazarmark: Correct me if I am wrong, but this will register the default client in the discovery, will it not? If so, all the problems I listed in the previous PR can still occur. I would rather have this as an optional feature stating you can turn this on for discovered clients if you want.

sagikazarmark: Can we find a shorter name? Like `profile_discovered_client` and `profile_discovered_async_client`

Nyholm: > Correct me if I am wrong, but this will register the default client in the discovery, will it not?

No, It will find take a (new) unconfigured client and add toolbar debugging on that and then make sure that this client is found by discovery in the future. 

This PR also lets to you decide if you want to use a configured client as auto discovery. 

> Can we find a shorter name? Like profile_discovered_client and profile_discovered_async_client

Yes please. It was quite late last night when I wrote this. =)

sagikazarmark: Okay, so by default a discovered, unconfigured client is simply wrapped with debug tools and put back to discovery and you can also choose to use a configured client for that, but that is optional, opt-out by default?

dbu:
you might need to cherry-pick commits into a new branch as #84 is now squashed. there is a conflict reported by github.

Nyholm:
Yes. I'll make this PR ready for review shortly.

Nyholm:
There. This PR is ready for review. 

Consider the following config
```yaml
httplug:
    toolbar:
      profile_discovered_client: true # Default
      profile_discovered_async_client: false # Default
    classes:
    clients:
        my_guzzle6:
        acme:
            factory: 'httplug.factory.curl'
```

This will create a two new private services: `httplug.auto_discovery_client.pure` and `httplug.auto_discovery_client.plugin`. The underlying client is a **new client** found by auto discovery. Next time a third party library is using auto discovery it will find the `httplug.auto_discovery_client.plugin` which is configured with debug stuff. 

If a user want to control what client is found by auto discovery he/she may add `use_with_discovery` to the client. 
```yaml
httplug:
    toolbar:
      profile_discovered_client: false # Does not matter when 'use_with_discovery'
      profile_discovered_async_client: false # Default
    classes:
    clients:
        my_guzzle6:
        acme:
            factory: 'httplug.factory.curl'
            use_with_discovery: 'http_client'
```

sagikazarmark:
> use_with_discovery: 'http_client'

What does this mean? Neither the key nor the value is obvious IMO.

Nyholm:
You can choose what client to be found by HttpClientDiscovery::find and HttpAsyncDiscovery::find. 

I added a comment about it here: https://github.com/php-http/HttplugBundle/pull/87/files#diff-850942b3ba24ab03a40aaa81b6152852R129

But maybe that was not the best one

sagikazarmark:
What if someone wants one client to be used as both? Is there a type check to make sure the chosen client is actually of the relevant type? What if this is configured for multiple clients?

Nyholm:
> What if someone wants one client to be used as both? 

Correct. That was missing

> Is there a type check to make sure the chosen client is actually of the relevant type?

Yes, In the discovery strategies contstuctor

> What if this is configured for multiple clients?

Then an exception will be thrown

sagikazarmark:
Then maybe we could have the following configuration:

``` yaml
httplug:
    discovery:
        client: client_id # optional
        async_client: async_client_id # optional
        profile_client: true
        profile_async_client: false
```

Nyholm:
I like it!
Will do

Nyholm:
Since `profile_client` as no effect when using httplug.discovery.client I suggest: 

``` yaml
httplug:
    discovery:
        client: 'auto' # could also be a service name like 'httplug.client.acme'
        async_client: 'none' # default
```

But this does not however tell you about the profiling... Is there a better config?

sagikazarmark:
When there is a concrete client configured, profiling is enabled for it automatically?

Nyholm:
Yes

sagikazarmark:
Then I am fine with that. Although probably we should have null (~) values by default and those should be mapped to auto and none. WDYT?

Nyholm:
This PR is now ready for review

DependencyInjection/Configuration.php
```diff
@@ -90,6 +90,20 @@ public function getConfigTreeBuilder()
...
+                ->arrayNode('discovery')
+                    ->addDefaultsIfNotSet()
+                    ->info('Control what clients should be found by the discovery.')
+                    ->children()
+                        ->scalarNode('client')
+                            ->defaultValue('auto')
+                            ->info('Set to "auto" to see auto discovered client in the web profiler. If provided a service id for a client then this client will be found by auto discovery.')
+                        ->end()
+                        ->scalarNode('async_client')
+                            ->defaultNull()
```
dbu: why not auto as well?

DependencyInjection/HttplugExtension.php
```diff
@@ -278,4 +280,53 @@ private function registerDebugPlugin(ContainerBuilder $container, $name)
...
+     *
+     * @param ContainerBuilder $container
+     * @param array            $config
+     */
+    private function configureAutoDiscoveryClients(ContainerBuilder $container, array $config)
+    {
+        $httpClient = $config['discovery']['client'];
+        if ($httpClient === 'auto') {
+            $httpClient = $this->registerAutoDiscoverableClientWithDebugPlugin($container, 'client');
+        } elseif ($httpClient !== null) {
```
dbu: in this case i would do `elseif ($httpClient) {` so that setting it to `false` or `''` works as well. none of those values would be valid service references.

dbu:
i have one question and one suggestion. but overall, this looks good, i like it. please also start with the documentation as soon as we are set on how this works, so that we don't forget the details of auto vs null and so on...

DependencyInjection/Configuration.php
Nyholm: Because if no AsyncClient is found we get an exception. I do not think I can catch that.

dbu: okay. but then lets change the info to mention that you should set this to auto if you know that you have an async client available.

Nyholm: Or should I put that in the docs?

sagikazarmark:
Can we merge this and tag v1.2?

DependencyInjection/Configuration.php
Nyholm: I added it in the docs: 
https://github.com/php-http/documentation/pull/118/files#diff-33b9b1161d0ce83931e73c48bf9e6900R71

Discovery/ConfiguredClientsStrategy.php
```diff
@@ -24,24 +25,37 @@ class ConfiguredClientsStrategy implements DiscoveryStrategy, EventSubscriberInt
...
+        static::$asyncClient = $asyncClient;
     }
 
     /**
      * {@inheritdoc}
      */
     public static function getCandidates($type)
     {
-        if (static::$client !== null && $type == HttpClient::class) {
+        if (static::$client !== null && $type === HttpClient::class) {
```
sagikazarmark: Should the type check rather come first? This way the first condition is evaluated even if the type is a MessageFactory.

Resources/config/services.xml
```diff
@@ -5,7 +5,6 @@
 
     <services>
         <service id="httplug.strategy" class="Http\HttplugBundle\Discovery\ConfiguredClientsStrategy" public="true">
```
sagikazarmark: Is public necessary here?

Nyholm: Yes, all event subscribers have to be public.

sagikazarmark: But they are public by default, aren't they?

Nyholm: Ah, correct. 

`public="true"` is not necessary .

PR #89, opened 2016-07-18
Do not run travis on dev-* branches.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

The `dev-*` branches are used for development and are most likely to end up in a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ env:
 branches:
     except:
         - /^analysis-.*$/
+        - /^dev-.*$/
 
 matrix:
     fast_finish: true


xabbuh:
I think it would be better to base pull requests on branches that are created in forks of this repository. Otherwise, all the branches that are pushed here end up as dev releases on Packagist too.

sagikazarmark:
Not sure about this. I don't really care if things are built twice.

Nyholm:
I wanted to reduce the time waiting on Travis. 

But @xabbuh, you are correct.

PR #90, opened 2016-07-18
Move the exception symbol one step up in the stack

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | https://github.com/php-http/documentation/pull/118
| License         | MIT

While making screenshots for the documentation I noticed that the exception symbol was misplaced. Consider the screenshot below. 

When an exception is thrown at the error plugin we should have the big red X next to it. Before this PR the symbol appeared first at the content_length plugin. 

error-plugin-failure


Nyholm:
Thank you for merging

Issue #95, opened 2016-07-19
Improve HttplugExtension


sagikazarmark:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.2.x

#### Actual Behavior

The extension became quite unstable with the 1.2.0 release. Profiling toolbar is not so well separated (enabled in production mode, etc). This has to be improved.

PR #96, opened 2016-07-19
Fix debug tool registration

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Fixes the bundle in production mode.

#### Why?

In production mode (and when debug mode is disabled) debug tools should not be registered in the clients, otherwise we get weird service not found errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Change Log
 
 
+## 1.2.2 - 2016-07-19
+
+### Fixed
+
+- Do not register debug tools when debugging is disabled (eg. in prod mode)
+
+
 ## 1.2.1 - 2016-07-19
 
 ### Fixed
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -39,10 +39,10 @@ class HttplugExtension extends Extension
         $loader->load('services.xml');
         $loader->load('plugins.xml');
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Change Log
 
 
+## 1.2.2 - 2016-07-19
+
+### Fixed
+
+- Do not register debug tools debugging is disabled (eg. in prod mode)
```
dbu: ...debug tools _when_ debugging is...

dbu:
good job, looks right to me. did you try it out?

wonder if we could (not in this PR, but in general) add integration tests for both dev and prod mode to make sure we don't break this again.

sagikazarmark:
Yes we should, along with this: https://github.com/php-http/HttplugBundle/issues/95

dbu:
didn't this PR fix that the toolbar is enabled in prod mode?

sagikazarmark:
`$toolbar = is_bool($config['toolbar']['enabled']) ? $config['toolbar']['enabled'] : $container->hasParameter('kernel.debug') && $container->getParameter('kernel.debug');`

This is the code that decides whether we have a toolbar or not

sagikazarmark:
The problem was that this configuration was not applied in all cases when some debug service was registered/linked.

PR #102, opened 2016-07-25
Added stopwatch plugin

### LINKED ISSUE #92: Integrate with Symfony Debug Timeline
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | n/a

Looking at @csarrazi's [Guzzle Bundle](https://github.com/csarrazi/CsaGuzzleBundle) they have a really nice integration with the Symfony Debug timeline. That is something we should support as well.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #92
| Documentation   | 
| License         | MIT

#### What's in this PR?

With this PR we integrate with the WebProfiler timeline

screen shot 2016-07-25 at 03 01 39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Added
 
 - Support for BatchClient
+- The stopwatch plugin in included by default when using profiling. 
 
 ### Changed
 
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -218,8 +218,10 @@ class HttplugExtension extends Extension
         $pluginClientOptions = [];
 
         if ($profiling) {
-            // Add the stopwatch plugin
-            array_unshift($arguments['plugins'], 'httplug.plugin.stopwatch');
+            if (!in_array('httplug.plugin.stopwatch', $arguments['plugins'])) {
+                // Add the stopwatch plugin
+                array_unshift($arguments['plugins'], 'httplug.plugin.stopwatch');
+            }
 
             // Tell the plugin journal what plugins we used
             $container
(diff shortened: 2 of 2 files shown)


sagikazarmark:
What happens if a custom stopwatch (either the httplug service or a completely custom one) is added to the plugin stack?

Nyholm:
If someone uses the default stopwatchPlugin we are doing the same calls to the stopwatch component twice. That means that our measured times would be twice as large. 

If someone uses the stopwatchPlugin with a custom stopwatch, then nothing special will happen.

sagikazarmark:
Okay, so probably we should add a check to see if the stopwatch plugin is already added? And then do what? Leave it as is, or remove that one and add our own to the beginning of the stack?

joelwurtz:
Or just a check in the StopwatchPlugin to do nothing if the event is already started ? (there is a `isStarted` method in stopwatch)

Nyholm:
Our (the automatic one) stopwatch plugin is the first to execute though.

Nyholm:
But why would someone add their own stopwatch plugin? I we could figure out that use case it may be easier to know the appropriate action to take.

Nyholm:
I answering my own question now. The reason you may want to add the stopwatch plugin may be because you want to change the order. Say you have one slow plugin in the beginning that you want to exclude for the stopwatch. 

I believe we should check if the stopwatch is added and only prepend the plugin list if the stopwatch is missing. What do you think?

sagikazarmark:
Like that, but maybe we should only check our stopwatch plugin service. If someone creates a custom stopwatch plugin service, it might not use the same stopwatch instance. WDYT?

Nyholm:
Agreed.

dbu:
looks good!

can you please rebase, there seems to be a conflict now.

Nyholm:
Sorry, 

I've now rebased the PR

dbu:
thanks!

Nyholm:
Thank you for merging

PR #105, opened 2016-07-25
Rename toolbar config to profiling

### LINKED ISSUE #99: Rename toolbar config to profiling
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 2.0.0

#### Actual Behavior

Currently we have the config called `toolbar`.

#### Expected Behavior

`profiling` seems to be a common standard to me. See symfony and DoctrineBundle.

This could be done supporting BC in the next minor, then remove `toolbar` in the next major.

sagikazarmark:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #99
| License         | MIT

#### What's in this PR?

Rename toolbar config to profiling as it seems to be more common.


DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,26 @@ public function getConfigTreeBuilder()
...
+            ->beforeNormalization() // @deprecated toolbar in 1.3.0, use profiling instead
+                ->ifTrue(function ($v) {
+                    return is_array($v) && array_key_exists('toolbar', $v) && is_array($v['toolbar']);
+                })
+                ->then(function ($v) {
+                    if (array_key_exists('enabled', $v['toolbar']) && 'auto' === $v['toolbar']['enabled']) {
+                        $v['toolbar']['enabled'] = $this->debug; // @deprecated value auto in 1.3.0
+                    }
+
+                    if (array_key_exists('profiling', $v) && is_array($v['profiling'])) {
```
dbu: imho we should consider configuration invalid if both profiling and toolbar exist. the user should decide for one.

do we have a problem at this stage with default values from the Configuration? i think they will only come after this step.

Tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -95,6 +95,7 @@ public function testEmptyConfiguration()
             'config/empty.yml',
             'config/empty.xml',
             'config/empty.php',
+            'config/toolbar.yml',
```
dbu: maybe have a dedicated BC test that makes sure old configuration results in the right new configuration. then we can also easily remove that test when we bump the major version.

DependencyInjection/Configuration.php
sagikazarmark: I think the BC promise at this point is that toolbar will work at least until 2.0 which means if toolbar is configured it should always work, otherwise it's a BC break, no matter if an error is thrown or overwritten by profiling.

What about default values? you mean the auto config? I added it here so that the enabled config can be defined as boolean.

dbu: imo the BC promise covers that you can update the bundle and it continues to work as before. but once you add new configuration that was not possible before the update, it is ok to complain when you leave the old config in place. merging old and new config feels strange, it would mean you can configure both at the same time.

with auto config i meant `->defaultValue()` places. i think those come after beforeNormalization - if i am wrong and they are already present, we indeed would need to merge.

sagikazarmark: Okay, so if the new config exists: complain or simply ignore the old config?

I still don't get the default value thing: yes, defaultValue comes after the normalization, but we need to check the auto value for BC in the normalization section.

Nyholm: > Okay, so if the new config exists: complain or simply ignore the old config?

I agree with David. If you configure both `toolbar` and `profiling` you should get an error.

dbu: i was not specifically talking about the auto thingy, but in general
about default values leading to profiling and toolbar both existing,
even if the user only specified one. but it looks like the only
situation when both exists at this point is when the user actually
specified values for both. in that case we should throw an exception.
its invalid configuration to mix the new and the old name. i see no BC
problem because it means the user just added new configuration that
would have been forbidden with the current release of the bundle, so the
intention was to update the config.

sagikazarmark:
Added exception on colliding config and separated BC tests.

DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,26 @@ public function getConfigTreeBuilder()
                     return $v;
                 })
             ->end()
+            ->beforeNormalization() // @deprecated toolbar in 1.3.0, use profiling instead
+                ->ifTrue(function ($v) {
+                    return is_array($v) && array_key_exists('toolbar', $v) && is_array($v['toolbar']);
+                })
+                ->then(function ($v) {
+                    if (array_key_exists('profiling', $v)) {
+                        throw new InvalidConfigurationException('"toolbar" config is deprecated as of version 1.3.0, please use "profiling" instead.');
```
dbu: lets change the message: "Can't configure both "toolbar" and "profiling" section. The "toolbar" config is deprecated as of version 1.3.0, please only use "profiling"."

DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,26 @@ public function getConfigTreeBuilder()
...
+                    }
+
+                    if (array_key_exists('enabled', $v['toolbar']) && 'auto' === $v['toolbar']['enabled']) {
+                        $v['toolbar']['enabled'] = $this->debug; // @deprecated value auto in 1.3.0
```
dbu: `@trigger_error`?

sagikazarmark:
Rewritten exception message, added trigger errors.

DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,29 @@ public function getConfigTreeBuilder()
                     return $v;
                 })
             ->end()
+            ->beforeNormalization()
+                ->ifTrue(function ($v) {
+                    @trigger_error('"httplug.toolbar" config is deprecated since version 1.3 and will be removed in 2.0. Use "httplug.profiling" instead.', E_USER_DEPRECATED);
```
sagikazarmark: You are right, done.

dbu:
can you please also update the documentation? with a ..versionadded note.

PR #111, opened 2016-08-06
Add support for default_host and force_host options on each client

### LINKED ISSUE #110: configure base path for a client
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes

#### Actual Behavior

To configure a client with a default host, we can use the AddHostPlugin. However, this plugin can not be configured within HttplugBundle as far as i can see. I ended up doing this:

```yaml
parameters:
    server: "http://localhost:8001"

services:
    app.http.todo_hostname_uri:
        class: Psr\Http\Message\UriInterface
        factory: ["@httplug.uri_factory", createUri]
        arguments: ["%server%"]
    app.http.plugin.todo_hostname:
        class: Http\Client\Common\Plugin\AddHostPlugin
        arguments: ["@app.http.todo_hostname_uri"]
```

```yaml
httplug:
    clients:
        todo:
            factory: "httplug.factory.guzzle6"
            plugins: ["app.http.plugin.todo_hostname", "httplug.plugin.error"]
```

#### Possible Solutions

Not sure, but it should be easier than this imo. AddHostPlugin typehints UriInterface, and afaik we can't use a factory for an argument, hence we need to define a service for the uri object.

We could add a configuration in the `clients.{name}` space for default_host and force_host but i am not sure how well this scales with other options. maybe an options subsection?

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #110
| Documentation   | TODO (want feedback first)
| License         | MIT

#### What's in this PR?

Add new configuration options for default_host and force_host for each client.

#### Why?

See #110

#### Example Usage

``` yaml
    clients:
        test:
            factory: httplug.factory.guzzle6
            http_methods_client: true
            options:
                default_host: http://localhost
```

You can specify the httplug.client.{name}.default|force_host_plugin in the plugins list if you dont want the host plugin to come last in the chain.

#### Checklist

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -5,6 +5,7 @@ namespace Http\HttplugBundle\DependencyInjection;
 use Http\Client\Common\BatchClient;
 use Http\Client\Common\FlexibleHttpClient;
 use Http\Client\Common\HttpMethodsClient;
+use Http\Client\Common\Plugin\AddHostPlugin;
 use Http\Client\Common\Plugin\AuthenticationPlugin;
 use Http\Discovery\HttpAsyncClientDiscovery;
 use Http\Discovery\HttpClientDiscovery;
@@ -14,6 +15,7 @@ use Http\HttplugBundle\Collector\DebugPlugin;
 use Http\Message\Authentication\BasicAuth;
 use Http\Message\Authentication\Bearer;
 use Http\Message\Authentication\Wsse;
+use Psr\Http\Message\UriInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
@@ -215,6 +217,7 @@ class HttplugExtension extends Extension
     {
         $serviceId = 'httplug.client.'.$name;
 
+        $plugins = $arguments['plugins'];
         $pluginClientOptions = [];
 
         if ($profiling) {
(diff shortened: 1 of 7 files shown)


dbu:
just tested it in the web summercamp tutorial app and the configuration works as expected.

DependencyInjection/HttplugExtension.php
```diff
@@ -234,6 +237,40 @@ private function configureClient(ContainerBuilder $container, $name, array $argu
...
+                        $uriService = $serviceId.'.force_host_uri';
+                        $addHostPlugin = $serviceId.'.force_host_plugin';
+                        $this->createUri($container, $uriService, $value);
+                        $container
+                            ->register($addHostPlugin, AddHostPlugin::class)
+                            ->setPublic(false)
+                            ->addArgument(new Reference($uriService))
+                            ->addArgument(['replace' => true])
+                        ;
+                        if (!in_array($addHostPlugin, $plugins)) {
```
dbu: by default, we append the AddHostPlugin to the very end of the chain, but you can explicitly set it in the plugins list. i will explain this in the documentation

Nyholm:
I do not like the new `options` key. I get questions like: Why isn't `http_methods_client` under `options`? How come we have both `options` and `config`?

``` yaml
    clients:
        test:
            factory: httplug.factory.guzzle6
            http_methods_client: true
            options:
                default_host: http://localhost
           config: 
               verify: false
```

Also, we should support other plugins like `HeaderAppend`, `HeaderDefaults`, `HeaderRemove` and `HeaderSet`. All these plugins (and `AddHost`) are very client specific. Ie, you are not likely to have the same host configured in the AddHost plugin for all clients. Compared to the CachePlugin which is very likely that you use for all clients. 

Possible solution: 
Maybe we should introduce two scopes of plugins. Public and Private? See this config: 

``` yaml
httplug:
    plugins: # Public plugins
        cache:
            enabled: true
            cache_pool: 'my_cache_pool'
            stream_factory: 'httplug.stream_factory'
            config:
                default_ttl: 3600
                respect_cache_headers: true
        logger:
            enabled: true
            logger: 'logger'
            formatter: null
        redirect:
            preserve_header: true
            use_default_for_multiple: true
    clients:
        test:
            factory: httplug.factory.guzzle6
            http_methods_client: true
            config: 
                verify: false
           plugins:  # Private plugins only for this client
                add_host:
                    host: http://localhost
                    replace: true
               header_append:
                   User-Agent: 'Foobar'
```

dbu:
i like that idea a lot! i guess the basic question is how much we want the user to know things like adding the host are plugins, but we already went with making this transparent for the general plugins. for people not familiar with the lib, there will be a tiny confusion why something is a config and something else is an extra plugin, but i think its manageable with doc.

i guess we should also repeat the public plugins in the extra_plugins of a client. so that you can have a different redirect/log/cache/whatever behaviour. and i will keep the behaviour that you can change the order by adding the private plugins in the plugins list of the client.

dbu:
see #112

PR #112, opened 2016-08-07
Allow to configure plugins specifically for a client

### LINKED ISSUE #110: configure base path for a client

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #110
| Documentation   | https://github.com/php-http/documentation/pull/142
| License         | MIT

#### What's in this PR?

Add new configuration for plugins specifically to a client. The global plugin configuration is duplicated to allow configuring all plugins with client specific information.

#### Why?

See #110

#### Example Usage

``` yaml
httplug:
    plugins:
        logger:
            enabled: true
            logger: 'logger'
            formatter: null
        redirect:
            enabled: true
            preserve_header: true
            use_default_for_multiple: true
        retry:
            enabled: true
            retry: 1
    clients:
        todo:
            factory: "httplug.factory.guzzle6"
            plugins:
                - 'httplug.plugin.logger'
                - 'httplug.plugin.redirect'
                -  add_host:
                     host: http://localhost
                - 'httplug.plugin.retry'
        acme:
            plugins: ['httplug.plugin.redirect', 'httplug.plugin.retry']
```

You can specify the httplug.client.{name}.plugin.add_host in the plugins list if you want the host plugin to come somewhere else than last in the chain.

#### Checklist

- [x] Add the other plugins we currently dont support (want feedback first)
  - [x] HeaderSet
  - [x] HeaderRemove
  - [x] HeaderDefaults 
  - [x] HeaderAppend
  - [x] -ContentLength- (configured as shared plugin, has no options)
- [x] Handle custom authentication plugin
- [x] Fix XML configuration
- [x] Documentation pull request created => https://github.com/php-http/documentation/pull/142

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Added
 
+- You can now also configure client specific plugins in the `plugins` option of a client.
+  Some plugins that you previously had to define in your own services can now be configured on the client.
 - Support for BatchClient
 - The stopwatch plugin in included by default when using profiling.
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -2,8 +2,8 @@
 
 namespace Http\HttplugBundle\DependencyInjection;
 
-use Symfony\Component\Config\Definition\ArrayNode;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+use Symfony\Component\Config\Definition\Builder\NodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
@@ -45,7 +45,7 @@ class Configuration implements ConfigurationInterface
         $rootNode = $treeBuilder->root('httplug');
 
         $this->configureClients($rootNode);
-        $this->configurePlugins($rootNode);
+        $this->configureSharedPlugins($rootNode);
 
         $rootNode
             ->validate()
@@ -167,6 +167,7 @@ class Configuration implements ConfigurationInterface
                     ->thenInvalid('A http client can\'t be decorated with both FlexibleHttpClient and HttpMethodsClient. Only one of the following options can be true. ("flexible_client", "http_methods_client")')->end()
                 ->useAttributeAsKey('name')
                 ->prototype('array')
+                ->fixXmlConfig('plugin')
                 ->children()
                     ->scalarNode('factory')
                         ->isRequired()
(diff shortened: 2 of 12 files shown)


DependencyInjection/Configuration.php
```diff
@@ -189,131 +191,215 @@ protected function configureClients(ArrayNodeDefinition $root)
...
-                                ->cannotBeEmpty()
-                            ->end()
-                        ->end()
-                    ->end() // End stopwatch plugin
+        $decoder = $children->arrayNode('decoder');
+        if ($disableAll) {
+            $decoder->canBeEnabled();
+        } else {
+            $decoder->canBeDisabled();
+        }
```
dbu: this construct is repeated quite a bit, but i can't think of a better solution. the global plugins need to be enabled by default, but on the client, we want to only see plugins that where explicitly configured.

DependencyInjection/Configuration.php
```diff
@@ -189,131 +191,215 @@ protected function configureClients(ArrayNodeDefinition $root)
...
+
+        $retry = $children->arrayNode('retry');
+        if ($disableAll) {
+            $retry->canBeEnabled();
+        } else {
+            $retry->canBeDisabled();
+        }
+        $retry->addDefaultsIfNotSet()
+            ->children()
+                ->scalarNode('retry')->defaultValue(1)->end() // TODO: should be called retries for consistency with the class
```
dbu: i will do this after the PR has been merged, with a BC handling of course

DependencyInjection/HttplugExtension.php
```diff
@@ -107,32 +109,43 @@ private function configureClients(ContainerBuilder $container, array $config)
...
      * @param array            $config
+     * @param string           $idPrefix  Start of service id for these plugins.
      */
-    private function configurePlugins(ContainerBuilder $container, array $config)
+    private function configurePlugins(ContainerBuilder $container, array $config, $idPrefix = 'httplug.plugin')
     {
+        $sharedPluginPrefix = 'httplug.plugin';
+        $shared = $sharedPluginPrefix === $idPrefix;
         if (!empty($config['authentication'])) {
+            // TODO: handle extra auth plugin on client
```
dbu: still need to do this

Resources/config/plugins.xml
```diff
@@ -27,5 +27,13 @@
         <service id="httplug.plugin.stopwatch" class="Http\Client\Common\Plugin\StopwatchPlugin" public="false">
             <argument />
         </service>
+
+        <!-- extra plugin definitions -->
+
+        <service id="httplug.plugin.add_host" class="Http\Client\Common\Plugin\AddHostPlugin" public="false" abstract="true">
```
dbu: abstract=true makes symfony remove the definition template when the container is built.

DependencyInjection/HttplugExtension.php
```diff
@@ -145,7 +158,9 @@ private function configurePluginByName($name, Definition $definition, array $con
                 $definition->replaceArgument(0, new Reference($config['cookie_jar']));
                 break;
             case 'decoder':
-                $definition->addArgument($config['use_content_encoding']);
+                $definition->addArgument([
+                    'use_content_encoding' => $config['use_content_encoding'],
```
dbu: separated the bugfix into #113 and then realized we have other changes in master already, so we won't do a patch release for the fix. but i guess after we merge this we want to tag 1.3 very soon so the bugfix will be released soon.

DependencyInjection/HttplugExtension.php
```diff
@@ -283,6 +326,23 @@ function ($id) {
     }
 
     /**
+     * Create a URI object with the default URI factory.
+     *
+     * @param ContainerBuilder $container
+     * @param string           $serviceId Name of the private service to create
+     * @param string           $uri       String representation of the URI
+     */
+    private function createUri(ContainerBuilder $container, $serviceId, $uri)
```
dbu: should we use an abstract service that we clone and update instead of this?

Tests/Resources/app/config/config_php.yml
```diff
@@ -3,4 +3,13 @@ framework:
   test: ~
```
dbu: will need to rebase once #115 is merged and do this config globally

sagikazarmark:
How can you control the order of plugins?

Maybe we can have this within the plugins section: numeric index: look for service, string index: configure extra plugin.

dbu:
without further information, extra_plugins are appended to the explicitly configured plugins in the order they have been defined. but you can specify each of the extra_plugin in the `plugins` field of the client. i will document how to know the service name of the extra plugins.

dbu:
> Maybe we can have this within the plugins section: numeric index: look
> for service, string index: configure extra plugin.

ah, now i see what you mean. if we want to mix, i would rather say that
inside client.{name}.plugins you can have string (= service id) or extra
plugin configuration. this will make it easier to handle the ordering.
the drawback would be that the configuration format is different from
shared plugins, we would need to add a type field.

or we have something like this:

```
plugins:
    logger:
        ...
client:
    acme:
        plugins:
            - add_host:
                  host: ...
            - httplug.plugin.logger
```

i think this last option is probably the most user friendly. (but i am
scared how to express that in the configuration definition :-( )

Nyholm:
I like this approach: 

```yaml
plugins:
    logger:
        ...
client:
    acme:
        plugins:
            - add_host:
                  host: ...
            - httplug.plugin.logger
```

It makes more sense. But yes, the configuration.php is going to be difficult. But we should try.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Added
 
+- You can now configure plugins on each client in `extra_plugins`. Plugins that you previously had to define your own services can be configured on the client.
```
dbu: needs update

DependencyInjection/HttplugExtension.php
```diff
@@ -242,6 +261,24 @@ private function configureClient(ContainerBuilder $container, $name, array $argu
             $pluginClientOptions['debug_plugins'] = [new Reference($debugPluginServiceId)];
         }
 
+        $plugins = [];
+        foreach ($arguments['plugins'] as $plugin) {
+            list($name, $pluginConfig) = each($plugin);
+            if ('reference' === $name) {
+                $plugins[] = $pluginConfig['id'];
+            } elseif ('authentication' === $name) {
+                // TODO handle custom authentication
```
dbu: still needs to be done

Tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -192,7 +209,7 @@ public function testSupportsAllConfigFormats()
             return __DIR__.'/../../Resources/Fixtures/'.$path;
         }, [
             'config/full.yml',
-            'config/full.xml',
+// TODO fix xml config            'config/full.xml',
```
dbu: i can't figure out how to write the xml config to make it work. any ideas? an example of a bundle having a similar conf format?

DependencyInjection/Configuration.php
```diff
@@ -185,136 +187,230 @@ private function configureClients(ArrayNodeDefinition $root)
...
+                ->end()
 
-                    ->arrayNode('retry')
-                    ->canBeDisabled()
-                    ->addDefaultsIfNotSet()
-                        ->children()
-                            ->scalarNode('retry')->defaultValue(1)->end()
-                        ->end()
-                    ->end() // End retry plugin
+                // TODO add aditional plugins that are only usable on a specific client
```
dbu: what are the plugins we should allow to configure?

DependencyInjection/HttplugExtension.php
```diff
@@ -228,7 +243,11 @@ private function configureClient(ContainerBuilder $container, $name, array $argu
         if ($profiling) {
             if (!in_array('httplug.plugin.stopwatch', $arguments['plugins'])) {
                 // Add the stopwatch plugin
-                array_unshift($arguments['plugins'], 'httplug.plugin.stopwatch');
+                array_unshift($arguments['plugins'], [
+                    'reference' => [
+                        'id' => 'httplug.plugin.stopwatch',
+                    ],
+                ]);
             }
```
dbu: TODO: handle more plugins

Nyholm:
The plugin_journal expects a single dimensional array at https://github.com/php-http/HttplugBundle/pull/112/files#diff-18f3f26a96b967c0a5a3e13086d8367dR256

(Try to open the web profiler page and you will see an error)

DependencyInjection/Configuration.php
Nyholm: HeaderSet, HeaderRemove, HeaderDefaults, HeaderAppend, ContentLength

DependencyInjection/HttplugExtension.php
```diff
@@ -70,8 +72,8 @@ public function load(array $configs, ContainerBuilder $container)
             ;
         }
 
-        $this->configurePlugins($container, $config['plugins']);
         $this->configureClients($container, $config);
+        $this->configureSharedPlugins($container, $config['plugins']); // must be after clients, as the extra_plugins in clients might use plugins as template that will be removed
```
Nyholm: Update this comment. We cant use (extra_plugins)

Nyholm:
I've reviewed this. It is a really good improvement. I like the way you add the shared configuration on each client in the Configuration.php. 

I made a few comments, It was a bug but otherwise it looks fine. 

----------

For the record. Here is a working configuration (we do not use `extra_plugins` as the first comment says): 

```yaml

httplug:
    plugins:
        logger:
            enabled: true
            logger: 'logger'
            formatter: null
        redirect:
            enabled: true
            preserve_header: true
            use_default_for_multiple: true
        retry:
            enabled: true
            retry: 1
    clients:
        todo:
            factory: "httplug.factory.guzzle6"
            plugins:
                - 'httplug.plugin.logger'
                - 'httplug.plugin.redirect'
                -  add_host:
                     host: http://localhost
                - 'httplug.plugin.retry'
        acme:
            plugins: ['httplug.plugin.redirect', 'httplug.plugin.retry']
```

dbu:
adding a new test in #118 to prevent the regression you spotted.

DependencyInjection/Configuration.php
```diff
@@ -153,8 +153,9 @@ public function getConfigTreeBuilder()
 
     private function configureClients(ArrayNodeDefinition $root)
     {
-        $root->children()
+        $pluginNode = $root->children()
             ->arrayNode('clients')
+                ->fixXmlConfig('plugin')
```
dbu: this must go after the `->prototype('array')` array line to apply to the things inside the array and not to the array itself (which has numerical keys)

DependencyInjection/Configuration.php
dbu: ContentLength is globally available and has no configuration options. did you mean something else?

Nyholm: No, my bad.

dbu:
yay, this seems to actually work out! trying to get the doc started, then lets merge this.

dbu:
this is now ready. i quickly tried in a symfony app and it works now fine, including debug toolbar.

Nyholm:
I've started tested it a little. Found a unrelated bug... (#121)

I'll will try to do a final review tonight if you are happy with the current state of this PR.

dbu:
great! yes i am happy with it. squashed the commits and it does look right to me. but am sure there are things to improve and i might have missed something.

DependencyInjection/Configuration.php
```diff
@@ -185,136 +186,280 @@ private function configureClients(ArrayNodeDefinition $root)
...
-                        ->info('A list of service ids of plugins. The order is important.')
-                        ->prototype('scalar')->end()
-                    ->end()
                     ->variableNode('config')->defaultValue([])->end()
-                ->end()
-            ->end();
+                    ->arrayNode('plugins')
+                        ->info('A list of plugin service ids and client specific plugin definitions. The order is important.')
+                        ->prototype('array')
+        ;
```
Nyholm: You do not call `->end()` here and I know that it is on purpose. You should at least add a comment here. Or event better: Create the plugin node and then add it to the tree. (I think that is possible)

DependencyInjection/Configuration.php
```diff
@@ -185,136 +186,280 @@ private function configureClients(ArrayNodeDefinition $root)
...
-        $root->children()
-            ->arrayNode('plugins')
+        $pluginsNode = $root
+            ->children()
+                ->arrayNode('plugins')
+                ->info('Global plugin configuration. Plugins need to be explicitly added to clients.')
                 ->addDefaultsIfNotSet()
-                ->children()
-                    ->append($this->addAuthenticationPluiginNode())
+        ;
```
Nyholm: Same here, you do not call `->end()`

DependencyInjection/Configuration.php
```diff
@@ -185,136 +186,280 @@ private function configureClients(ArrayNodeDefinition $root)
...
+                        ->scalarNode('host')
+                            ->info('Host name including protocol and optionally the port number, e.g. https://api.local:8000')
+                            ->isRequired()
+                            ->cannotBeEmpty()
                         ->end()
-                    ->end() // End cache plugin
-
-                    ->arrayNode('cookie')
+                        ->scalarNode('replace')
+                            ->info('Whether to replace the host if request already specifies it')
```
Nyholm: s/it/one/ ?

DependencyInjection/Configuration.php
```diff
@@ -185,136 +186,280 @@ private function configureClients(ArrayNodeDefinition $root)
...
-                            ->end()
+                    ->end()
+                ->end()
+                ->arrayNode('header_remove')
+                    ->canBeEnabled()
+                    ->info('Remove headers from requests.')
+                    ->fixXmlConfig('header')
+                    ->children()
+                        ->arrayNode('headers')
+                            ->info('List of header names to remove')
```
Nyholm: Why is `->normalizeKeys(false)` needed for all *HeaderPlugins except for header_remove?

DependencyInjection/Configuration.php
```diff
@@ -185,136 +186,280 @@ private function configureClients(ArrayNodeDefinition $root)
...
+                    ->scalarNode('cookie_jar')
+                        ->info('This must be a service id to a service implementing Http\Message\CookieJar')
+                        ->isRequired()
+                        ->cannotBeEmpty()
+                    ->end()
+                ->end()
+            ->end();
+        // End cookie plugin
+
+        $decoder = $children->arrayNode('decoder');
```
Nyholm: I like how you separated the plugins.

DependencyInjection/HttplugExtension.php
Nyholm: I do not mind either way. Let's leave it.

DependencyInjection/HttplugExtension.php
```diff
@@ -307,7 +370,7 @@ private function configureAutoDiscoveryClients(ContainerBuilder $container, arra
                     $container,
                     'auto_discovered_client',
                     [HttpClientDiscovery::class, 'find'],
-                    $config['profiling']['enabled']
+                    $this->isConfigEnabled($container, $config['profiling'])
```
Nyholm: I've never seen this. Why is it needed? Is it to make sure we do not get "undefined key 'enabled'"?

Nyholm:
Im happy to merge when all my comments have been addressed or answered. 

Good job 👍

DependencyInjection/Configuration.php
dbu: because for the others, the header names are the keys, while here the headers are the values. only the keys are normalized, never the values.

DependencyInjection/HttplugExtension.php
dbu: its the counterpart to canBeEnabled/canBeDisabled. i feel when we use canBeEnabled/Disabled we should also use this method.

Nyholm: Okey, Cool

dbu:
did some more cleanup based on your feedback, @Nyholm

any more suggestions or is this good?

Nyholm:
Im all good. I'm very happy with this PR. It is an amazing improvement. Thank you

PR #118, opened 2016-08-12
test that journal is set up correctly

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no (only tests added)
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Test to prevent regression in #112
| Documentation   | -
| License         | MIT

#### What's in this PR?

A test about the debug toolbar because tests in #112 where green but the toolbar broken.

#### Why?

We should prevent regressions.

### DIFF
--- a/Tests/Functional/ServiceInstantiationTest.php
+++ b/Tests/Functional/ServiceInstantiationTest.php
@@ -2,7 +2,11 @@
 
 namespace Http\HttplugBundle\Tests\Functional;
 
+use Http\Client\HttpClient;
+use Http\HttplugBundle\Collector\DebugPluginCollector;
+use Http\HttplugBundle\Collector\PluginJournal;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Symfony\Component\HttpKernel\Profiler\Profiler;
 
 class ServiceInstantiationTest extends WebTestCase
 {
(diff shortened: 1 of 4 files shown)


Tests/Functional/ServiceInstantiationTest.php
```diff
@@ -21,6 +25,25 @@ public function testHttpClientNoDebug()
...
+        $profiler = $container->get('profiler');
+        $this->assertInstanceOf(Profiler::class, $profiler);
+        $this->assertTrue($profiler->has('httplug'));
+        $collector = $profiler->get('httplug');
+        $this->assertInstanceOf(DebugPluginCollector::class, $collector);
+        /** @var PluginJournal $journal */
+        $journal = $collector->getJournal();
+        $plugins = $journal->getPlugins('acme');
+        $this->assertEquals([
+            'httplug.plugin.stopwatch',
```
dbu: @Nyholm its expected that the journal also tracks the stopwatch, right? or should we hide that from the journal?

Nyholm: Yes, The stopwatch plugin should always be available when we do profiling. 

It should not be hidden from the journal.

Nyholm:
Awesome @dbu. Thank you!

PR #122, opened 2016-09-12
DummyClient should implement HttpClient, HttpAsyncClient

mekras:
| Q               | A
| --------------- | ---
| Bug fix?        | not sure
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Some IDEs use Symfony container for static analysis and code autocomplete. But DummyClient does not provide any methods. Because of this, IDE shows errors:

IDE shows error

This PR adds to `DummyClient` dummy implementation of `HttpClient` and `HttpAsyncClient` to fix IDE behavior.

#### Checklist

### DIFF
--- a/ClientFactory/DummyClient.php
+++ b/ClientFactory/DummyClient.php
@@ -2,11 +2,22 @@
 
 namespace Http\HttplugBundle\ClientFactory;
 
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Psr\Http\Message\RequestInterface;
+
 /**
  * This client is used as a placeholder for the dependency injection. It will never be used.
  *
  * @author Tobias Nyholm <tobias.nyholm@gmail.com>
  */
-class DummyClient
+class DummyClient implements HttpClient, HttpAsyncClient
 {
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you! I agree, it should implement HttpClient, even though it is never used.

sagikazarmark:
Actually: is this class necessary? Can't we just use an interface in the `class` parameter of the service definition?

Nyholm:
It definitely has to be **something**. Im not sure if you can use an interface. However, what interface do we have that implements both HttpClient and HttpAsyncClient?

dbu:
i am pretty sure an interface would work in the class parameter, if it is never actually tried to be instantiated - which would be wrong anyways. we could change the type of DummyClient to interface and then it can implement both interfaces without needing to add empty methods. and it can't be accidentally implemented.

Nyholm:
> We could change the type of DummyClient to interface and then it can implement both interfaces without needing to add empty methods. and it can't be accidentally implemented.

👍 @mekras, can you try that?

mekras:
@Nyholm, why not? But not now. May be later today.

PR #124, opened 2016-11-25
Remove non-existent dependency

spyl94:


Nyholm:
Thank you for the contribution. However, that is a virtual package. It makes sure you have at least one client installed. 
Read more about it at http://docs.php-http.org/en/latest/httplug/users.html

spyl94:
Indeed! Despite the same PR has been merged into laravel-httplug (https://github.com/php-http/laravel-httplug/commit/fae3d1b2a653f645cbe09f32aaff4bab34de14fc).
I will update Readme to add some informations for developers who don't read the docs!

Nyholm:
Thank you. I was not aware of that. 
I have opened [an issue](https://github.com/php-http/laravel-httplug/issues/13) for that. Maybe there was a good reason for doing so in Laravel.

PR #125, opened 2016-11-25
Add information about client installation

spyl94:
`php-http/client-implementation` makes sure we install a client otherwise composer installation fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Via Composer
 ``` bash
 $ composer require php-http/httplug-bundle
 ```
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you. That is great. 
What do you think, should we link to the documentation instead of the repositories?

http://docs.php-http.org/en/latest/clients/guzzle6-adapter.html
http://docs.php-http.org/en/latest/clients/curl-client.html

xabbuh:
Or simply link to http://docs.php-http.org/en/latest/clients.html with the list of available adapters?

spyl94:
Done!

dbu:
thanks!

PR #129, opened 2017-02-16
Stop running tests for symfony 3.0 which is EOL

fbourigault:
Stop testing the bundle against symfony/symfony 3.0 which is EOL and incompatible with twig/twig 2.x.

I also added symfony/symfony 3.2 in the matrix :)

Related symfony issue: https://github.com/symfony/symfony/issues/20284


Nyholm:
Thank you

PR #134, opened 2017-02-22
Make sure we always have a stack

### LINKED ISSUE #133: There's not always a Stack on the Collector
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.4.0

In some cases the call to `getCurrentStack()` on the collector can return `false`. https://github.com/php-http/HttplugBundle/blob/master/Collector/Collector.php#L67-L70
This happens when there has not been any `Stack` added to the `Collector` and `end()` only returns `
false`

This results in a break when it's not taken in consideration here: https://github.com/php-http/HttplugBundle/blob/master/Collector/ProfilePlugin.php#L61

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #133
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

If we (for some reason) do not have a stack, it should not crash. Just pretend everything is normal. 

@dkvk, does this patch solve your issue?

### DIFF
--- a/Collector/ProfilePlugin.php
+++ b/Collector/ProfilePlugin.php
@@ -58,7 +58,9 @@ class ProfilePlugin implements Plugin
     {
         $profile = new Profile($this->pluginName, $this->formatter->formatRequest($request));
 
-        $this->collector->getCurrentStack()->addProfile($profile);
+        if (false !== $stack = $this->collector->getCurrentStack()) {
+            $stack->addProfile($profile);
+        }
 
         return $this->plugin->handleRequest($request, $next, $first)->then(function (ResponseInterface $response) use ($profile) {
             $profile->setResponse($this->formatter->formatResponse($response));
--- a/Collector/ProfilePlugin.php
+++ b/Collector/ProfilePlugin.php
@@ -58,7 +58,7 @@ class ProfilePlugin implements Plugin
     {
         $profile = new Profile($this->pluginName, $this->formatter->formatRequest($request));
 
-        if (false !== $stack = $this->collector->getCurrentStack()) {
+        if ($stack = $this->collector->getCurrentStack()) {
             $stack->addProfile($profile);
         }
 
--- a/Collector/Collector.php
+++ b/Collector/Collector.php
@@ -62,7 +62,7 @@ class Collector extends DataCollector
     }
 
     /**
-     * @return Stack
+     * @return Stack|boolean false if no current stack.
      */
     public function getCurrentStack()
     {
--- a/Collector/Collector.php
+++ b/Collector/Collector.php
@@ -62,7 +62,7 @@ class Collector extends DataCollector
     }
 
     /**
-     * @return Stack|boolean false if no current stack.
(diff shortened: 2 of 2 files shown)


Collector/ProfilePlugin.php
```diff
@@ -58,7 +58,9 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
     {
         $profile = new Profile($this->pluginName, $this->formatter->formatRequest($request));
 
-        $this->collector->getCurrentStack()->addProfile($profile);
+        if (false !== $stack = $this->collector->getCurrentStack()) {
```
jeromegamez: In addition to @dbu's https://github.com/php-http/HttplugBundle/pull/134#pullrequestreview-23207240: as `getCurrentStack()` either returns a `Stack` or not, would 

```php
if ($stack = $this->collector->getCurrentStack()) {
```

perhaps suffice?

Nyholm:
> is it "legal" that we have no stack? or should that never happen?

To be honest, I do not know. Im not sure how we can be in ProfilerPlugin without a stack. @fbourigault do you know?

Collector/ProfilePlugin.php
Nyholm: Thank you for the reivew. You are correct. I should update. 

We are trying out https://app.assignees.io/. A service that automatically finds the best reviewer. Somehow it selected you.

fbourigault:
It should never happen as the `StackPlugin` is inserted to be the very first plugin (and it's not a decorated plugin).

Collector/ProfilePlugin.php
jeromegamez: Because it somehow knows that I'm a nitpicker and micromanager 😂🙈. Scary.

dbu: i would write it like jerome proposes, thats simpler to read.

the question remains if this situation can ever happen. maybe we should instead have getCurrentStack throw an exception if there is no current stack. that is more disruptive but would force to figure out why it happens.

though maybe as this is just a debugging tool, we should rather silently fail than make a noise about it.

fbourigault: Silently fail will just hide details in the profiler, but it may be hard for users to find why my plugin is not in the profiler.

Throwing a LogicException may do the job unless I missed some cases while refactoring the profiler.

Nyholm:
What if you: 

```php
$plugin = $this->container->get('httplug.plugin.logger');
$client = new PluginClient($httpClient, [$plugin]);
```

But that's not how Daniel did it...

fbourigault:
This is not a problem as the plugin doesn't get decorated.

Nyholm:
I'll merge this and tag a patch release. Having a request missing from the profile is better than causing errors

dbu:
thanks. agree that we better have a non-ideal workaround than nothing at
all.

PR #135, opened 2017-03-17
Extend profiler toolbar item

chr-hertel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

**Depends on [stopwatch plugin PR](https://github.com/php-http/stopwatch-plugin/pull/3)**

#### What's in this PR?

Extended profiler toolbar item to display more details and duration

bildschirmfoto 2017-03-17 um 17 47 31

### DIFF
--- a/Collector/Collector.php
+++ b/Collector/Collector.php
@@ -108,4 +108,14 @@ class Collector extends DataCollector
             return $stack->getClient() == $client;
         });
     }
+
+    /**
+     * @return int
+     */
+    public function getDurationSum()
+    {
+        return array_reduce($this->data['stacks'], function ($carry, Stack $stack) {
+            return $carry + $stack->getDuration();
+        }, 0);
+    }
 }
--- a/Collector/Collector.php
+++ b/Collector/Collector.php
@@ -112,7 +112,7 @@ class Collector extends DataCollector
     /**
      * @return int
      */
-    public function getDurationSum()
+    public function getTotalDuration()
     {
         return array_reduce($this->data['stacks'], function ($carry, Stack $stack) {
             return $carry + $stack->getDuration();
--- a/Collector/ProfileClient.php
+++ b/Collector/ProfileClient.php
@@ -42,6 +42,11 @@ class ProfileClient implements HttpClient, HttpAsyncClient
     private $stopwatch;
 
     /**
+     * @var array
+     */
+    private $eventNames = [];
+
+    /**
      * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement both HttpClient and
      *                                              HttpAsyncClient interfaces.
      * @param Collector                  $collector
@@ -81,6 +86,7 @@ class ProfileClient implements HttpClient, HttpAsyncClient
 
                 return $response;
             }, function (\Exception $exception) use ($event, $stack) {
+                $event->stop();
                 $this->collectExceptionInformations($exception, $event, $stack);
 
                 throw $exception;
@@ -173,6 +179,14 @@ class ProfileClient implements HttpClient, HttpAsyncClient
      */
     private function getStopwatchEventName(RequestInterface $request)
     {
-        return sprintf('%s %s', $request->getMethod(), $request->getUri()->__toString());
+        $name = sprintf('%s %s', $request->getMethod(), $request->getUri());
+
+        if (isset($this->eventNames[$name])) {
+            $name .= sprintf(' [#%d]', ++$this->eventNames[$name]);
+        } else {
+            $this->eventNames[$name] = 1;
+        }
+
+        return $name;
     }
 }
--- a/Collector/ProfileClient.php
+++ b/Collector/ProfileClient.php
@@ -8,6 +8,8 @@ use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\Stopwatch\Stopwatch;
+use Symfony\Component\Stopwatch\StopwatchEvent;
 
 /**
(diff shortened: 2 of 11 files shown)


Collector/StackPlugin.php
```diff
@@ -49,12 +49,22 @@ public function __construct(Collector $collector, Formatter $formatter, $client)
...
+            $this->formatter->formatRequest($request)
+        );
 
         $this->collector->addStack($stack);
 
         return $next($request)->then(function (ResponseInterface $response) use ($stack) {
             $stack->setResponse($this->formatter->formatResponse($response));
+            $stack->setResponseCode($response->getStatusCode());
+
+            if ($response->hasHeader('X-Duration')) {
```
dbu: where does the x-duration header come from?

dbu: ah, now i got it, this depends on https://github.com/php-http/stopwatch-plugin/pull/3

dbu:
@Nyholm it depends on https://github.com/php-http/stopwatch-plugin/pull/3 and we currently are trying to figure out how to track requests and responses in the plugin chain...

fbourigault:
What do you thing about removing the stopwatch plugin from the profiled stacks and directly use the stopwatch component in the `StackPlugin`?

```php
    /**
     * {@inheritdoc}
     */
    {
        $stack = new Stack(
            $this->client,
            $request->getMethod(),
            $request->getRequestTarget(),
            $this->formatter->formatRequest($request)
        );

        $this->collector->addStack($stack);
        $eventName = sprintf('%s %s', $request->getMethod(), $request->getUri()->__toString());
        $event = $this->stopwatch->start($eventName, 'php_http.request');
        return $next($request)->then(function (ResponseInterface $response) use ($stack, $event) {
            $stack->setResponse($this->formatter->formatResponse($response));
            $stack->setResponseCode($response->getStatusCode());
            
            $event->stop();
            $stack->setDuration($event->getDuration());

            return $response;
        }, function (Exception $exception) use ($stack) {
            $stack->setResponse($this->formatter->formatException($exception));
            $stack->setFailed(true);

            $event->stop();
            $stack->setDuration($event->getDuration());

            throw $exception;
        });
    }
```

Such implementation does not require changes neither in symfony nor in the stopwatch plugin :)

chr-hertel:
@fbourigault yes, would be an easy solution and more convincing than extending symfony's stopwatch. thanks :)
but IMO it's still a valid option to manipulate the response in stopwatch-plugin. we're just adding plugin related meta information to response.
@dbu @Nyholm let me know what your favorite choice is and i will adjust those PRs.

dbu:
i think using the stopwatch directly instead of the plugin sounds better to me. much less complicated and achieves the same functionality in a clean way. at that point we would not need the stopwatch plugin for the bundle at all anymore, right?

fbourigault:
The dependency will still be required. See https://github.com/php-http/HttplugBundle/blob/master/DependencyInjection/HttplugExtension.php#L185-L187.

dbu:
oh good. then we don't need to decide if it would be a BC break to change the dependency...

chr-hertel:
alright, great! :) i'll take care of this the next days

Nyholm:
I've added this to the 1.5 milestone. Can you see why the tests are failing?

fbourigault:
Could you also rebase your branch to get thinks working with `ProfileClient`?

chr-hertel:
rebase is done and tests are not failing anymore. (honestly no clue what happened there.)

any ideas how to deal with the stopwatch-plugin? i think we need to disabled or even remove it because it interferes with the current implementation. but we could also stop using `debug.stopwatch` service in `ProfileClient` and instantiate a separate stopwatch.

fbourigault:
About the stopwatch, I would remove the plugin and mesure time in the StackPlugin.

Collector/ProfileClient.php
```diff
@@ -63,16 +72,20 @@ public function sendAsyncRequest(RequestInterface $request)
...
+            function (ResponseInterface $response) use ($event, $stack) {
+                $event->stop();
+                $this->collectResponseInformations($response, $event, $stack);
 
-            return $response;
-        }, function (\Exception $exception) use ($stack) {
-            $this->collectExceptionInformations($exception, $stack);
+                return $response;
+            }, function (\Exception $exception) use ($event, $stack) {
+                $this->collectExceptionInformations($exception, $event, $stack);
```
fbourigault: Should we call `$event->stop()` here?

Collector/ProfileClient.php
```diff
@@ -115,32 +131,48 @@ private function collectRequestInformations(RequestInterface $request, Stack $st
...
+    /**
+     * Generates the event name.
+     *
+     * @param RequestInterface $request
+     *
+     * @return string
+     */
+    private function getStopwatchEventName(RequestInterface $request)
+    {
+        return sprintf('%s %s', $request->getMethod(), $request->getUri()->__toString());
```
fbourigault: The `-->toString()` is useless as `%s` will cast the URI as string.

chr-hertel:
@fbourigault Thanks for your review! i wasn't able to reproduce your issues exactly but i had quite similar effects caused by duplicate stopwatch event names, so i improved `ProfileClient::getStopwatchEventName`.
can you please test, if that fixes your problem as well?

Collector/Collector.php
```diff
@@ -108,4 +108,14 @@ public function getClientStacks($client)
             return $stack->getClient() == $client;
         });
     }
+
+    /**
+     * @return int
+     */
+    public function getDurationSum()
```
dbu: could we call this getTotalDuration?

fbourigault:
Thank you @chr-hertel.

Issue #108, opened 2016-08-02
Use cUrl formatter


Nyholm:
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |
| Version |  |

We should implement the curl formatter on the profiler page.

Issue #93, opened 2016-07-18
Update design on WebProfiler


Nyholm:
| Q | A |
| --- | --- |
| Bug? | yes |
| New Feature? | yes |
| Version | n/a |

Our WebProfiler has just been updated with much more information and new design. But there is room for improvements. If we look at @florianpreusner's [Guzzle Bundle](https://github.com/8p/GuzzleBundle) we see some nice screenshots of how it could look like.

sagikazarmark:
Doesn't it use the builtin VarDumper? To be honest I am not sure if representing eg. headers as a php data structure is the best we can do.

fbourigault:
@csarrazi's [Guzzle Bundle](https://github.com/csarrazi/CsaGuzzleBundle) show us nice screenshots too.

I will start updating the design by introducing Symfony 2.8+ badges count in the toolbar and the menu.

As the 2.7 and 2.8 are really different, what do you think about doing first a small refactoring to avoid writing everywhere `{% if constant('Symfony\\Component\\HttpKernel\\Kernel::VERSION_ID') >= 20800 %}`! I propose to create 3 templates files:
- `webprofiler.html.twig` which would be the 2.8+ profiler design.
- `webprofiler_legacy.html.twig` which would be the 2.7 profiler design.
- `webprofiler_common.html.twig` which would be the shared design.
With those 3 templates and the extensive use of blocks, updating the design would be easier.

The right design would be picked by the DI Extension which would set the `template` tag attribute to the right one.

WDYT?

dbu:
another option would be to bump the minimal symfony version to 2.8. that is no big jump from 2.7 and will simplify the work here... any objections to that?

fbourigault:
Would be even better! But does this require a new major release?

dbu:
imho no. composer.json will make sure it does not get accidentally installed. it should be a new minor version though, so that we can still release bugfixes for 2.7 if somebody really needs one. 1.4 is not yet released, so that would be fine.

fbourigault:
If everybody agrees, I can do the cleanup and bump the requirement to 2.8+.

Nyholm:
I agree. Bump it to 2.8 =)

fbourigault:
For the next improvement, what do you think to have one small line per request and having to click to expand this request?

It would be something like https://github.com/8p/GuzzleBundle and in later improvements, we will be able to push badges on the right like a curl formater button, stopwatch informations, etc..

[...]

PR #136, opened 2017-04-01
add profile client

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Partially #108, Partially #93, Partially #135 
| Documentation   | N/A
| License         | MIT

#### What's in this PR?

This add tracking of request target url and method. I choose to decorate the Client at the lowest possible level (i.e not the PluginClient, but the inner one) to be able to catch what goes in and out the Client and avoid false informations because of the registered plugins.

Technically, when the profiling is enabled, each client factory is decorated with a `ProfileClientFactory` which returns the original client decorated with a `ProfileClient`.

This is what I suggested for #135 

#### Why?

Those new collected informations are required to write a better profiler.

#### To Do

- [x] Unit tests
- [x] Update changelog

### DIFF
--- a/Collector/ProfileClient.php
+++ b/Collector/ProfileClient.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Http\HttplugBundle\Collector;
+
+use Http\Client\Common\FlexibleHttpClient;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * The ProfileClient decorates any client that implement both HttpClient and HttpAsyncClient interfaces to gather target
+ * url and response status code.
+ *
+ * @author Fabien Bourigault <bourigaultfabien@gmail.com>
+ *
+ * @internal
+ */
+class ProfileClient implements HttpClient, HttpAsyncClient
+{
+    /**
+     * @var HttpClient|HttpAsyncClient
+     */
+    private $client;
+
+    /**
+     * @var Collector
+     */
+    private $collector;
+
+    /**
+     * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement both HttpClient and
+     *                                              HttpAsyncClient interfaces.
+     * @param Collector                  $collector
+     */
+    public function __construct($client, Collector $collector)
+    {
+        if (!($client instanceof HttpClient && $client instanceof HttpAsyncClient)) {
+            throw new \RuntimeException(sprintf(
+                '%s first argument must implement %s and %s. Consider using %s.',
+                    __METHOD__,
+                    HttpClient::class,
+                    HttpAsyncClient::class,
+                    FlexibleHttpClient::class
+            ));
+        }
+        $this->client = $client;
+        $this->collector = $collector;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendAsyncRequest(RequestInterface $request)
+    {
+        $this->collectRequestInformations($request);
+
+        return $this->client->sendAsyncRequest($request);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function sendRequest(RequestInterface $request)
+    {
+        $this->collectRequestInformations($request);
+
+        return $this->client->sendRequest($request);
+    }
+
+    /**
+     * @param RequestInterface $request
+     */
+    private function collectRequestInformations(RequestInterface $request)
+    {
+        if (!$stack = $this->collector->getCurrentStack()) {
+            return;
+        }
+
+        $stack = $this->collector->getCurrentStack();
+        $stack->setRequestTarget($request->getRequestTarget());
+        $stack->setRequestMethod($request->getMethod());
+    }
+}
--- a/Collector/ProfileClientFactory.php
+++ b/Collector/ProfileClientFactory.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Http\HttplugBundle\Collector;
+
+use Http\Client\Common\FlexibleHttpClient;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Http\HttplugBundle\ClientFactory\ClientFactory;
+
+/**
+ * The ProfileClientFactory decorates any ClientFactory and returns the created client decorated by a ProfileClient.
+ *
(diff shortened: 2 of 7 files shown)


Collector/ProfileClient.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    private $collector;
+
+    /**
+     * @param HttpClient|HttpAsyncClient $client
+     * @param Collector                  $collector
+     */
+    public function __construct($client, Collector $collector)
+    {
+        if ($client instanceof HttpAsyncClient && $client instanceof HttpClient) {
```
Nyholm: You can remove all this logic by using the FlexibleHttpClient

Collector/ProfileClientFactory.php
```diff
@@ -0,0 +1,41 @@
...
+    private $collector;
+
+    /**
+     * @param ClientFactory|callable $factory
+     * @param Collector $collector
+     */
+    public function __construct($factory, Collector $collector)
+    {
+        if (!$factory instanceof ClientFactory && !is_callable($factory)) {
+            throw new \RuntimeException(sprintf('Second argument to PluginClientFactory::createPluginClient must be a "%s" or a callable.', ClientFactory::class));
```
Nyholm: First argument...

Collector/ProfileClient.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Http\HttplugBundle\Collector;
+
+use Http\Client\Common\EmulatedHttpAsyncClient;
+use Http\Client\Common\EmulatedHttpClient;
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Psr\Http\Message\RequestInterface;
+
+class ProfileClient implements HttpClient, HttpAsyncClient
```
Nyholm: Maybe a class comment that explains the purpose of this new client.

Collector/ProfileClient.php
fbourigault: Should I use it directly without testing anything or should I use a condition to juste use `FlexibleHttpClient` when only one of both interface is implemented?

fbourigault:
I added the `ProfileClientTest`. Does it worth to test the `ProfileClientFactory` class?

Collector/ProfileClient.php
fbourigault: I moved this logic to the `ProfileClientFactory` and now `FlexibleHttpClient` is used.

fbourigault:
I would do extensive tests about my code but how can I create a client using a callable instead of a factory with the bundle?

Collector/ProfileClientFactory.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createClient(array $config = [])
+    {
+        $client = is_callable($this->factory) ? $this->factory($config) : $this->factory->createClient($config);
+
+        if (!$client instanceof HttpClient || !$client instanceof HttpAsyncClient) {
```
dbu: same remark as above, i think !(instanceof && instanceof) would be easier to read

Tests/Unit/Collector/ProfileClientTest.php
```diff
@@ -0,0 +1,108 @@
...
+    public function testCollectAsyncRequestInformations()
+    {
+        $this->subject->sendAsyncRequest($this->request);
+
+        $this->assertEquals('GET', $this->currentStack->getRequestMethod());
+        $this->assertEquals('/target', $this->currentStack->getRequestTarget());
+    }
+}
+
+interface ClientInterface extends HttpClient, HttpAsyncClient
```
dbu: or just mock flexible client? not sure whats better.

fbourigault: `FlexibleHttpClient` is `final`!

dbu: does that prevent from doing `$this->createMock(FlexiblehttpClient::class)` ?

Nyholm: yes

dbu: argh, thats quite stupid. ok then this is fine.

fbourigault: That's the drawback of the `final` keyword. But this case is special as we have to Mock both `HttpClient` and `HttpAsyncClient` interface.

Should this interface moved in an other package (in the tests section)?

dbu: yeah i think its right to have it here. its not something that anybody should rely on.

fbourigault:
From my POV, this is ready :)

Collector/ProfileClient.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    private $client;
+
+    /**
+     * @var Collector
+     */
+    private $collector;
+
+    /**
+     * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement both HttpClient and
```
greg0ire: How about introducing a new interface extending both interfaces and type hinting against that?

Nyholm: That may be convenient but it does not comply with the Interface Segregation principle. 

I know we talked about this before and decided not to, but I cannot find that thread atm.

Nyholm: Found it on Slack: https://php-http.slack.com/archives/C0D2U9SR3/p1487721978001035

(register here: http://slack.httplug.io/)

greg0ire: The ISP states not to make big interfaces, so that people don't need to implement all of them. In your case, they still can choose to implement either inherited interface, so ISP wouldn't be violated here IMO. And if you do need both methods, it looks like the best bay to express that need.

greg0ire: To be honest, I don't understand what inversion of control has to do with this, but I trust you have you (maybe fairly complicated reasons). Plus, I wasn't aware there were other capabilities, and if an interface is introduced for every combination, this could get quite messy… the real thing to fix here might be php itself , to add the possibility to type hint against several interfaces :P

greg0ire: More about this [here](https://symfony-devs.slack.com/archives/C399ABHQQ/p1491230945392441)

PR #137, opened 2017-04-05
Fix access to private static properties

stof:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a

#### Why?

Using late static binding to access private static properties is wrong, as child classes are not allowed to access them. This would lead to a fatal error when using a child class.


stof:
Note that Scrutinizer knows about this kind of issues and reported it on your project already.

fbourigault:
Thank you @stof.

PR #138, opened 2017-04-05
Fix the validation of incompatible options

stof:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | not really
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

All clients must be validated, not only the first one (which was the case due to returning during the first iteration of the loop over clients).
The easiest way to handle it is to put the validation inside the prototype node, so that each client is validated separately. This also allows to have a better error message (as the error message will tell you which client is invalid as the error will be inside the client configuration node).
The error message is also fixed to talk about the third incompatible features.

I answered `not really` for BC breaks, because there is a small change. If someone configures incompatible features in a client other than the first one, the DI configuration layer was not rejecting it previously.
But this is not a huge issue, because configuring both would not work properly today anyway (which is why features are marked as incompatible), as they would only get access to features of the last one (as other ones would be hidden inside the decorator stack)

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -155,19 +155,16 @@ class Configuration implements ConfigurationInterface
     {
         $root->children()
             ->arrayNode('clients')
-                ->validate()
-                    ->ifTrue(function ($clients) {
-                        foreach ($clients as $name => $config) {
-                            // Make sure we only allow one of these to be true
-                            return (bool) $config['flexible_client'] + (bool) $config['http_methods_client'] + (bool) $config['batch_client'] >= 2;
-                        }
-
-                        return false;
-                    })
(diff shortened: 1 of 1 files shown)


stof:
@Nyholm I totally agree with you about it being a bug fix. I just wanted to be 100% honest about the implication.

Nyholm:
That you. That is appreciated.

dbu:
indeed, thanks @stof 

re BC: i am also much less afraid of a change that might hit people when they update the bundle over a change that might be "hidden" and only show up under certain conditions at runtime. if somebody has borked configuration, they will immediately see when updating the bundle as the container can't be built anymore. and if they do that directly production then they better learn now than later...

PR #144, opened 2017-04-06
default to null as respect_cache_headers is now deprecated

alcohol:


alcohol:
Fixed the test, but now scrutinizer failed :<

Nyholm:
Thank you for this PR

PR #145, opened 2017-04-06
replace deprecated directive with new one

alcohol:


DependencyInjection/Configuration.php
```diff
@@ -370,7 +370,7 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
                         ->addDefaultsIfNotSet()
                         ->children()
                             ->scalarNode('default_ttl')->defaultNull()->end()
-                            ->scalarNode('respect_cache_headers')->defaultTrue()->end()
+                            ->arrayNode('respect_response_cache_directives')->defaultNull()->end()
```
Nyholm: We cannot remove `respect_cache_headers` because that would be a BC break. 
We should have both in the configuration.

alcohol: But having both configured triggers a `\RuntimeException`: https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L69-L74

Nyholm: We should use `scalarNode('respect_cache_headers')->defaultNull()->end()`. That would be the same as "not using it".

alcohol: Pushed an amendment. Looks better now?

alcohol:
Look at the diff now, I feel like the merge didn't go entirely ok. Either that or I need to make some amendments.

Nyholm:
Did I do something wrong? Aren't the diff as you intended now?

alcohol:
`Tests/Unit/DependencyInjection/ConfigurationTest.php` looks wrong. Double entries in the array.

Nyholm:
Yes, and it should be `0` and not `null`, Can you update that? I do not think I can without stealing cred from you.

DependencyInjection/Configuration.php
```diff
@@ -371,6 +372,17 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
                         ->children()
                             ->scalarNode('default_ttl')->defaultValue(0)->end()
                             ->scalarNode('respect_cache_headers')->defaultNull()->end()
+                            ->variableNode('respect_response_cache_directives')
+                                ->validate()
+                                ->always(function ($v) {
+                                    if (is_null($v) || is_array($v)) {
```
stof: if it must always be an array, use prototyped array node rather than a variable node. It will allow better handling of the data.

Rule of thumb: if you start using `variableNode`, rethink what you are doing. Most of the time, you should not need it.

stof:
you need to convert `respect_cache_headers` to the new way and trigger a deprecation warning.
The existing config must keep working for projects using it, but they should be warned that they should migrate their config to the new way.

alcohol:
I'll have more time next week at the earliest unfortunately.

DependencyInjection/Configuration.php
alcohol: Except it can also be null.

alcohol:
@stof what approach do you suggest? Converting it is not entirely straightforward since the formats are different (`bool` vs `null | array`).

stof:
@alcohol what are the new configurations corresponding to `true` and `false` ?

The other option would be to rely on a BC layer at the library level (which should be here anyway) and only add a validation rule forbidding to configure both the old and the new way at the same time. The deprecation warning would still be trigger asking to change the configuration though.

alcohol:
Afaik there is a BC layer in the cache plugin (see [here](https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L361-L367)).

Not sure what would correspond to `true` and `false`. I guess `false` would be `[]` (empty array) and `true` would default to the headers that the plugin uses by default `['no-cache', 'private', 'max-age', 'no-store']`?

Tests/Resources/Fixtures/config/full.xml
```diff
@@ -44,7 +44,10 @@
                 <my_service type="service" service="my_auth_service"/>
             </authentication>
             <cache cache-pool="my_cache_pool" stream-factory="my_other_stream_factory">
-                <config default-ttl="42" respect-cache-headers="false"/>
+                <config default-ttl="42" respect-cache-headers="false">
+                    <respect-response-cache-directives>X-Foo</respect-response-cache-directives>
```
dbu: this is supposed to raise an error: https://github.com/php-http/cache-plugin/blob/b2f15cee4ff46f61f7b90137b41627d17fd57be9/src/CachePlugin.php#L69-L72

we should remove the deprecated configuration option from the test. if you want to be precise, you can add a separate legacy test for it

alcohol:
Ok I am a bit unclear on what it is that I should implement at this point.

- [ ]  trigger a deprecation warning?
- [ ]  automatically convert legacy values to new values?
- [ ] disallow old style configuration (this would break BC)?

.. please give me some direction :-|

stof:
@alcohol if the library has a proper BC layer, here is my suggestion:

- trigger a deprecation warning in the config when using the old style config (just to make it easier for people to spot where they should change something, as they don't configure the library directly, and we must force them to get rid of the old style config before the next major version of the bundle)
- disallow configuring both the old style and new style config at the same time (to avoid having to merge both, which is harder to understand

Then, make sure both the new and old config styles work fine for different use cases (enabling directives, disabling them, keeping default settings)

alcohol:
I'm not very familiar with the Symfony Config Definition stuff. Is it possible to specify that a node can have either A or B as a child, but not both?

fbourigault:
You can look at https://github.com/php-http/HttplugBundle/blob/master/DependencyInjection/Configuration.php#L161-L167 for a similar example.

alcohol:
Eh, I give up. The documentation of this config stuff is quite useless and I just keep getting different errors that provide no clear instructions as to how to fix what is wrong.

dbu:
don't give up please. basically its enough to allow the old and the new configuration and trigger a deprecation warning when the old config is used. all the BC handling is done by the library. if the config can complain when both are set, all the bether. otherwise there is an exception on trying to load the service that is thrown by the plugin because it does not want both settings to be set.

alcohol:
Using
```
                    ->arrayNode('config')
                        ->addDefaultsIfNotSet()
                        ->validate()
                        ->ifTrue(function ($config) {
                            // Cannot set both respect_cache_headers and respect_response_cache_directives
                            return isset($config['respect_cache_headers'], $config['respect_response_cache_directives']);
                        })
                        ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
                        ->end()
                        ->children()
                            ->scalarNode('default_ttl')->defaultValue(0)->end()
                            ->scalarNode('respect_cache_headers')->defaultNull()->end()
                            ->variableNode('respect_response_cache_directives')
                                ->validate()
                                ->always(function ($v) {
                                    if (is_null($v) || is_array($v)) {
                                        return $v;
                                    }
                                    throw new InvalidTypeException();
                                })
                                ->end()
                            ->defaultNull()
                            ->end()
                        ->end()
                    ->end()
```
it adds both `respect_response_cache_directives` and `respect_cache_headers`.

Using
```
                    ->arrayNode('config')
                        ->addDefaultChildrenIfNoneSet(['default_ttl', 'respect_response_cache_directives'])
                        ->validate()
                        ->ifTrue(function ($config) {
                            // Cannot set both respect_cache_headers and respect_response_cache_directives
                            return isset($config['respect_cache_headers'], $config['respect_response_cache_directives']);
                        })
                        ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
                        ->end()
                        ->children()
                            ->scalarNode('default_ttl')->defaultValue(0)->end()
                            ->scalarNode('respect_cache_headers')->defaultNull()->end()
                            ->variableNode('respect_response_cache_directives')
                                ->validate()
                                ->always(function ($v) {
                                    if (is_null($v) || is_array($v)) {
                                        return $v;
                                    }
                                    throw new InvalidTypeException();
                                })
                                ->end()
                            ->defaultNull()
                            ->end()
                        ->end()
                    ->end()
```
it throws a cryptic exception that doesn't tell me anything about how to solve this problem (neither does the documentation):

```
Failed asserting that exception of type "Symfony\Component\Config\Definition\Exception\InvalidDefinitionException" matches expected exception "\Symfony\Component\Config\Definition\Exception\InvalidConfigurationException". Message was: "->addDefaultChildrenIfNoneSet() is not applicable to concrete nodes at path "httplug.clients..plugins..cache.config""
```

stof:
`addDefaultChildrenIfNoneSet` does not make sense on the `config`node. It is only for prototyped nodes. What you need is `addDefaultsIfNotSet`

alcohol:
@stof but that one sets all 3 configured nodes, while I only want to set 2 by default (skip the deprecated one if not provided).

stof:
Well, don't put a default value on the deprecated node if it should not be there by default

alcohol:
Ok I think I have the error scenario covered. Where should I do the legacy conversion and warning trigger?

stof:
@alcohol don't do the conversion (the library does it). And for the warning trigger, do it in a `beforeNormalization()` on the old node (with `->always()` as condition to trigger it, as we always want to trigger it when it appears in your config file)

alcohol:
Hmm the problem I run into now is that if the deprecated `respect_cache_headers` config option is set (someone has it in their old legacy config), the `respect_response_cache_directives` config option gets added because it has a default of `null`. However, if I remove the default then it never sets it, regardless of whether or not the user has a legacy config or not. Is it safe to rely on the cache plugin itself to provide a default value?

stof:
@alcohol I think it is.

alcohol:
Okay I think that covers it, although I am not sure how to test that it actually triggers the deprecation since it gets silenced.

alcohol:
Oh FFS. On Travis it fails cause the deprecation is seen. But locally (running same phpunit version) I get nothing like that.

```
/code/github.com/alcohol/HttplugBundle > vendor/bin/phpunit 
PHPUnit 5.7.19 by Sebastian Bergmann and contributors.

..................................................                50 / 50 (100%)

Time: 469 ms, Memory: 12.00MB

OK (50 tests, 112 assertions)
```

stof:
Make sure you use the same versions of deps too

alcohol:
For some reason I did not have the phpunit-bridge installed. Okay, lets see if I can find documentation on how to specify this deprecation as "expected".

alcohol:
Good lord I wasted far too much time on this today.

alcohol:
Not really, I'm still not clear for example on what the difference is between a concrete node and a not-concrete node :-(. There is zero documentation on this subject.

dbu:
i squashed the commits, fixed the thing that styleci complained about and merged to master.

thanks a lot for your persistence @alcohol 

and i agree with you, the documentation on the configuration definition is lacking. i tend to look up examples in other bundles i know to figure out how to do things. doctrine is quite useful because they have to do some complicated use cases ;-)

alcohol:
Thanks, glad I could still contribute something at least :-)

PR #148, opened 2017-04-13
run tests against php 7.1

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | N/A
| License         | MIT

This add php 7.1 in travis-ci matrix.


xabbuh:
Currently, all three listed Symfony versions will be run against every configured PHP versions. Thus, the build matrix does not only get one but three additional jobs. As the Travis CI quota for the number of jobs run in parallel applies to all repositories in the `php-http` organisation what do you think about reducing the number of jobs here and run tests with different Symfony versions only with one PHP version?

Nyholm:
@xabbuh you are correct. I will make a PR with the suggested changes. 

@fbourigault Thank you!

PR #154, opened 2017-04-19
Display discovered clients in the profiler

### LINKED ISSUE #133: There's not always a Stack on the Collector

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #133 
| Documentation   | N/A
| License         | MIT

## Content

Auto discovered clients were not visible in the profiler. This behavior is broken since at least 1.3. I didn't test any further back to find when it was broken.

This also fixes the root cause of #133. The easier way to test it is with `knplabs/github-api`.

The bug root cause was caused by the stopwatch plugin which was shared between every clients. When a client was configured, the stopwatch plugin gets decorated with the `ProfilePlugin` and when executed, no stack plugin was executed before.

## About plugins state

So we have a plugin instance per client. I don't know if it's a good or a bad thing but we probably have to be clear on what to do. I think its ok to share a plugin if it is stateless, but if the plugin is stateful it should not be shared at all (like the StackPlugin which hold the client name).

I think we should choose a way to go, shared plugins or no shared plugins. Doing the choice will probably allow us to do some refactoring in the bundle container extension, which looks like really complicated to met.

Ping @Nyholm @sagikazarmark @dbu

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -10,6 +10,7 @@ use Http\Discovery\HttpAsyncClientDiscovery;
 use Http\Discovery\HttpClientDiscovery;
 use Http\HttplugBundle\ClientFactory\DummyClient;
 use Http\HttplugBundle\ClientFactory\PluginClientFactory;
+use Http\HttplugBundle\Collector\ProfileClientFactory;
 use Http\HttplugBundle\Collector\ProfilePlugin;
 use Http\Message\Authentication\BasicAuth;
 use Http\Message\Authentication\Bearer;
@@ -272,16 +273,7 @@ class HttplugExtension extends Extension
             } elseif ('authentication' === $pluginName) {
                 $plugins = array_merge($plugins, $this->configureAuthentication($container, $pluginConfig, $serviceId.'.authentication'));
             } else {
-                $pluginServiceId = $serviceId.'.plugin.'.$pluginName;
-
-                $definition = class_exists(ChildDefinition::class)
-                    ? new ChildDefinition('httplug.plugin'.'.'.$pluginName)
-                    : new DefinitionDecorator('httplug.plugin'.'.'.$pluginName)
-                ;
-
-                $this->configurePluginByName($pluginName, $definition, $pluginConfig, $container, $pluginServiceId);
-                $container->setDefinition($pluginServiceId, $definition);
-                $plugins[] = $pluginServiceId;
+                $plugins[] = $this->configurePlugin($container, $serviceId, $pluginName, $pluginConfig);
             }
         }
 
@@ -294,27 +286,12 @@ class HttplugExtension extends Extension
 
             //Decorate each plugin with a ProfilePlugin instance.
             foreach ($plugins as $pluginServiceId) {
-                $container->register($pluginServiceId.'.debug', ProfilePlugin::class)
-                    ->setDecoratedService($pluginServiceId)
-                    ->setArguments([
-                        new Reference($pluginServiceId.'.debug.inner'),
-                        new Reference('httplug.collector.collector'),
-                        new Reference('httplug.collector.formatter'),
-                        $pluginServiceId,
-                    ])
-                    ->setPublic(false)
-                ;
+                $this->decoratePluginWithProfilePlugin($container, $pluginServiceId);
             }
 
-            // Add the newstack plugin
-            $definition = class_exists(ChildDefinition::class)
-                ? new ChildDefinition('httplug.plugin.stack')
-                : new DefinitionDecorator('httplug.plugin.stack')
-            ;
-
-            $definition->addArgument($clientName);
-            $container->setDefinition($serviceId.'.plugin.newstack', $definition);
-            array_unshift($plugins, $serviceId.'.plugin.newstack');
+            // To profile the requests, add a StackPlugin as first plugin in the chain.
+            $stackPluginId = $this->configureStackPlugin($container, $clientName, $serviceId);
+            array_unshift($plugins, $stackPluginId);
         }
 
         $container
@@ -397,7 +374,12 @@ class HttplugExtension extends Extension
                 $httpClient = $this->registerAutoDiscoverableClient(
                     $container,
                     'auto_discovered_client',
-                    [HttpClientDiscovery::class, 'find'],
+                    $this->configureAutoDiscoveryFactory(
+                        $container,
+                        HttpClientDiscovery::class,
+                        'auto_discovered_client',
+                        $config
+                    ),
                     $this->isConfigEnabled($container, $config['profiling'])
                 );
             }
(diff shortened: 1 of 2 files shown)


Nyholm:
I think it is fair to share plugins. The plugins should generally be stateless.. But the StackPlugin is special. Can it be treated in a special way o should we try to make it stateless?

fbourigault:
StackPlugin is already managed in a special way (i.e. not decorated with PluginClient).

I think we have to work on how plugins are decorated. If the plugin is a shared one, we have to register it once and decorate only once. If the plugin is defined in the client, from httplug catalogue (means stopwatch, redirect), we have to create an instance per client and decorate this instance.

What we can do for user created plugins? Shall we consider them like shared plugins and document that sharing a plugin not designed for can be harmful?

In the current definition, I would make all plugins from `plugins.xml` abstract, then iterate over the shared plugin configuration and create required plugins from abstract definition (decorated with ProfileClient if profiling is on).
Then loop over clients definition and create specific definition for httplug shipped plugins (decorated with ProfileClient).
For the user defined plugins, we have to write some logic to ensure that the service will be decorated only once with ProfileClient.

And then we add a StackPlugin as first plugin.

WDYT?

Nyholm:
> StackPlugin is already managed in a special way

Good

> What we can do for user created plugins? Shall we consider them like shared plugins and document that sharing a plugin not designed for can be harmful

Yes. 

>  I would make all plugins from plugins.xml abstract,

That is a BC break, right? Why don't we use the same logic for user defined plugins as we do with ours?

DependencyInjection/HttplugExtension.php
```diff
@@ -467,4 +457,64 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
+     * @param array            $pluginConfig
+     *
+     * @return string
+     */
+    private function configurePlugin(ContainerBuilder $container, $serviceId, $pluginName, array $pluginConfig)
+    {
+        $pluginServiceId = $serviceId.'.plugin.'.$pluginName;
+
+        $definition = class_exists(ChildDefinition::class)
+            ? new ChildDefinition('httplug.plugin'.'.'.$pluginName)
```
Nyholm: There too many concats here. 

```php
ChildDefinition('httplug.plugin.'.$plug...
```

Nyholm: Same on the next line

fbourigault:
> That is a BC break, right? Why don't we use the same logic for user defined plugins as we do with ours?
This is probably a BC break. I will not change that.

I will share again the stopwatch plugin and only focus this PR on fixing #133.

Nyholm:
Im testing this PR. The auto_discovered_client does not show up the the profiler.

fbourigault:
Finally I focused this PR about displaying auto discovered clients in the profiler and fixing #133.

DependencyInjection/HttplugExtension.php
```diff
@@ -467,4 +467,90 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
+
+        $this->configurePluginByName($pluginName, $definition, $pluginConfig, $container, $pluginServiceId);
+        $container->setDefinition($pluginServiceId, $definition);
+
+        return $pluginServiceId;
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param $pluginServiceId
```
dbu: missing typehint

DependencyInjection/HttplugExtension.php
```diff
@@ -467,4 +467,90 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
         return new Configuration($container->getParameter('kernel.debug'));
     }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param string           $serviceId
+     * @param string           $pluginName
+     * @param array            $pluginConfig
+     *
+     * @return string
```
dbu: please add something like "service id for this plugin"

fbourigault:
Thanks for your review. I will add some description to extracted methods.

fbourigault:
I fixed issues pointed out by @dbu. I also added one line description for new methods. Proofreading is welcome!

DependencyInjection/HttplugExtension.php
```diff
@@ -430,33 +417,46 @@ private function configureAutoDiscoveryClients(ContainerBuilder $container, arra
...
-        $pluginClientOptions = [];
-
+        $plugins = [];
         if ($profiling) {
-            // Tell the plugin journal what plugins we used
-            $container
-                ->getDefinition('httplug.collector.plugin_journal')
-                ->addMethodCall('setPlugins', [$name, ['httplug.plugin.stopwatch']])
-            ;
+            // Add the newstack plugin
```
dbu: new stack plugin? or newstack? if i guess correctly what we do here, lets do: `// To profile requests, we put a stack plugin instance as first plugin in the chain, followed by the stopwatch plugin`

DependencyInjection/HttplugExtension.php
```diff
@@ -467,4 +467,98 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
+     *
+     * @param ContainerBuilder $container
+     * @param string           $clientName
+     * @param string           $serviceId
+     *
+     * @return string configured new stack plugin service id
+     */
+    private function configureNewStackPlugin(ContainerBuilder $container, $clientName, $serviceId)
+    {
+        $pluginServiceId = $serviceId.'.plugin.newstack';
```
dbu: i see that this is a refactoring and you just kept the previous name, but why do we call it "newstack"? its kind of confusing to me. its not really more new than other services, just specific to this httplug client, right?

DependencyInjection/HttplugExtension.php
```diff
@@ -467,4 +467,98 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
+            ])
+            ->setPublic(false);
+    }
+
+    /**
+     * Configure a new StackPlugin for a client.
+     *
+     * @param ContainerBuilder $container
+     * @param string           $clientName
+     * @param string           $serviceId
```
dbu: is this the service id of a httplug client? if so, lets say that.

DependencyInjection/HttplugExtension.php
```diff
@@ -467,4 +467,98 @@ public function getConfiguration(array $config, ContainerBuilder $container)
...
+                $pluginServiceId,
+            ])
+            ->setPublic(false);
+    }
+
+    /**
+     * Configure a new StackPlugin for a client.
+     *
+     * @param ContainerBuilder $container
+     * @param string           $clientName
```
dbu: this is used to configure the stack plugin so that we can show the right client name in the profiler, right? maybe mention this here...

fbourigault:
@dbu I improved comments. Hope this is now easier to understand.

PR #155, opened 2017-04-26
expand configuration definition and adjust tests, closes #151

### LINKED ISSUE #151: Allow more cache config
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | dev-master

#### Expected Behavior

The cache plugin has a lot of configuration options right now. See [here](https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L337)

We should let the user to specify those by configuration.

alcohol:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #151
| License         | MIT

I would like to point out that the official documentation at http://docs.php-http.org/en/latest/plugins/cache.html does not mention the `hash_algo` configuration option.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -2,6 +2,13 @@
 
 namespace Http\HttplugBundle\DependencyInjection;
 
+use Http\Client\Common\Plugin\Cache\Generator\CacheKeyGenerator;
+use Http\Client\Common\Plugin\Journal;
+use Http\Message\CookieJar;
+use Http\Message\Formatter;
+use Http\Message\StreamFactory;
+use Psr\Cache\CacheItemPoolInterface;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\NodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
@@ -352,69 +359,33 @@ class Configuration implements ConfigurationInterface
         $children = $pluginNode->children();
 
         $children->append($this->createAuthenticationPluginNode());
+        $children->append($this->createCachePluginNode());
 
-        $children->arrayNode('cache')
-            ->canBeEnabled()
-            ->addDefaultsIfNotSet()
-                ->children()
-                    ->scalarNode('cache_pool')
-                        ->info('This must be a service id to a service implementing Psr\Cache\CacheItemPoolInterface')
-                        ->isRequired()
-                        ->cannotBeEmpty()
-                    ->end()
-                    ->scalarNode('stream_factory')
-                        ->info('This must be a service id to a service implementing Http\Message\StreamFactory')
-                        ->defaultValue('httplug.stream_factory')
-                        ->cannotBeEmpty()
-                    ->end()
-                    ->arrayNode('config')
-                        ->addDefaultsIfNotSet()
-                        ->validate()
-                        ->ifTrue(function ($config) {
-                            // Cannot set both respect_cache_headers and respect_response_cache_directives
-                            return isset($config['respect_cache_headers'], $config['respect_response_cache_directives']);
-                        })
-                        ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
-                        ->end()
-                        ->children()
-                            ->scalarNode('default_ttl')->defaultValue(0)->end()
-                            ->enumNode('respect_cache_headers')
-                                ->values([null, true, false])
-                                ->beforeNormalization()
-                                ->always(function ($v) {
-                                    @trigger_error('The option "respect_cache_headers" is deprecated since version 1.3 and will be removed in 2.0. Use "respect_response_cache_directives" instead.', E_USER_DEPRECATED);
-
-                                    return $v;
-                                })
-                                ->end()
-                            ->end()
-                            ->variableNode('respect_response_cache_directives')
-                                ->validate()
-                                ->always(function ($v) {
-                                    if (is_null($v) || is_array($v)) {
-                                        return $v;
-                                    }
-                                    throw new InvalidTypeException();
-                                })
-                                ->end()
-                            ->end()
-                        ->end()
-                    ->end()
-                ->end()
-            ->end();
-        // End cache plugin
-
-        $children->arrayNode('cookie')
-            ->canBeEnabled()
+        $children
+            ->arrayNode('cookie')
+                ->canBeEnabled()
                 ->children()
                     ->scalarNode('cookie_jar')
-                        ->info('This must be a service id to a service implementing Http\Message\CookieJar')
+                        ->info('This must be a service id to a service implementing '.CookieJar::class)
                         ->isRequired()
                         ->cannotBeEmpty()
                     ->end()
                 ->end()
             ->end();
         // End cookie plugin
+
+        $children
+            ->arrayNode('history')
+                ->canBeEnabled()
+                ->children()
+                    ->scalarNode('journal')
+                        ->info('This must be a service id to a service implementing '.Journal::class)
+                        ->isRequired()
+                        ->cannotBeEmpty()
+                    ->end()
+                ->end()
+            ->end();
+        // End history plugin
 
         $decoder = $children->arrayNode('decoder');
         $disableAll ? $decoder->canBeEnabled() : $decoder->canBeDisabled();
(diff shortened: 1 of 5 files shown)


alcohol:
I opted to drop all default values for the config section and let the plugin handle those instead.

Currently I have not yet figured out how to make it not default `methods` to an empty array though. This is still a work in progress.

DependencyInjection/Configuration.php
```diff
@@ -377,25 +377,53 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
                         ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
                         ->end()
                         ->children()
-                            ->scalarNode('default_ttl')->defaultValue(0)->end()
+                            ->scalarNode('cache_lifetime')
+                                ->info('The minimum time we should store a cache item')
+                            ->end()
+                            ->enumNode('cache_key_generator')
+                                ->info('A class implementing Http\Client\Common\Plugin\Cache\Generator\CacheKeyGenerator to generate a PSR-6 cache key')
+                                ->values([null, 'Http\Client\Common\Plugin\Cache\Generator\CacheKeyGenerator'])
```
dbu: please use CacheKeyGenerator::class instead (we are php 5.5 or later so its supported)

alcohol: No problem.

alcohol:
@stof any idea if I can get the config definition builder to somehow not force `methods` to always present in the final tree? Only option I see at the moment is to replace prototyped array node with a variable node :-\

DependencyInjection/Configuration.php
```diff
@@ -368,34 +370,61 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
                         ->cannotBeEmpty()
                     ->end()
                     ->arrayNode('config')
-                        ->addDefaultsIfNotSet()
```
stof: if you remove this, you MUST check whether the `config` key is defined in the DI extension before using it. Currently, the DI extension relies on the fact that the Configuration class ensures it is always defined, but your change means it is not the case anymore.

DependencyInjection/Configuration.php
```diff
@@ -368,34 +370,61 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
...
                         ->children()
-                            ->scalarNode('default_ttl')->defaultValue(0)->end()
+                            ->enumNode('cache_key_generator')
+                                ->info('A class implementing '.CacheKeyGenerator::class.' to generate a PSR-6 cache key')
+                                ->values([null, CacheKeyGenerator::class])
+                            ->end()
+                            ->scalarNode('cache_lifetime')
+                                ->info('The minimum time we should store a cache item')
+                            ->end()
+                            ->scalarNode('default_ttl')
```
stof: integer node looks even better to me for these

DependencyInjection/Configuration.php
```diff
@@ -368,34 +370,61 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
...
+                            })
+                            ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
                         ->end()
                         ->children()
-                            ->scalarNode('default_ttl')->defaultValue(0)->end()
+                            ->enumNode('cache_key_generator')
+                                ->info('A class implementing '.CacheKeyGenerator::class.' to generate a PSR-6 cache key')
+                                ->values([null, CacheKeyGenerator::class])
+                            ->end()
+                            ->scalarNode('cache_lifetime')
```
stof: why removing the default value ?

Tests/Resources/Fixtures/config/full.xml
```diff
@@ -44,9 +44,11 @@
                 <my_service type="service" service="my_auth_service"/>
             </authentication>
             <cache cache-pool="my_cache_pool" stream-factory="my_other_stream_factory">
-                <config default-ttl="42">
+                <config default-ttl="42" cache-lifetime="2592000" hash-algo="sha1" cache-key-generator="null">
                     <respect-response-cache-directives>X-Foo</respect-response-cache-directives>
                     <respect-response-cache-directives>X-Bar</respect-response-cache-directives>
+                    <methods>GET</methods>
+                    <methods>HEAD</methods>
```
stof: this is wrong. The XML tag should be named `method` and we need to have the call `->fixXmlConfig('method')` on the parent node.

Prototyped nodes work a bit special in XML, due to the way child tags are handled when parsing XML in SimpleXML (if you put a single method with your current code, you will not get an array).

thus, the tag name will be better with a singular

Tests/Resources/Fixtures/config/full.xml
```diff
@@ -44,9 +44,11 @@
                 <my_service type="service" service="my_auth_service"/>
             </authentication>
             <cache cache-pool="my_cache_pool" stream-factory="my_other_stream_factory">
-                <config default-ttl="42">
+                <config default-ttl="42" cache-lifetime="2592000" hash-algo="sha1" cache-key-generator="null">
                     <respect-response-cache-directives>X-Foo</respect-response-cache-directives>
                     <respect-response-cache-directives>X-Bar</respect-response-cache-directives>
```
stof: this prototyped node is also broken in XML btw

stof:
@alcohol prototyped array nodes always have a default value (the empty array, unless you override it).
What you could do is treating an empty array in a special way when using the config in the DI extension.

DependencyInjection/Configuration.php
```diff
@@ -368,34 +370,61 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
...
+                                // Cannot set both respect_cache_headers and respect_response_cache_directives
+                                return isset($config['respect_cache_headers'], $config['respect_response_cache_directives']);
+                            })
+                            ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
                         ->end()
                         ->children()
-                            ->scalarNode('default_ttl')->defaultValue(0)->end()
+                            ->enumNode('cache_key_generator')
+                                ->info('A class implementing '.CacheKeyGenerator::class.' to generate a PSR-6 cache key')
+                                ->values([null, CacheKeyGenerator::class])
```
alcohol: I don't know if it does an instanceof test or literal string comparison. I guess this needs work.

DependencyInjection/Configuration.php
alcohol: See my first response on this PR.

DependencyInjection/Configuration.php
alcohol: It's always been scalar, so I kept it scalar. Not sure why originally integer was not chosen o_O

Tests/Resources/Fixtures/config/full.xml
alcohol: I'll look into this.

DependencyInjection/Configuration.php
alcohol: Can you elaborate what you mean by this? I don't quite follow.

DependencyInjection/Configuration.php
stof: enumNode is just this: an enum. So the value being configured must be one of the provided values. `CacheKeyGenerator::class` is just a string. There is nothing about instanceof in string comparisons

DependencyInjection/Configuration.php
stof: probably because support for Symfony 2.3 was needed in the past, which did not have the integer node yet.

alcohol:
I resorted to just providing the default value instead. I think that is the easiest approach without having to implement too many work-arounds or do all sorts of additional checks.

DependencyInjection/Configuration.php
alcohol: Dropped the validation in favor of letting the plugin itself handle it directly.

DependencyInjection/Configuration.php
```diff
@@ -564,4 +517,102 @@ private function validateAuthenticationType(array $expected, array $actual, $aut
...
+            ->validate()
+                ->ifTrue(function ($config) {
+                    // Cannot set both respect_cache_headers and respect_response_cache_directives
+                    return isset($config['respect_cache_headers'], $config['respect_response_cache_directives']);
+                })
+                ->thenInvalid('You can\'t provide config option "respect_cache_headers" and "respect_response_cache_directives" simultaniously. Use "respect_response_cache_directives" instead.')
+            ->end()
+            ->children()
+                ->scalarNode('cache_key_generator')
+                    ->info('A class implementing '.CacheKeyGenerator::class.' to generate a PSR-6 cache key')
```
dbu: actually the cache plugin phpdoc is confusing on this one. if this option is set, it must contain an instance of CacheKeyGenerator. in symfony config that means it would be a service id.

dbu: https://github.com/php-http/cache-plugin/pull/41

alcohol: So...

> This must be a service id to a service implementing Http\Client\Common\Plugin\Cache\Generator\CacheKeyGenerator

?

dbu:
exactly. if it is set at all - if its empty, the cache plugin creates an
object on its own.

DependencyInjection/Configuration.php
stof: And then it means that the DI extension has to process the config to turn the id into a Reference object.

stof: Currently, the feature is just broken

alcohol:
Anything that still needs changing/modification?

DependencyInjection/Configuration.php
```diff
@@ -564,4 +523,101 @@ private function validateAuthenticationType(array $expected, array $actual, $aut
...
+     *
+     * @return NodeDefinition Definition for the cache node in the plugins list.
+     */
+    private function createCachePluginNode()
+    {
+        $builder = new TreeBuilder();
+
+        $config = $builder->root('config');
+        $config
+            ->fixXmlConfig('method')
```
alcohol: Should this be considered a BC btw?

stof: Well, as there is no XSD file, the existing XML config file will continue to work, for people lucky enough to have a working config before (it was broken when having a single method). And it will allow everyone to have a working config when using XML, for all cases.

Btw, it cannot be considered `a BC`. The sentence should be `considered a BC break`. The important work in this sentence is `break` 😄

PR #157, opened 2017-05-02
Display stack duration in profiler

### LINKED ISSUE #156: Display the stack duration in the profiler stack header
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | N/A

Display the stack duration in the profiler stack header.

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #156 
| Documentation   | N/A
| License         | MIT

This add the stack duration in the profiler.

httplugbundle4

### DIFF
--- a/Resources/public/style/httplug.css
+++ b/Resources/public/style/httplug.css
@@ -80,6 +80,11 @@
     font-size: 12px;
 }
 
+.httplug-duration {
+    min-width: 6ch;
+    text-align:center;
+}
+
 /**
  * HTTP method colors from swagger-ui.
  */
--- a/Resources/views/webprofiler.html.twig
+++ b/Resources/views/webprofiler.html.twig
@@ -94,6 +94,7 @@
                             <span class="httplug-target">{{ stack.requestTarget }}</span>
                         </div>
                         <div>
+                            <span class="label httplug-duration">{{ stack.duration|number_format }} ms</span>
                             {% if stack.responseCode >= 400 and stack.responseCode <= 599 %}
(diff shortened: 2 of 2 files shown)


Resources/public/style/httplug.css
```diff
@@ -80,6 +80,11 @@
     font-size: 12px;
 }
 
+.httplug-duration {
+    min-width: 6ch;
```
Nyholm: "ch"? What is that? 

Someone once told me to always use "rem".

fbourigault: 1 ch is the width of a "0" in the current font. It looks easier to me as I want the box at least large as if it was filled with 6 characters.

Nyholm: hm. Interesting. Thanks. 

What if it takes more then one sec? "1000 ms" That is 7 chars, right? Will we get a line break? If so, change this to `7ch` and we will be all happy.

fbourigault: I use `number_format` twig filter, so you get a thousand separator. But anyway, I set this as a `min-width` to get the same width when duration is lower than 100ms and lower than 1000ms. When duration is over 1000ms, the box get larger.

Nyholm: Sorry. It was e reading this to fast. It is all good then =)

Nyholm:
Thank you.

PR #159, opened 2017-05-04
Display horizontal scrollbar when necessary for profiler messages

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | N/A
| License         | MIT

Since #146, messages are truncated which may definitively hide informations to user. I replaced the `text-overflow: ellipsis` with a `overflow-x:auto` to display an horizontal scrollbar when required.
httplugbundle6

### DIFF
--- a/Resources/public/style/httplug.css
+++ b/Resources/public/style/httplug.css
@@ -25,8 +25,7 @@
     padding: 5px;
     flex: 1;
     margin: 5px;
-    overflow: hidden;
-    text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


Nyholm:
Do I understand it correct that the URL in the header is truncated but the content is using horizontal scroll?

fbourigault:
Yes.

PR #164, opened 2017-05-11
Make sure $stack is null or Stack. Never boolean.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes, but they are not released yet
| Deprecations?   | no
| Related tickets | fixes issues from scrutinizer
| Documentation   | 
| License         | MIT

This is just a minor change in the return value of `getCurrentStack`

### DIFF
--- a/Collector/ProfilePlugin.php
+++ b/Collector/ProfilePlugin.php
@@ -58,7 +58,8 @@ class ProfilePlugin implements Plugin
     {
         $profile = new Profile($this->pluginName, $this->formatter->formatRequest($request));
 
-        if (null !== $stack = $this->collector->getCurrentStack()) {
+        $stack = $this->collector->getCurrentStack();
+        if (null !== $stack) {
             $stack->addProfile($profile);
         }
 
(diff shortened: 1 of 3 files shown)


fbourigault:
Since #133 is really fixed, maybe we could go back to code there was before #134?

Nyholm:
Are we sure we do always have a stack? Can we remove all these checks if $stack === null?

fbourigault:
Now, the only case I see to not having the stack is a user which use by himself the ProfilePlugin decorator without the StackPlugin.

Collector/ProfilePlugin.php
```diff
@@ -58,7 +58,7 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
     {
         $profile = new Profile($this->pluginName, $this->formatter->formatRequest($request));
 
-        if ($stack = $this->collector->getCurrentStack()) {
+        if (null !== $stack = $this->collector->getCurrentStack()) {
```
fbourigault: Maybe this line will be easier to read if the assignation is moved out of the condition.

Nyholm: You are correct, It is me that really like this syntax

Nyholm:
Thank you for the review.

PR #166, opened 2017-05-11
Allow respect_cache_headers to be "false"

Nyholm:
This PR adds a test for an issue I found. 

We define `respect_cache_headers` like: 

```php
->enumNode('respect_cache_headers')
    ->info('Whether we should care about cache headers or not [DEPRECATED]')
    ->values([null, true, false])
    ->beforeNormalization()
        ->always(function ($v) {
            @trigger_error('The option "respect_cache_headers" is deprecated since version 1.3 and will be removed in 2.0. Use "respect_response_cache_directives" instead.', E_USER_DEPRECATED);

            return $v;
        })
    ->end()
->end()
```

On Symfony 2.8, the EnumNode runs `array_unique` over the possible values. Though, `null` and `false` is treated the same. See https://3v4l.org/KWuNP

This will result in that the possible values for `respect_cache_headers` will only be `true` and `null`. This issue will break BC since we promise that all previous configuration will work on newer versions. 

FYI: In the last release we defined the `respect_cache_headers` like: 

```php
->scalarNode('respect_cache_headers')->defaultTrue()->end()
```

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -573,15 +573,20 @@ class Configuration implements ConfigurationInterface
                         ->end()
                     ->end()
                 ->end()
-                ->enumNode('respect_cache_headers')
+                ->scalarNode('respect_cache_headers')
                     ->info('Whether we should care about cache headers or not [DEPRECATED]')
-                    ->values([null, true, false])
                     ->beforeNormalization()
                         ->always(function ($v) {
                             @trigger_error('The option "respect_cache_headers" is deprecated since version 1.3 and will be removed in 2.0. Use "respect_response_cache_directives" instead.', E_USER_DEPRECATED);
 
                             return $v;
                         })
+                    ->end()
+                    ->validate()
+                        ->ifTrue(function ($v) {
+                            return !in_array($v, [null, true, false]);
+                        })
+                        ->thenInvalid('Invalid value for respect_cache_headers: %s. Must be "null", "true" or "false"')
                     ->end()
                 ->end()
                 ->variableNode('respect_response_cache_directives')
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -584,7 +584,7 @@ class Configuration implements ConfigurationInterface
                     ->end()
                     ->validate()
                         ->ifNotInArray([null, true, false])
-                            ->thenInvalid('Value for "respect_cache_headers" must be null or boolean')
+                        ->thenInvalid('Value for "respect_cache_headers" must be null or boolean')
                     ->end()
                 ->end()
                 ->variableNode('respect_response_cache_directives')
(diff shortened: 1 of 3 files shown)


Nyholm:
@stof, have you seen this issue before? Do you know a workaround?

stof:
hmm, I think the enumNode was always used with strings until now, which is why such issue was never spotted before

Nyholm:
Thank you. I use a ScalarNode for this now.

DependencyInjection/Configuration.php
```diff
@@ -573,15 +573,18 @@ private function createCachePluginNode()
...
-                    ->values([null, true, false])
                     ->beforeNormalization()
                         ->always(function ($v) {
                             @trigger_error('The option "respect_cache_headers" is deprecated since version 1.3 and will be removed in 2.0. Use "respect_response_cache_directives" instead.', E_USER_DEPRECATED);
 
                             return $v;
                         })
+                        ->end()
+                    ->validate()
+                    ->ifNotInArray([null, true, false])
```
stof: this one should be indented inside `->validate`

DependencyInjection/Configuration.php
```diff
@@ -573,15 +573,18 @@ private function createCachePluginNode()
...
+                ->scalarNode('respect_cache_headers')
                     ->info('Whether we should care about cache headers or not [DEPRECATED]')
-                    ->values([null, true, false])
                     ->beforeNormalization()
                         ->always(function ($v) {
                             @trigger_error('The option "respect_cache_headers" is deprecated since version 1.3 and will be removed in 2.0. Use "respect_response_cache_directives" instead.', E_USER_DEPRECATED);
 
                             return $v;
                         })
+                        ->end()
```
stof: the end call should be unindented to make the tree structure visual. This is the end of `->beforeNormalization()`

DependencyInjection/Configuration.php
Nyholm: Thank you

DependencyInjection/Configuration.php
```diff
@@ -581,10 +581,10 @@ private function createCachePluginNode()
 
                             return $v;
                         })
-                        ->end()
+                    ->end()
                     ->validate()
-                    ->ifNotInArray([null, true, false])
-                        ->thenInvalid('Value for "respect_cache_headers" must be null or boolean')
+                        ->ifNotInArray([null, true, false])
+                            ->thenInvalid('Value for "respect_cache_headers" must be null or boolean')
```
stof: `ifNotInArray` does not create a new node in the tree (otherwise it would need an end node too). It configures the validation node. So this one should be unindented (it was right before your last update)

Nyholm: Sorry, it was me that misunderstood. Thank you.

Nyholm:
Thank you for the reviews.

PR #169, opened 2017-05-15
Collect the request after the plugin execution

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

As I explained on Slack, before this fix, the request was collected before the decorated plugin execution and so, changes to the request made by the plugin were displayed in the next plugin in the chain.

Now the request is collected after the plugin execution and so, the plugin changes are displayed within the right plugin block.

Before:
before

After:
after

### DIFF
--- a/Collector/Profile.php
+++ b/Collector/Profile.php
@@ -33,12 +33,10 @@ final class Profile
 
     /**
      * @param string $plugin
-     * @param string $request
      */
-    public function __construct($plugin, $request)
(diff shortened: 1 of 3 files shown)


fbourigault:
Thanks for the review.

PR #170, opened 2017-05-21
Allow you to skip the factory keyword.

### LINKED ISSUE #168: Make `clients.acme.factory` optional and use discovery instead
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | N/A

I propose to make the `factory` key of the client configuration optional (`clients.acme.factory`).
In such case, we will use the discovery to find an available implementation. Of course, the `config` node would be disabled when no factory is set.

Nyholm:
This will fix #168

### DIFF
--- a/ClientFactory/AutoDiscoveryFactory.php
+++ b/ClientFactory/AutoDiscoveryFactory.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Http\HttplugBundle\ClientFactory;
+
+use Http\Discovery\HttpClientDiscovery;
+
+/**
+ * Use auto discovery to find a HTTP client.
+ *
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class AutoDiscoveryFactory implements ClientFactory
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function createClient(array $config = [])
+    {
+        return HttpClientDiscovery::find();
+    }
+}
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -177,7 +177,7 @@ class Configuration implements ConfigurationInterface
                     ->ifTrue(function ($config) {
                         return $config['factory'] === 'httplug.factory.auto' && !empty($config['config']);
                     })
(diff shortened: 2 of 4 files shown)


DependencyInjection/Configuration.php
```diff
@@ -173,9 +173,15 @@ private function configureClients(ArrayNodeDefinition $root)
                     })
                     ->thenInvalid('A http client can\'t be decorated with several of FlexibleHttpClient, HttpMethodsClient and BatchClient. Only one of the following options can be true. ("flexible_client", "http_methods_client", "batch_client")')
                 ->end()
+                ->validate()
+                    ->ifTrue(function ($config) {
+                        return $config['factory'] === 'httplug.factory.auto' && !empty($config['config']);
+                    })
+                    ->thenInvalid('You must specify a valid "factory" in order to use the "config"')
```
xabbuh: IMO this message is not really clear. What does "valid config" mean in this context?

ClientFactory/AutoDiscoveryFactory.php
```diff
@@ -0,0 +1,21 @@
...
+ * @author Tobias Nyholm <tobias.nyholm@gmail.com>
+ */
+class AutoDiscoveryFactory implements ClientFactory
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function createClient(array $config = [])
+    {
+        return HttpClientDiscovery::find();
```
fbourigault: Does it worth to fallback to `HttpAsyncClientDiscovery` if `HttpClientDiscovery` does not returned any implementation?

DependencyInjection/Configuration.php
Nyholm: Thank you. Is it better now?

ClientFactory/AutoDiscoveryFactory.php
Nyholm: I doubt it. I believe it is very rare that an async adapter/client does not implement HttpClient interface (not async).

DependencyInjection/Configuration.php
xabbuh: Looks better

Nyholm:
Thank you for the reviews.

PR #174, opened 2017-05-29
enable autowiring for every discoverable service

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | TODO
| License         | MIT

This continues the work done in #142. By declaring an alias for every service.

This also remove the `DiscoveryPass` in favor of plain xml configuration as I think the provided configuration is equivalent to the compiler pass but easier to maintain.


Resources/config/services.xml
```diff
@@ -8,8 +8,31 @@
...
+
+        <service id="httplug.stream_factory.default" class="Http\Message\StreamFactory">
+            <factory class="Http\Discovery\StreamFactoryDiscovery" method="find" />
+        </service>
+        <service id="Http\Message\StreamFactory" alias="httplug.stream_factory.default" public="false" />
+
+        <service id="httplug.uri_factory.default" class="Http\Message\UriFactory">
+            <factory class="Http\Discovery\UriFactoryDiscovery" method="find" />
+        </service>
+        <service id="Http\Message\UriFactory" alias="httplug.uri_factory.default" public="false" />
```
dbu: is the symfony 4 way of doing this to still have the old-style service name plus the name that is the interface name? that is, should this be public and we deprecate the old-style service and remove it in the next major version?

fbourigault: I'm 200% with this idea, but I think we have a lot of work around all those services. The main issue is probably the ability to override any service with the dependency injection container. I don't know if BC will be kept if we invert the real definition and the alias as a user may replace the old-style name by a definition (instead of an alias) and so break the equity between new-style = old-style. But can this be considered as a BC break?

dbu: if we want to remove httplug.uri_factory.default in the next major version we should make the alias public instead of private so that users can be forward compatible. but i am not sure what the intention of symfony itself is on that question... so lets merge this as is and try to find out what to do for the future.

fbourigault:
Before merging, I will map FQCN aliases to `main_alias` when relevant. This will avoid to break #142 changes.

PR #175, opened 2017-06-19
Profiling might not work with CachePlugin

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | not yet
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| Documentation   | N/A
| License         | MIT

This a a failing functional test about profiling when the `CachePlugin` is involved.

When the cache is hit and the `CachePlugin` does not perform any cache validation, the `ProfileClient` is never reached. This cause the profiled stack about not being complete and will display empty boxes in the profiler panel.

### DIFF
--- a/Collector/ProfilePlugin.php
+++ b/Collector/ProfilePlugin.php
@@ -65,14 +65,14 @@ class ProfilePlugin implements Plugin
 
         // wrap the next callback to profile the plugin request changes
         $wrappedNext = function (RequestInterface $request) use ($next, $profile) {
-            $profile->setRequest($this->formatter->formatRequest($request));
+            $this->onOutgoingRequest($request, $profile);
 
             return $next($request);
         };
 
         // wrap the first callback to profile the plugin request changes
         $wrappedFirst = function (RequestInterface $request) use ($first, $profile) {
-            $profile->setRequest($this->formatter->formatRequest($request));
+            $this->onOutgoingRequest($request, $profile);
 
             return $first($request);
         };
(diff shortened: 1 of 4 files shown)


fbourigault:
As a solution, we could move the ProfileClient code into the ProfilePlugin and let the ProfilePlugin fill all those data if not yet set. By doing so, the last executed plugin would fill the request data.

fbourigault:
I also added a failing test case when the plugin throw an Exception.

fbourigault:
I think the exception case can be handled with a try/catch block in the `StackPlugin`. The cache case could be handled by checking in every `ProfilePlugin` if `ProfileClient` job is done and do it if it's not the case. I will try those solutions as soon as possible.

Nyholm:
👍

fbourigault:
Tests are now green :tada:.

I think I should work `ProfilePlugin` readability as it gets complicated with all those closures!

Also I'm open to any suggestion about not duplicating code between `ProfileClient` and `ProfilePlugin` (`collectRequestInformation`)

composer.json
```diff
@@ -37,11 +37,13 @@
         "php-http/guzzle6-adapter": "^1.1.1",
         "php-http/react-adapter": "^0.2.1",
         "php-http/buzz-adapter": "^0.3",
+        "php-http/mock-client": "^1.0",
         "symfony/phpunit-bridge": "^3.2",
         "symfony/twig-bundle": "^2.8 || ^3.0",
         "symfony/twig-bridge": "^2.8 || ^3.0",
         "symfony/web-profiler-bundle": "^2.8 || ^3.0",
         "symfony/finder": "^2.7 || ^3.0",
+        "symfony/cache": "^3.1",
```
Nyholm: Oh, I see. php-cache is not good enough for you? :)

Nyholm: Jokes aside, can you require this on sf2.8?

fbourigault: Yes if you use PHP 5.5.9+

fbourigault:
> But I do not understand the changes in ProfilePlugin to 100%

Tell me I will try to explain :)

Collector/ProfilePlugin.php
```diff
@@ -77,15 +77,60 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
             return $first($request);
         };
 
-        return $this->plugin->handleRequest($request, $wrappedNext, $wrappedFirst)->then(function (ResponseInterface $response) use ($profile) {
+        try {
+            $promise = $this->plugin->handleRequest($request, $wrappedNext, $wrappedFirst);
+        } catch (Exception $e) {
```
Nyholm: From where is this exception thrown?

Collector/ProfilePlugin.php
```diff
@@ -77,15 +77,60 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
             $profile->setResponse($this->formatter->formatException($exception));
+            $this->collectRequestInformation($request, $stack);
 
             throw $exception;
         });
     }
+
+    /**
+     * Collect request information when not already done by the HTTP client. This happens when using the CachePlugin
+     * and the cache is hit without re-validation.
```
Nyholm: What do you mean with "without re-validation"?

Collector/ProfilePlugin.php
```diff
@@ -77,15 +77,60 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+     *
+     * @param RequestInterface $request
+     * @param Stack|null       $stack
+     */
+    private function collectRequestInformation(RequestInterface $request, Stack $stack = null)
```
Nyholm: So running this does only have an impact when cache plugin and "without re-validation"?

Collector/ProfilePlugin.php
fbourigault: It's thrown from any plugin or even the client.

Collector/ProfilePlugin.php
fbourigault: I mean when the plugin does not add `If-Modified-Since` or `If-None-Match` header.

Collector/ProfilePlugin.php
fbourigault: If the request goes through the whole chain, request information will be already collected by the ProfileClient.

Collector/ProfilePlugin.php
Nyholm: Thanks

Collector/ProfilePlugin.php
```diff
@@ -77,15 +77,60 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
...
+        }
+
+        return $promise->then(function (ResponseInterface $response) use ($profile, $request, $stack) {
             $profile->setResponse($this->formatter->formatResponse($response));
+            $this->collectRequestInformation($request, $stack);
 
             return $response;
-        }, function (Exception $exception) use ($profile) {
+        }, function (Exception $exception) use ($profile, $request, $stack) {
             $profile->setFailed(true);
```
Nyholm: From where is this exception thrown?

How does it differ from https://github.com/php-http/HttplugBundle/pull/175/files#r124504876

fbourigault: This exception can be thrown from anywhere too.

But, as some `Http\Promise\Promise` implementation does not catch all exceptions, we have to use both to catch exceptions in any case. See [`Http\Promise\FulfilledPromise`](https://github.com/php-http/promise/blob/master/src/FulfilledPromise.php) and [`Http\Client\Promise\HttpfulfilledPromise`](https://github.com/php-http/httplug/blob/master/src/Promise/HttpFulfilledPromise.php) for implementation differences.

Nyholm: Okey. I see. Thanks

fbourigault:
Thank you for reviewing this PR.

PR #176, opened 2017-06-26
Stop mocking immutable/value objects

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #175 
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

This is a cleanup of tests to stop mocking immutable/value objects.

`Psr\Http\Message\MessageInterface` and `Psr\Http\Message\ResponseInterface`
mocks are replaced with `GuzzleHttp\Psr7\Request` and
`GuzzleHttp\Psr7\Response` respectively.

`Http\Promise\Promise` mocks are replaced either with
`Http\Client\Promise\HttpFulfilledPromise` or
`Http\Client\Promise\HttpRejectedPromise` depending on the test
requirements.

#### Why?

This make tests easier to read, thus  to maintain, etc... Also this will help me to solve #175 as it will require some profiling rewrite.

### DIFF
--- a/Tests/Unit/Collector/FormatterTest.php
+++ b/Tests/Unit/Collector/FormatterTest.php
@@ -2,13 +2,13 @@
 
 namespace Http\HttplugBundle\Tests\Unit\Collector;
 
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
 use Http\Client\Exception\HttpException;
 use Http\Client\Exception\TransferException;
 use Http\HttplugBundle\Collector\Formatter;
 use Http\Message\Formatter as MessageFormatter;
 use Http\Message\Formatter\CurlCommandFormatter;
-use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\ResponseInterface;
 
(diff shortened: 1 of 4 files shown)


fbourigault:
> Type annotations does not need to be changed. But I'm happy with this. Just make sure the tests passes.

I was hesitant about this, and ended to change temp to reduce the `use` list. But I will revert those as we are dealing with the PSR interfaces not guzzle implementation.

fbourigault:
I reverted the type annotations, rebased against #177 and reverted to `setExpectedException`. Things should be fine now.

fbourigault:
There is still an error when dependencies are installed using the `--prefer-lowest` flag. The error occurs because `php-http/httplug` `1.0.0` is installed and I use `HttpfulfliedPromise` in my tests. We have to options here, either bump minimum `php-http/httplug` requirement to `1.1.0` or make the code compatible with both versions.

WDYT?

Nyholm:
Bump it to 1.1.0. No questions about it.

fbourigault:
Thank your for reviewing this PR.

PR #177, opened 2017-06-26
build hhvm on trusty box

fbourigault:
Same as https://github.com/php-http/message/pull/77.


Nyholm:
You need to remove hhvm on line 14

fbourigault:
Build is now green! I just fixed one test which was probably working for php because of a mocking side-effect.

Issue #109, opened 2016-08-02
Show plugins created without the Symfony config


Nyholm:
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | question |
| Version |  |

I just tested the bundle together with the knp github client. It is creating a PluginClient and adds a lot of extra plugins. I was (foolishly) expecting to see all plugin steps in the WebProfiler. 

Is it some way we could make this happen?... Or is it anyway one could configure the knp GitHub client differently?

sagikazarmark:
What does "adds lot of extra plugins" mean?

Nyholm:
https://github.com/KnpLabs/php-github-api/blob/master/lib/Github/Client.php#L153-L160

They build up their own stack of plugins and creates a client.

sagikazarmark:
I don't think it's a good idea. The API client should not know about HTTP client internals, it just relies on the interface. You should create a separate factory which can be used for configuring a PluginClient. In that case you can manually create a PluginClient in Symfony and inject it into the client.

dbu:
if there was a bundle for the knp github client, it could also integrate with the HttplugBundle for better configurability. and i agree with mark that this part could be put into a separate factory so that one can inject a tailored client instead of having the github client set up the plugins.

Nyholm:
Yes, I agree with you. It should be a factory for building the HTTP client in the GitHub api client. 

> if there was a bundle for the knp github client, it could also integrate with the HttplugBundle for better configurability. 

If we (in some magic way) could automatically show something like [this](http://docs.php-http.org/en/latest/_images/request-stack.png), but with "GitHub client" instead of "my_guzzle". That would be awesome. 

What if I create a `PluginFactoryInterface` in `client-common` and also a `PluginFactory` implementation. And then I create another PluginFactory here in this bundle with debug features. A GitHubBundle then could make sure GitHubClient is using the PluginFactory form the bundle and we would have lots of debug info in our toolbar.

Nyholm:
My idea worked.. Not sure if it is beautiful... 

![symfony profiler](https://cloud.githubusercontent.com/assets/1275206/17459284/d2430ca4-5c31-11e6-9fff-e5e026c9382c.png)

dbu:
i think it would make sense that this bundle offers such a functionality to other bundles that integrate libraries into symfony. it feels better than trying to somehow magically capture clients or plugins.

[...]

PR #180, opened 2017-07-06
guess client names from collected stacks

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | N/A
| License         | MIT

This is a subset of #109 which aims to display clients names in the profiler from what is collected. The only "negative" change here is we no longer display empty tabs for clients which doesn't sent any request.

## ToDo
* [x] Add basic Collector unit test.


Nyholm:
Thank you for this PR.

PR #181, opened 2017-07-06
display children requests as nested in the profiler

### LINKED ISSUE #109: Show plugins created without the Symfony config

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | update screenshots?
| License         | MIT

While working on #109, I just discovered the refactoring I was doing allows to easily display children requests as nested in the profiler. To keep things small, I decided to do a separate PR which add a nice feature in the profiler and helps to fix #109.

This introduce the concept of active stack. When coming into the StackPlugin, the created stack is made active. Then the request go through every plugins to reach the ProfileClient. The ProfileClient deactivate the stack. Then other code execute, maybe an other request. Then when ProfileClient callbacks are executed, the stack is set as active until we reach StackPlugin callbacks where the stack is marked as inactive.

I also removed the null $stack checks as #133 is definitively solved (since #154).

### DIFF
--- a/Collector/Collector.php
+++ b/Collector/Collector.php
@@ -20,6 +20,11 @@ use Symfony\Component\HttpKernel\DataCollector\DataCollector;
  */
 class Collector extends DataCollector
 {
+    /**
+     * @var Stack|null
+     */
+    private $activeStack;
+
     public function __construct()
     {
         $this->data['stacks'] = [];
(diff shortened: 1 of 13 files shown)


Collector/Collector.php
```diff
@@ -20,6 +20,11 @@
  */
 class Collector extends DataCollector
 {
+    /**
+     * @var Stack
```
Nyholm: Stack|null

Collector/Collector.php
```diff
@@ -42,6 +47,38 @@ public function getName()
...
+
+        $this->activeStack = $stack;
+    }
+
+    /**
+     * Mark the stack as inactive.
+     *
+     * @param Stack $stack
+     */
+    public function desactivateStack(Stack $stack)
```
Nyholm: Typo. `deactivateStack `

Resources/public/style/httplug.css
```diff
@@ -75,6 +75,13 @@
 }
 
 /**
+ * Stack
+ */
+.httplug-stack>.httplug-stack {
+    margin-left: 35px;
```
Nyholm: `4rem` instead.

fbourigault: Is `4rem` a value you want as 35px is not enough? I changed it to `2.5em` which is `35px` on a 96dpi screen.

fbourigault:
@Nyholm all yours comments have been fixed.
I also set minimum `phpunit/php-token-stream` version to `1.1.8` as we use the `finally` keyword.

PR #183, opened 2017-07-07
Use xml to configure discovered clients

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | n/a
| License         | MIT

This move discovered clients configuration from `HttplugExtension` to `services.xml` and `data-collector.xml`.

### To Do
* [x] Add tests when discovered client is set to a configured service.

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -6,11 +6,8 @@ use Http\Client\Common\BatchClient;
 use Http\Client\Common\FlexibleHttpClient;
 use Http\Client\Common\HttpMethodsClient;
 use Http\Client\Common\Plugin\AuthenticationPlugin;
-use Http\Discovery\HttpAsyncClientDiscovery;
-use Http\Discovery\HttpClientDiscovery;
 use Http\HttplugBundle\ClientFactory\DummyClient;
 use Http\HttplugBundle\ClientFactory\PluginClientFactory;
-use Http\HttplugBundle\Collector\ProfileClientFactory;
 use Http\HttplugBundle\Collector\ProfilePlugin;
 use Http\Message\Authentication\BasicAuth;
 use Http\Message\Authentication\Bearer;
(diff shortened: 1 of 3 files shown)


DependencyInjection/HttplugExtension.php
```diff
@@ -359,94 +356,26 @@ private function createUri(ContainerBuilder $container, $serviceId, $uri)
...
-                        $config
-                    ),
-                    $this->isConfigEnabled($container, $config['profiling'])
-                );
-            }
-
-            $httpClient = new Reference($httpClient);
+        if (!empty($httpClient) && $httpClient !== 'auto') {
+            $container->setAlias('httplug.auto_discovery.auto_discovered_client', $httpClient);
+        } elseif (null === $httpClient) {
```
Nyholm: Why "elseif"?

DependencyInjection/HttplugExtension.php
```diff
@@ -359,94 +356,26 @@ private function createUri(ContainerBuilder $container, $serviceId, $uri)
...
-                        $config
-                    ),
-                    $this->isConfigEnabled($container, $config['profiling'])
-                );
-            }
-
-            $asyncHttpClient = new Reference($asyncHttpClient);
+        if (!empty($asyncHttpClient) && $asyncHttpClient !== 'auto') {
+            $container->setAlias('httplug.auto_discovery.auto_discovered_async', $asyncHttpClient);
+        } elseif (null === $asyncHttpClient) {
```
Nyholm: Same here..

DependencyInjection/HttplugExtension.php
```diff
@@ -359,94 +356,26 @@ private function createUri(ContainerBuilder $container, $serviceId, $uri)
...
+            $container->removeDefinition('httplug.auto_discovery.auto_discovered_async');
+            $container->removeDefinition('httplug.collector.auto_discovered_async');
         }
 
         if (null === $httpClient && null === $asyncHttpClient) {
```
Nyholm: Is this check still valid?

DependencyInjection/HttplugExtension.php
fbourigault: Because possible values are `null`, `auto` or any configured client name. So the first case handle the "any configured client name" and the elseif the null case which means remove the definition from the container.

DependencyInjection/HttplugExtension.php
fbourigault: Yes, If we have no discovered client (sync and async), we remove the useless listener.

fbourigault:
@dbu I will make conditional loading of some xml files. Far better than loading services and to remove them in the extension.I still have to fix a broken test I added in #184 and then I will improve configuration loading.

DependencyInjection/HttplugExtension.php
```diff
@@ -359,94 +356,34 @@ private function createUri(ContainerBuilder $container, $serviceId, $uri)
...
-                        HttpClientDiscovery::class,
-                        'auto_discovered_client',
-                        $config
-                    ),
-                    $this->isConfigEnabled($container, $config['profiling'])
-                );
-            }
-
-            $httpClient = new Reference($httpClient);
+        if (!empty($httpClient) && $httpClient !== 'auto') {
```
Nyholm:
```php
if ($httpClient !== 'auto') {
   //...
  if (!empty($httpClient) {
     //...
  }
}
```

fbourigault:
I've update the code to make #184 tests fine.

About conditional xml loading, I don't think it worth it. I would need to load 5 different files (`sync`, `async`, `sync_with_profiling`, `async_with_profiling` and `httplug.discovery`).

I think it would be easier later as the profiling stuff will be handled differently (in the PluginClientFactory injected factory callback).

DependencyInjection/HttplugExtension.php
fbourigault: I don't used to write nested conditions (object calisthenics :heart:) but I will make an exception.

fbourigault:
Thank you for reviews!

PR #184, opened 2017-07-07
add discovery functional tests when discovery is forced or disabled

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #183 
| Documentation   | n/a
| License         | MIT

This add more tests related to discovery and will help to get #183 refactoring safe.


Tests/Resources/app/AppKernel.php
```diff
@@ -28,9 +28,10 @@ public function registerBundles()
      */
     public function registerContainerConfiguration(LoaderInterface $loader)
     {
-        $this->isDebug()
-            ? $loader->load(__DIR__.'/config/config_debug.yml')
-            : $loader->load(__DIR__.'/config/config.yml');
+        $loader->load(__DIR__.'/config/config_'.$this->getEnvironment().'.yml');
+        if ($this->isDebug()) {
+            $loader->load(__DIR__.'/config/config_debug.yml');
+        }
```
Nyholm: else {
//load config/config.yml
}

fbourigault: `config.yml` have been renamed to `config_test.yml` to always have a configuration file matching the current environment.

Tests/Functional/DiscoveredClientsTest.php
```diff
@@ -71,9 +71,32 @@ public function testDiscoveredAsyncClientWithProfilingEnabled()
...
+        $this->assertFalse($container->has('httplug.auto_discovery.auto_discovered_async'));
+        $this->assertFalse($container->has('httplug.strategy'));
+    }
+
+    public function testForcedDiscovery()
+    {
+        $container = $this->getContainer(true, 'discovery_forced');
+
+        $this->assertFalse($container->has('httplug.auto_discovery.auto_discovered_client'));
+        $this->assertFalse($container->has('httplug.auto_discovery.auto_discovered_async'));
```
Nyholm: Should these be false on both `testDisabledDiscovery` and `testForcedDiscovery`?

fbourigault: Yes, when forced, we use the service id provided by the user. IMHO, those services shouldn't be public as we use them as `httplug.discovery` arguments.

PR #185, opened 2017-07-12
use service locator for configured client strategy

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes?
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

When available (Symfony 3.3+), a `ServiceLocator` is used in the `ConfiguredClientStrategy` to lazy load discoverable clients. A fallback injecting the Container is implemented for older Symfony versions.

#### Why?

This is required to solve a chicken/egg problem for #109. This should also improve a few resources usage by only building clients when discovery is called (before discovered clients were built at `kernel.request` or `console.command`).

### DIFF
--- a/Tests/Functional/DiscoveredClientsTest.php
+++ b/Tests/Functional/DiscoveredClientsTest.php
@@ -6,6 +6,7 @@ use Http\Client\Common\PluginClient;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Http\HttplugBundle\Collector\StackPlugin;
+use Http\HttplugBundle\Discovery\ConfiguredClientsStrategy;
 use Nyholm\NSA;
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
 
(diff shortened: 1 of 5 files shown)


Tests/Functional/DiscoveredClientsTest.php
```diff
@@ -88,10 +89,10 @@ public function testForcedDiscovery()
         $this->assertFalse($container->has('httplug.auto_discovery.auto_discovered_async'));
         $this->assertTrue($container->has('httplug.strategy'));
 
-        $strategy = $container->get('httplug.strategy');
+        $container->get('httplug.strategy');
 
-        $this->assertEquals($container->get('httplug.client.acme'), NSA::getProperty($strategy, 'client'));
-        $this->assertEquals($container->get('httplug.client.acme'), NSA::getProperty($strategy, 'asyncClient'));
+        $this->assertEquals($container->get('httplug.client.acme'), ConfiguredClientsStrategy::getCandidates(HttpClient::class)[0]['class']());
```
Nyholm: This is not proper PHP 5 code. 

```php
$foo = ConfiguredClientsStrategy::getCandidates(HttpClient::class)[0]['class'];
$foo();
```

fbourigault:
Thank you for reviews!

Issue #188, opened 2017-07-14
Error with auto discovery


Nyholm:
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | dev-master

#### Actual Behavior
```
ServiceCircularReferenceException: Circular reference detected for service "httplug.auto_discovery.auto_discovered_client", path: "httplug.auto_discovery.auto_discovered_client".
```

#### Steps to Reproduce

```yaml
httplug:
  clients:
    curl:
      factory: 'httplug.factory.curl'
      http_methods_client: true
```

```php
HttpClientDiscovery::find();
```

#### To fix this PR

I added a failing test for this issue. Please make a PR to branch [issue-188](https://github.com/php-http/HttplugBundle/tree/issue-188)

PR #189, opened 2017-07-15
Revert "use service locator for configured client strategy"

Nyholm:
This reverts commit c5ac6da7516d6390b7cf8ce01b74fddfe007a839. 

PR #185 made discovery so extra lazy so it does not work. See #188.


fbourigault:
Thank you!

PR #193, opened 2017-07-17
Don't use dependency injection decoration for plugin profiling

### LINKED ISSUE #191: Profiling hide original service definition
Once decorated, the following plugin no longer have `setAuthorization` method available in the `App\Http\Plugin\AuthorizationPlugin` service because of being decorated with `ProfilePlugin`. To get access to the original plugin, you have to reference it using the `App\Http\Plugin\AuthorizationPlugin.debug.inner` name.

```php
namespace App\Http\Plugin;

use Http\Client\Common\Plugin;

class AuthorizationPlugin implements Plugin
{
    /**
     * @var string
     */
    private $authorization;

    /**
     * @param string $authorization
     */
    public function setAuthorization($authorization)
    {
        $this->authorization = $authorization;
    }

    /**
     * {@inheritdoc}
     */
    {
        if (!$this->authorization) {
            return $next($request);
        }

        return $next($request->withHeader('Authorization', $this->authorization));
    }
}
```
```yaml
App\Http\Plugin\AuthorizationPlugin:
    class: App\Http\Plugin\AuthorizationPlugin
```

The ongoing work with `PluginClientFactory` is supposed to fix this. The only thing to consider is about possible BC break.

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #191 
| Documentation   | n/a
| License         | MIT

This fix #191 by no longer use Symfony dependency injection decoration for plugin profiling.

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -273,9 +273,11 @@ class HttplugExtension extends Extension
         $pluginClientOptions = [];
         if ($profiling) {
             //Decorate each plugin with a ProfilePlugin instance.
-            foreach ($plugins as $pluginServiceId) {
+            $plugins = array_map(function ($pluginServiceId) use ($container) {
                 $this->decoratePluginWithProfilePlugin($container, $pluginServiceId);
-            }
+
+                return $pluginServiceId.'.debug';
+            }, $plugins);
 
             // To profile the requests, add a StackPlugin as first plugin in the chain.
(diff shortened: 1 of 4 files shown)


fbourigault:
You are right! I was in a hurry to fix this and tests are green... I will try to add one!

fbourigault:
Fixed!

PR #196, opened 2017-08-03
Fix profiler display when using in a symfony/flex project

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

This fix the profiler panel display when used in a symfony/flex application.

I added a test to ensure the panel is loading (the case is with no requests sent).

The bug was caused by the profiler template path which was using the old twig path syntax (no `@` and `::`).

I also found that the profiler is dependent on the `symfony/asset` component.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "symfony/framework-bundle": "^2.8 || ^3.0",
         "php-http/message": "^1.4",
         "php-http/discovery": "^1.0",
-        "twig/twig": "^1.18 || ^2.0"
+        "twig/twig": "^1.18 || ^2.0",
+        "symfony/asset": "^2.8 || ^3.0"
     },
     "require-dev": {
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -45,6 +46,7 @@
         "symfony/web-profiler-bundle": "^2.8 || ^3.0",
         "symfony/finder": "^2.7 || ^3.0",
         "symfony/cache": "^3.1",
+        "symfony/browser-kit": "^2.8 || ^3.0",
```
Nyholm: Why do we need this?

fbourigault: We need it for `testShowProfiler`.

Nyholm:
Thank you

PR #198, opened 2017-08-04
Added more clear exception when logger service is not found

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes.. maybe
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

When one is using the Logger plugin and the default MonologBundle is not installed you would get an error message like:

> The service "httplug.client" has a dependency on a non-existent service "logger".

This PR makes that error message more clear. 

Ping @Koc can you confirm that you had the logger plugin configured?

Ref: https://github.com/symfony/recipes-contrib/issues/67#issuecomment-320227761

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -162,6 +162,9 @@ class HttplugExtension extends Extension
                 $definition->replaceArgument(0, new Reference($config['journal']));
                 break;
             case 'logger':
+                if (!$container->has($config['logger'])) {
+                    throw new \LogicException(sprintf('To use the LoggerPlugin you must specify a service that implements PSR3 at httplug.plugins.logger.logger. Current value "%s" is not a valid service.', $config['logger']));
+                }
                 $definition->replaceArgument(0, new Reference($config['logger']));
                 if (!empty($config['formatter'])) {
                     $definition->replaceArgument(1, new Reference($config['formatter']));


Koc:
Can we disable logger plugin in this case and add warning instead of exception? Make it optional

Nyholm:
It is disabled by default. Are you saying it isn't? What does your confit look like?

Koc:
Now ok. I think that we can enable it when debug enabled if logger available. See similar case in Symfoby

https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Configuration.php#L605

DependencyInjection/HttplugExtension.php
```diff
@@ -162,6 +162,9 @@ private function configurePluginByName($name, Definition $definition, array $con
                 $definition->replaceArgument(0, new Reference($config['journal']));
                 break;
             case 'logger':
+                if (!$container->has($config['logger'])) {
```
stof: this is not working. The ContainerBuilder you get in the extension is not the main one. It is an empty one, and services added in it by the extension are then merged into the main ContainerBuilder.

the reason for that is to forbid extension to change logic based on which services are defined by other bundles (as they might not have been called yet at this point)

Nyholm: You are correct. Thank you for the review.

dbu:
the correct way to build such logic, if needed, is a compiler pass. those
see all bundle's services

Am 04.08.2017 5:32 nachm. schrieb "Tobias Nyholm" <notifications@github.com

PR #201, opened 2017-08-22
minor profiler updates

fbourigault:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | n/a
| License         | MIT

This is part of #109 and solve some edge cases when displaying the profiler panel:

* don't display empty client tabs in profiler. This happen when the discovery is only used by a third party library.
* return only collector root stacks. This improve template readability.
* better count of client messages. When discovery is forced to a configured client, the message count in the client tab is wrong. This introduce a different way to count messages starting by root stacks and going down to nested requests.

### DIFF
--- a/Collector/Collector.php
+++ b/Collector/Collector.php
@@ -131,9 +131,13 @@ class Collector extends DataCollector
      */
     public function getClients()
     {
+        $stacks = array_filter($this->data['stacks'], function (Stack $stack) {
+            return $stack->getParent() == null;
+        });
+
         return array_unique(array_map(function (Stack $stack) {
             return $stack->getClient();
-        }, $this->data['stacks']));
+        }, $stacks));
     }
 
     /**
@@ -141,11 +145,39 @@ class Collector extends DataCollector
      *
      * @return Stack[]
      */
-    public function getClientStacks($client)
+    public function getClientRootStacks($client)
     {
         return array_filter($this->data['stacks'], function (Stack $stack) use ($client) {
-            return $stack->getClient() == $client;
+            return $stack->getClient() == $client && $stack->getParent() == null;
         });
+    }
+
+    /**
+     * Count all messages for a client.
+     *
+     * @param $client
+     *
+     * @return int
+     */
+    public function countClientMessages($client)
+    {
+        return array_sum(array_map(function (Stack $stack) {
+            return $this->countStackMessages($stack);
+        }, $this->getClientRootStacks($client)));
+    }
+
+    /**
+     * Recursively count message in stack.
+     *
+     * @param Stack $stack
+     *
+     * @return int
+     */
+    private function countStackMessages(Stack $stack)
+    {
+        return 1 + array_sum(array_map(function (Stack $child) {
+            return $this->countStackMessages($child);
+        }, $this->getChildrenStacks($stack)));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


fbourigault:
I re-added `getClientStacks` with a deprecation notice.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## Unreleased
+
+### Changed
+
+- The `Http\HttplugBundle\Collector\Collector::getClientStack` method returns only stacks without parent.
```
dbu: Stack or Stacks in the method name?

fbourigault: `Stacks`!

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+
+### Deprecated
+
+- The `Http\HttplugBundle\Collector\Collector` `getClientStacks` method is deprecated. Use `getClientRootStacks` method instead.
```
Nyholm: `Http\HttplugBundle\Collector\Collector::getClientStacks`

Nyholm: Should we really deprecate the method AND change its behaviour? If fine with deprecate it and leave its current behaviour.

Nyholm:
Oh, my comments disappeared but are still valid: https://github.com/php-http/HttplugBundle/pull/201#discussion_r134944991

fbourigault:
As the class is internal I decided to keep the method as a proxy. Maybe I should not document the behavior change. Behavior changed for `getClients` but its not documented. WDYT?

dbu:
actually if we leave a BC method, we can just as well leave it unchanged in behaviour. might be the easiest.

Collector/Collector.php
```diff
@@ -131,9 +131,13 @@ public function getFailedStacks()
      */
     public function getClients()
     {
+        $stacks = array_filter($this->data['stacks'], function (Stack $stack) {
+            return $stack->getParent() == null;
```
Nyholm: Should be `===`

Collector/Collector.php
```diff
@@ -143,12 +147,52 @@ public function getClients()
      */
     public function getClientStacks($client)
     {
+        @trigger_error(sprintf('Method %s() is deprecated since version 1.8 and will be removed in 2.0. Use getClientRootStacks() instead.', __METHOD__), E_USER_DEPRECATED);
```
Nyholm: Remove this method completely or leave it unchanged (and deprecate it). It is fine to remove it since the Collector class is `@internal`.

dbu:
on the BC topic: my point was that even though the class is marked as internal, its used in the template so it might be customized. but i agree, lets drop the method. at worst, somebody gets an exception in their customized toolbar template. lets just do an entry in the changelog that the method has been renamed and only returns root stacks now.

fbourigault:
Do we really need to clutter the changelog with internal changes?

dbu:
the one person having a problem would probably open an issue here if they don't find what happened, so i'd say i am fine without a changelog entry.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## Unreleased
+
+### Changed
+
+- The `Http\HttplugBundle\Collector\Collector` `getClientStacks` method have been renamed to `getClientRootStacks`
+and now returns only stacks without any parent.
```
dbu: i am fine to remove this if you want.

fbourigault: Nice, I didn't saw the _without_ in you previous comment!

PR #203, opened 2017-08-28
Moved TEST_COMMAND to global section

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

I did also add php 7.2

### DIFF
--- a/Tests/Functional/ProfilingTest.php
+++ b/Tests/Functional/ProfilingTest.php
@@ -15,11 +15,12 @@ use Http\HttplugBundle\Collector\StackPlugin;
 use Http\Message\Formatter\CurlCommandFormatter;
 use Http\Message\Formatter\FullHttpMessageFormatter;
 use Http\Mock\Client;
+use PHPUnit\Framework\TestCase;
 use Psr\Http\Message\RequestInterface;
 use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Symfony\Component\Stopwatch\Stopwatch;
 
-class ProfilingTest extends \PHPUnit_Framework_TestCase
+class ProfilingTest extends TestCase
 {
     /**
      * @var Collector
@@ -63,13 +64,14 @@ class ProfilingTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals('example.com', $stack->getRequestHost());
     }
 
+    /**
+     * @expectedException \Exception
+     */
     public function testProfilingWhenPluginThrowException()
     {
         $client = $this->createClient([
             new ExceptionThrowerPlugin(),
         ]);
-
-        $this->setExpectedException(\Exception::class);
 
         try {
             $client->sendRequest(new Request('GET', 'https://example.com'));
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,11 @@
             "Http\\HttplugBundle\\": ""
         }
     },
+    "autoload-dev": {
+        "classmap": [
+            "Tests/Resources/app/AppKernel.php"
+        ]
+    },
     "scripts": {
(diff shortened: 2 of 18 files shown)


fbourigault:
To see if dependencies have an incoming release with PHP 7.2, it would make sense to use PHP 7.2 here:
https://github.com/php-http/HttplugBundle/blob/766687eef1562f292f6561247aeb32f37ed40cc9/.travis.yml#L29-L30

fbourigault:
Do you plan to solve failing tests in this PR?

fbourigault:
:tada: HttplugBundle is now successfully tested against PHP 7.2

- I added the KERNEL_CLASS to phpunit.xml.dist because it will be deprecated in 3.4.
- I removed a `each` function call which is deprecated since PHP 7.2.
- Dev dependencies are now tested against PHP 7.2.
- As `each` is deprecated, we need to use PHPUnit 6, so, I bumped older version of PHPUnit to the first one which contains the forward compatibility layer and did some changes to make tests working in every allowed PHPUnit version.
- I also fixed a tiny styleci complain about missing blank line before `break` statement.
- I disabled xdebug when code coverage is not enabled to speed up the build.

Tests/Functional/ProfilingTest.php
```diff
@@ -63,14 +64,15 @@ public function testProfilingWithCachePlugin()
         $this->assertEquals('example.com', $stack->getRequestHost());
     }
 
+    /**
+     * @expectedException \Exception
```
dbu: i discussed once with sebastian bergman and he is saying he agrees that the annotation is less clean than calling setExpectedException. the reason is that if the exception is thrown too early, the test would wrongly be green. this is particularly true when we expect such a generic exception

composer.json
```diff
@@ -49,7 +50,7 @@
         "symfony/browser-kit": "^2.8 || ^3.0",
         "symfony/dom-crawler": "^2.8 || ^3.0",
         "polishsymfonycommunity/symfony-mocker-container": "^1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "^1.0",
+        "matthiasnoback/symfony-dependency-injection-test": "^1.0 || ^2.0",
```
dbu: do we need both versions to support the range of php versions we support?

fbourigault: Yes, we can't install PHPUnit 6.0 because of a conflict on `sebastian/exporter`.

Tests/Functional/ProfilingTest.php
fbourigault: So, it's better to have a if/else at the right position in the test?

dbu: why if/else? afaik the $this->setExpectedException call is the best practice way. this makes sure the exception is only considered correct when thrown after that line in the code, and you have namespace resolution.

fbourigault: `setExpectedException` was removed in 6.0 and replaced by `expectException`. That's why I moved to annotation.

Nyholm: Yeah, We need the annotation for BC purpose...

Nyholm:
Thank you. I'm happy to merge this.

Tests/Functional/ProfilingTest.php
fbourigault: We can keep this as-is and switch to `expectException` in next-major.

Nyholm:
What do you think @dbu?

fbourigault:
Thank you all!

PR #205, opened 2017-09-03
Make sure we use different cache directories every time we run the tests

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes, kind of
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

When you run the tests you build some cache in your system's temp directory. If you say, bump some versions that cache may not be properly cleared. This fix make sure we always start with a clear cache every time we run the tests.

### DIFF
--- a/Tests/Resources/app/AppKernel.php
+++ b/Tests/Resources/app/AppKernel.php
@@ -12,6 +12,11 @@ class AppKernel extends Kernel
     use MicroKernelTrait;
 
     /**
+     * @var string
+     */
+    private static $cachePrefix;
+
+    /**
      * {@inheritdoc}
      */
     public function registerBundles()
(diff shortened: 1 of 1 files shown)


Tests/Resources/app/AppKernel.php
```diff
@@ -12,6 +12,11 @@ class AppKernel extends Kernel
     use MicroKernelTrait;
 
     /**
+     * @var string
+     */
+    private static $cachePrefix;
```
dbu: nitpick: this is not just the prefix but the whole folder name

Nyholm: Thanks.

fbourigault:
StyleCI errors are fixed in #203.

Nyholm:
The PR is updated and ready to be merged.

PR #208, opened 2017-09-05
Add autowiring for RequestFactory and ResponseFactory

stof:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | in/a
| License         | MIT

#### What's in this PR?

MessageFactory has 2 parent interfaces (allowing to typehint against only the part you want). This makes these interfaces work for autowiring too (by using the existing message_factory service, as it implements these interfaces due to this inheritance)

#### Why?

Autowiring is great, and the bundle should make it work fine.


Nyholm:
Thank you @xabbuh for the review.

PR #209, opened 2017-09-06
Use flex client wrapper for profiling if needed

### LINKED ISSUE #206: ProfileClient overrides other feature clients
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | dev-master

#### Actual Behavior

When HttplugBundle's client "foo" is configured with `HttpMethodsClient`. If "foo" is used in discovery it will be picked up by third party bundles. If that third party bundle uses the new ClientFactory from php-http/common we will get an error [in the `ProfileClient`](https://github.com/php-http/HttplugBundle/blob/master/Collector/ProfileClient.php#L58) since `HttpMethodClient` is not a `HttpClient` AND `HttpAsyncClient`. 

Also, even if `HttpMethodClient` were to implement `HttpClient` and `HttpAsyncClient`, the `ProfileClient` would not implement the features/functions of the `HttpMethodClient`.

We should reconsider if decorating pattern is feasible here. If that is our only option, we should maybe consider using dynamic proxys for creating the `ProfileClient`. I made something like that in the [CacheBundle](https://github.com/php-cache/cache-bundle/pull/75)

-------------

Edit: If we decorate a client with PluginClient we will not expect the functions from MyCustomClient to be available. 

However, the PluginClient must accept an HttpClient **or** an AsyncHttpClient.

Nyholm:
fix #206

FYI @fbourigault

### DIFF
--- a/Collector/ProfileClient.php
+++ b/Collector/ProfileClient.php
@@ -47,8 +47,8 @@ class ProfileClient implements HttpClient, HttpAsyncClient
     private $eventNames = [];
 
     /**
-     * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement both HttpClient and
-     *                                              HttpAsyncClient interfaces.
+     * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement both HttpClient or
+     *                                              HttpAsyncClient interface.
      * @param Collector                  $collector
      * @param Formatter                  $formatter
      * @param Stopwatch                  $stopwatch
@@ -56,14 +56,9 @@ class ProfileClient implements HttpClient, HttpAsyncClient
     public function __construct($client, Collector $collector, Formatter $formatter, Stopwatch $stopwatch)
     {
         if (!($client instanceof HttpClient && $client instanceof HttpAsyncClient)) {
-            throw new \RuntimeException(sprintf(
-                '%s first argument must implement %s and %s. Consider using %s.',
-                    __METHOD__,
-                    HttpClient::class,
-                    HttpAsyncClient::class,
-                    FlexibleHttpClient::class
-            ));
+            $client = new FlexibleHttpClient($client);
         }
+
         $this->client = $client;
         $this->collector = $collector;
         $this->formatter = $formatter;
--- a/Collector/ProfileClient.php
+++ b/Collector/ProfileClient.php
@@ -47,7 +47,7 @@ class ProfileClient implements HttpClient, HttpAsyncClient
     private $eventNames = [];
 
     /**
-     * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement both HttpClient or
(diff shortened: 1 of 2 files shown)


Tests/Functional/Issue206.php
```diff
@@ -0,0 +1,26 @@
...
+use Http\Client\Common\HttpMethodsClient;
+use Http\Client\Common\PluginClientFactory;
+use Http\Discovery\HttpClientDiscovery;
+use Http\Discovery\MessageFactoryDiscovery;
+use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
+use Http\HttplugBundle\Collector\PluginClientFactory as DebugFactory;
+
+class Issue206 extends WebTestCase
+{
+    public function testCustomClientStillAvailable()
```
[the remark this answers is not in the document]
Nyholm: Thank you

Tests/Functional/Issue206.php
```diff
@@ -22,5 +22,8 @@ public function testCustomClientStillAvailable()
         // Create a client
         $myCustomClient = new HttpMethodsClient(HttpClientDiscovery::find(), MessageFactoryDiscovery::find());
         $pluginClient = (new PluginClientFactory())->createClient($myCustomClient, []);
+
+        // If we get to this line, no exceptions has been thrown.
```
fbourigault: Is this still relevant as we removed the `throw` statement from `ProfileClient`. Maybe you could use `NSA` to check that the `$client` inside `$pluginClient` is a `FlexibleHttpClient`.

fbourigault:
This one was left aside since ages :/ Could you rebase to check that build is still green? This is required for next release.

Nyholm:
Im done with this PR. We should use the flexClient. 

This will fix #206.

fbourigault:
Thank you!

Collector/ProfileClient.php
```diff
@@ -47,23 +47,18 @@ class ProfileClient implements HttpClient, HttpAsyncClient
...
+     * @param HttpClient|HttpAsyncClient $client    The client to profile. Client must implement HttpClient or
+     *                                              HttpAsyncClient interface.
      * @param Collector                  $collector
      * @param Formatter                  $formatter
      * @param Stopwatch                  $stopwatch
      */
     public function __construct($client, Collector $collector, Formatter $formatter, Stopwatch $stopwatch)
     {
         if (!($client instanceof HttpClient && $client instanceof HttpAsyncClient)) {
-            throw new \RuntimeException(sprintf(
```
xabbuh: Shouldn't the exception still be thrown if neither of the interfaces is implemented? I know that the `FlexibleHttpClient` throws anyway, but I am not sure if relying on that behaviour would be a good idea and also could the stack trace be confusing for users.

Nyholm: Yes, an exception are thrown, But yes. It could be a good idea to throw en exception here as well.

PR #211, opened 2017-10-22
Update composer.json for Symfony 4.

dbrumann:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | ---
| Documentation   | ---
| License         | MIT

#### What's in this PR?

Update Symfony-related dependencies to allow 4.x-versions to make this bundle work in Symfony 4.0 applications. This PR requires php-http/stopwatch-plugin#4 to work.


.travis.yml
```diff
@@ -33,6 +33,8 @@ matrix:
           # Test against LTS versions
         - php: 7.0
           env: SYMFONY_VERSION=2.8.*
+        - php: 7.1
+          env: SYMFONY_VERSION=4.0.*
```
Nyholm: Write version 3.4 here instead. Then add another entry for 4.0 above the comment say "Test against LTS versions"

Soullivaneuh:
@Nyholm It seems to be already done: https://github.com/php-http/HttplugBundle/pull/211/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R59

dbu:
can you please rebase this on master? i just merged #216 which should fix the fatal error.

dbu:
and we need to figure out if we simply need to make those services public that we get warnings for in symfony 3.4, or if we can wire things so as to not need to get those service from the container but inject them instead.

fbourigault:
All services the container complains about have to be public to keep backward compatibility!

dbu:
> All services the container complains about have to be public to keep backward compatibility!

they only become inaccessible in symfony 4, so thats no BC break of us. discusssion is in #218, it looks like most of them are explicitly set to private...

xabbuh:
@dbrumann Any updates? Will you be able to work on this? :)

dbrumann:
Very funny @xabbuh. I will update this PR on Friday

dbu:
> support Symfony 2.8 - 4.0 in next minor release?

unless there is a big problem popping up with that yes, i would want to support all those.

when the user updates to symfony 4, services that have been public only by accident will become private, which is imo a BC break of symfony itself, not of this bundle, and thus seems acceptable. we should nonetheless list the services in the changelog to help the users with upgrading.

composer.json
```diff
@@ -42,13 +42,13 @@
         "php-http/buzz-adapter": "^0.3",
         "php-http/mock-client": "^1.0",
         "symfony/phpunit-bridge": "^3.2",
```
xabbuh: What about allowing 4.0 here too?

dbu:
i think the version 4 can't install because one of our dependencies does not support it. do you see which one it is?

the 3.4 build has warnings about accessing private services. i think the httplug clients should be public. for the rest we have to see, we maybe also need to change some tests to not try to get the service from the container in this way.

Nyholm:
@dbrumann Do you mind if I try to make this work?

dbrumann:
@Nyholm Not at all. Since prophecy and phpspec merged the Symfony 4 updates I should be able to make php-http/stopwatch-plugin pass now and then hopefully this PR as well, but feel free to add commits. I will refrain from doing any changes requiring a force push.

Nyholm:
Much cleaner, thank you!

Nyholm:
I think you are done with this PR. You are blocked by #221. When that PR is merged you just have to rebase and everything will be green

Nyholm:
Hey @dbrumann. If you do a final rebase on master then you should be just fine. I've updated .travis.yml to test all proper symfony versions.

Tests/Functional/ServiceInstantiationTest.php
```diff
@@ -24,17 +24,17 @@ public function testHttpClient()
     {
         static::bootKernel();
         $container = static::$kernel->getContainer();
-        $this->assertTrue($container->has('httplug.client'));
-        $client = $container->get('httplug.client');
```
dbu: i think we should make the `httplug.client` service public too, it seems correct to me to reference it by that name.

Tests/Resources/app/config/config_test.yml
```diff
@@ -26,3 +26,4 @@ httplug:
 services:
     app.http.plugin.custom:
         class: Http\Client\Common\Plugin\RedirectPlugin
+        public: true
```
dbu: why do the plugins need to be public? that smells like an issue in how we wire them into the plugin-client, no?

fbourigault: Isn't just because test assertions are written against the compiled container instead of the container builder?

dbu: indeed, seems the problem is in ServiceInstantiationTest::testProfilingShouldNotChangeServiceReference where we get the container from the kernel and then assert that nothing messes with the plugin service. i don't know if there is an easy solution for this - maybe simply add a comment here to avoid confusion? `# plugin services usually do not need to be public. this one is made public so that we can do functional tests on the service`

xabbuh: That's basically the approach we still need to document with now "private by default" services in Symfony 4 (see symfony/symfony-docs#8097). In this case, we do not need to work around with an alias.

Tests/Functional/ServiceInstantiationTest.php
dbrumann: According to the current configuration it is configured as private. I can restore this test, but then I would also change the `public=false` in the service configuration to make clear that this id can be used.

Tests/Resources/app/config/config_test.yml
dbrumann: Are there any preferences as to whether I should use a test alias or just add a comment? I personally prefer the comment. Making this public for tests because we want to get it from the container feels more honest to me.

xabbuh: The alias IMO is only needed when the service would be present outside the test too (which is not the case here).

dbu: agreed, lets just add a comment

Resources/config/services.xml
```diff
@@ -4,47 +4,47 @@
...
             <factory class="Http\Discovery\HttpAsyncClientDiscovery" method="find" />
         </service>
         <service id="Http\Client\HttpAsyncClient" alias="httplug.async_client.default" public="false" />
 
-        <service id="httplug.client.default" class="Http\Client\HttpClient">
+        <service id="httplug.client.default" class="Http\Client\HttpClient" public="true">
             <factory class="Http\Discovery\HttpClientDiscovery" method="find" />
         </service>
-        <service id="Http\Client\HttpClient" alias="httplug.client" public="false" />
+        <service id="Http\Client\HttpClient" alias="httplug.client" public="true" />
```
dbu: hm, is this what we need to make public? this is just the alias to enable autowiring. this is not defining httplug.client

xabbuh: I agree. These aliases shouldn't be public (in Symfony, they are private too).

dbu:
i looked at the code and don't see where the service httplug.client is defined. its neither in services.xml nor in the HttplugExtension, it seems

Nyholm:
The `httplug.client` definition is super old code.. 

It is defined here: https://github.com/php-http/HttplugBundle/blame/v1.7.1/DependencyInjection/HttplugExtension.php#L53

fbourigault:
I think we should follow the Symfony 4 trend and let those service became private in Symfony 4. Users will get deprecation notices if getting those services using `ContainerInterface::get`.

For the test, could we rewrite it to not write assertion on compiled container? If no adding a comment where the `public=true` is set would be great!

fbourigault:
Could you rebase again to get https://github.com/php-http/HttplugBundle/pull/225 and really run tests against Symfony 4?

Nyholm:
Thank you @dbrumann. This was merged via #228.

PR #225, opened 2017-11-27
remove stability settings from composer.json

xabbuh:
With `prefer-stable` set to `true`, the `DEPENDECIES=dev` build job
will (almost) never install not yet released dependencies. Without
the `prefer-stable` setting we should also drop the `minimum-stability`
setting (which is now useless as the config value is set in the Travis
job).


Nyholm:
Awesome. Thanks!

PR #226, opened 2017-11-28
turn off unneeded noise in tests

xabbuh:
There is no need to log all notices when running tests. Seeing just critical errors is sufficient.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,6 +6,10 @@
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
          syntaxCheck="true">
+    <php>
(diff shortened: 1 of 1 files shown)


Nyholm:
I do not see any difference from the [output on master](https://travis-ci.org/php-http/HttplugBundle/jobs/308660487) and [this PR](https://travis-ci.org/php-http/HttplugBundle/jobs/308666991). 
Could you show me?

xabbuh:
I cannot right now. 🙈 I had expected to get version 3.4 of the HttpKernel component. But it looks like there is no requirement for this component at all.

xabbuh:
@Nyholm now you can see the difference: https://travis-ci.org/php-http/HttplugBundle/jobs/308680217 :)

PR #228, opened 2017-11-28
Add Symfony 4 support

### LINKED ISSUE #218: should autodiscovered clients be public or private?
when building with symfony 3.4, we get warnings that we access private services. but at least some of those are [explicitly set to private](https://github.com/php-http/HttplugBundle/blob/542d39d9c03ebd3044f35fbd08fb36cc87fde941/DependencyInjection/HttplugExtension.php#L374). what was the intention of creating those alias but then setting them to private? 

the other failure is on httplug.client - should that not be public? and what about named clients? should they be public or not? imo the plugins and stuff should be private and only handled in DI, but for the clients themselves, they should be public.

@Nyholm and @fbourigault your names show up in git blame on those lines, what do you think? shall i just remove a bunch of `setPublic(false)` calls?

fbourigault:
This is #211 with a hack to get services private in Symfony 4 without rewriting a large part of tests.

I fixedup @dbrumann commits and added mine to get @dbrumann credited.

Close #211. Close #218.

The `PublicServicesForFunctionalTestsPass` is a hack and we must rewrite tests to remove this in the future.


DependencyInjection/HttplugExtension.php
```diff
@@ -297,6 +298,7 @@ private function configureClient(ContainerBuilder $container, $clientName, array
 
         $container
             ->register($serviceId, PluginClient::class)
+            ->setPublic(true)
```
fbourigault: How can we avoid to set this one public just for tests?

Nyholm: Psudo code:
```
$services = $container->getAllServiceDefinitionKeys()
foreach ($services as $service) {
  if ($service->getId() matches "httplug.client.*") {
    $container->getDefinition($service)->setPublic(true);
  }
}
```

I do not know any better... sorry

Nyholm: Btw, will the tests fail if we remove this line?

Tests/Resources/app/AppKernel.php
```diff
@@ -86,4 +87,32 @@ public function indexAction()
...
+            'httplug.message_factory.default',
+            'httplug.stream_factory.default',
+            'httplug.uri_factory.default',
+            'httplug.async_client.default',
+            'httplug.client.default',
+            'app.http.plugin.custom',
+        ];
+        foreach ($services as $service) {
+            if ($container->hasDefinition($service)) {
+                $container->getDefinition($service)->setPublic(true);
```
dbu: this could be quite confusing when reading the tests, or when creating a new test. is there any chance we can instead have our tests operate on the container builder instead of the closed container?

DependencyInjection/HttplugExtension.php
dbu: if we keep the compiler pass, this line can be reverted, right?

Tests/Resources/app/AppKernel.php
fbourigault: Yes, that's the idea but it's too many work ATM.

fbourigault:
This version looks really fine to me. @dbu @Nyholm what do you think?

fbourigault:
We are still waiting for `matthiasnoback/symfony-dependency-injection-test`, `matthiasnoback/symfony-config-test` and `php-http/stopwatch-plugin` to really test against Symfony 4 only components.

DependencyInjection/HttplugExtension.php
```diff
@@ -50,7 +51,7 @@ public function load(array $configs, ContainerBuilder $container)
 
         // Set main aliases
         foreach ($config['main_alias'] as $type => $id) {
-            $container->setAlias(sprintf('httplug.%s', $type), $id);
+            $container->setAlias(sprintf('httplug.%s', $type), new Alias($id, true));
```
xabbuh: the changes in this file should then be reverted too

fbourigault:
I removed the public aliases.

PR #229, opened 2017-11-28
update the list of required dependencies

xabbuh:
To make it easier to see which dependencies are actually required
packages are now sorted. That was not a technical requirement.

* added `symfony/config`, this package is needed by the dependency
  injection extension

* added `symfony/http-kernel` which provides the base `Bundle` class

* added `php-http/httplug` and `psr/http-message` which provide
  interfaces the bundle relies on to be present

* moved `symfony/framework-bundle` to the `require-dev` section, it is
  only needed in tests where the `WebTestCase` class is used for
  functional tests

* added `guzzlehttp/psr7`, `php-http/promise`, `symfony/stopwatch`, and
  `symfony/http-foundation` as dev dependencies

* removed `phpunit/php-token-stream`, `symfony/finder`, and
  `symfony/twig-bridge` from `require-dev`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,40 +17,45 @@
     ],
     "require": {
         "php": "^5.5 || ^7.0",
-        "php-http/client-implementation": "^1.0",
-        "php-http/message-factory": "^1.0.2",
         "php-http/client-common": "^1.6",
+        "php-http/client-implementation": "^1.0",
         "php-http/cache-plugin": "^1.4",
-        "php-http/logger-plugin": "^1.0",
-        "php-http/stopwatch-plugin": "^1.0",
-        "symfony/options-resolver": "^2.8 || ^3.0 || ^4.0",
-        "symfony/event-dispatcher": "^2.8 || ^3.0 || ^4.0",
-        "symfony/framework-bundle": "^2.8 || ^3.0 || ^4.0",
-        "php-http/message": "^1.4",
         "php-http/discovery": "^1.0",
-        "twig/twig": "^1.18 || ^2.0",
+        "php-http/httplug": "^1.0",
+        "php-http/logger-plugin": "^1.0",
+        "php-http/message": "^1.4",
+        "php-http/message-factory": "^1.0.2",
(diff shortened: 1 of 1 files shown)


Nyholm:
@fbourigault, If I recall correctly you added the symfony/asset because it failed for you when installing the bundle with flex a while back... 
I thought that the twig function `asset` lived there. But it lives in the Symfony Twig Bridge. 

Travis seams to imply that we do need `symfony/framework-bundle` for tests.

xabbuh:
I still have to check tests and the dependencies listed under `require-dev`. I'll do that later today.

dbu:
thanks for this, the sort is particularly useful. i think i used framework bundle as dependency in all my symfony bundles, assuming that thats what you need to do a bundle... looking at the travis failure, the WebTestCase is part of framework bundle, and we use that in the tests, so its at least a require-dev.

fbourigault:
`symfony/asset` is required for the profiler.

xabbuh:
I added `symfony/asset` back and moved the FrameworkBundle to the `require-dev` section

composer.json
```diff
@@ -17,40 +17,45 @@
...
+        "php-http/react-adapter": "^0.2.1",
+        "php-http/socket-client": "^1.0",
+        "polishsymfonycommunity/symfony-mocker-container": "^1.0",
         "symfony/browser-kit": "^2.8 || ^3.0",
+        "symfony/cache": "^3.1",
         "symfony/dom-crawler": "^2.8 || ^3.0",
-        "polishsymfonycommunity/symfony-mocker-container": "^1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "^1.1 || ^2.0",
-        "nyholm/nsa": "^1.1"
+        "symfony/framework-bundle": "^2.8.1 || ^3.0.1",
```
xabbuh: the Finder component was only required before because of a bug in the FrameworkBundle's dependencies (see symfony/symfony#17057)

Nyholm:
Could you rebase this PR on master?

xabbuh:
rebased

PR #232, opened 2017-11-29
[WIP] Full symfony 4 dependencies

fbourigault:
This is a tentative to test against Symfony 4 only dependencies.

ping @Nyholm


fbourigault:
We need a new release of https://github.com/php-http/client-common!

.travis.yml
```diff
@@ -1,16 +1,13 @@
...
 sudo: false
-
 cache:
     directories:
         - $HOME/.composer/cache/files
         - $HOME/symfony-bridge/.phpunit
 
 env:
     global:
-        - TEST_COMMAND="composer test"
```
Nyholm: The php-http organisation has (maybe silently) decided that all our packages should be tested with `composer test` and `composer test-ci`. 

We do not need the `PHPUNIT_FLAGS`. You should revert this and continue using `TEST_COMMAND`

fbourigault: Ok. I pushed a better version IMHO. Let's see what's the build result.

fbourigault:
This is now ready to merge.

.travis.yml
```diff
@@ -21,36 +21,40 @@ matrix:
...
-        - php: 7.1
-          env: DEPENDENCIES="dev"
+        - php: 7.2
+          env: STABILITY="dev"
 
     allow_failures:
           # dev-master is allowed to fail.
-        - php: 7.1
-          env: DEPENDENCIES="dev"
+        - php: 7.2
```
dbu: according to the symfony docs PR, you can drop this and just keep the bit about env to allow failures

fbourigault: You mean:
```yaml
allow_failures:
    env: DEPENDENCIES="dev"
```
?

dbu: yes. see https://github.com/symfony/symfony-docs/pull/8701/files#diff-9d0d9156660f746e55727b52fcbf918dR214 - it will be good to see here if that works as expected ;-)

fbourigault: We will not be able to know this with this PR as its 100% green ;)

dbu: we can see it: when you click on the build, you see a travis page where the allowed to fail build are separated, even if they succeed.

the point of omitting the php version is that we don't care about the version and when php 7.3 comes out we would forget to adjust this line.

fbourigault:
🎉

PR #236, opened 2017-12-03
Removed "end year" form the license file

Nyholm:
This is not needed. You will automatically have copy right for what ever you create for 50(?) years.


dbu:
styleci can handle the header, then it would just be one place to update. but i am happy to remove it.

ianal, there seems to be reasons why some people do add the current year, but honestly i don't think its that big of a deal. its not like we are under GPL and are code that people would want to steal.

Nyholm:
Thank you for the reviews

PR #237, opened 2017-12-04
Share some data with a secret test tool by Nyholm

Nyholm:
Im currently developing a new tool. Could I start collecting some data from our builds?

It is no big secret but I've got nothing to show yet...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,5 +68,7 @@ script:
 after_success:
     - if [[ $COVERAGE = true ]]; then wget https://scrutinizer-ci.com/ocular.phar; fi
     - if [[ $COVERAGE = true ]]; then php ocular.phar code-coverage:upload --format=php-clover build/coverage.xml; fi
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -68,3 +68,5 @@ script:
 after_success:
     - if [[ $COVERAGE = true ]]; then wget https://scrutinizer-ci.com/ocular.phar; fi
     - if [[ $COVERAGE = true ]]; then php ocular.phar code-coverage:upload --format=php-clover build/coverage.xml; fi
+    - wget http://tnyholm.se/reporter.phar
+    - php reporter.phar build:upload
```
xabbuh: Will the build still report an error if these steps fail?

Nyholm: Yes. The build will get it status from the `script`. What we do afterwards is not important.

Nyholm:
Thank you for merging.

Issue #212, opened 2017-10-24
Document how to use mock with httplug


Soullivaneuh:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | no
| Version      | N/A

Httplug has a mock client for testing: http://docs.php-http.org/en/latest/clients/mock-client.html

This is great, but it would be better to be able to replace the current client service by mocks on test environment.

I suppose it's possible, but we don't have any "official" way described on the doc.

It would be a great addition IMHO.

Soullivaneuh:
A quick working implement:

```yaml
# config_test.yml

services:
  httplug.client.mock:
    class: Http\Mock\Client
  httplug.client.default: '@httplug.client.mock'
```

And then, the services using the default client is using the mock.

Plus, I can access the mock client to check the requests. Example on a `WebTestCase`:

```php
$client->getContainer()->get('httplug.client.mock')->getRequests()
```

This might be documented, or event better, we might propose a configuration option to replace some or all the clients by mocks.

WDYT?

Soullivaneuh:
Well, my first weird issue is here. With this test:

```
public function testSettingsBadAccessToken(): void
{
    /** @var Repository $repository */
    $repository = $this->getReference('repository_github_enabled');
    $client = $this->makeClient(true);
    $crawler = $client->request('GET', $this->getUrl('app_repository_show', [
        'id' => $repository->getId(),
    ]));
    $form = $crawler->selectButton('repository_settings[submit]')->form([
        'repository_settings[accessToken]' => 'not-valid-token',
    ]);
    $httplugMockClient = $client->getContainer()->get('httplug.client.mock');
    // The token will be tested for admin rights. We should return a 404 error here.
    $httplugMockClient->addResponse(new JsonResponse([
        'message' => 'Not Found',
    ], 404));
    dump($httplugMockClient);
    $client->submit($form);
    $this->assertStatusCode(200, $client);
    $this->assertValidationErrors([
        'data.accessToken',
    ], $client->getContainer());
}
```

And this dump on the concerned controller:

```
```

Soullivaneuh:
Well, my first weird issue is here. With this test:

```php
public function testSettingsBadAccessToken(): void
{
    /** @var Repository $repository */
    $repository = $this->getReference('repository_github_enabled');
    $client = $this->makeClient(true);
    $crawler = $client->request('GET', $this->getUrl('app_repository_show', [
        'id' => $repository->getId(),
    ]));
[...]

PR #242, opened 2018-01-03
Add mock client factory for test purpose.

GaryPEGEOT:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes partially #212 
| Documentation   | https://github.com/php-http/documentation/pull/220
| License         | MIT

#### What's in this PR?

Add a mock client factory to ease functional testing.

#### Example Usage
Configuration:
```yaml
    # config_test.yml
    httplug:
        clients:
            my_awesome_client:
                factory: 'httplug.factory.mock' # replace factory
```
Usage:
``` php
$client = static::createClient();
// $client->disableReboot(); You might uncomment this if your client (BrowserKit) make multiple requests as kernel is rebooted on each request.

$response = $this->createMock('Psr\Http\Message\ResponseInterface');
$response->method('getBody')->willReturn(/* Psr\Http\Message\Interface instance containing expected response content. */);
$client->getContainer()->get('httplug.client.mock')->addResponse($response);
```

### DIFF
--- a/ClientFactory/MockFactory.php
+++ b/ClientFactory/MockFactory.php
@@ -27,10 +27,14 @@ class MockFactory implements ClientFactory
      */
     public function createClient(array $config = [])
     {
-        if (!class_exists('Http\Mock\Client')) {
+        if (!class_exists(Client::class)) {
             throw new \LogicException('To use the mock adapter you need to install the "php-http/mock-client" package.');
         }
 
-        return $this->client ?: new Client();
+        if (!$this->client) {
+            $this->client = new Client();
+        }
+
+        return $this->client;
     }
 }
--- a/ClientFactory/MockFactory.php
+++ b/ClientFactory/MockFactory.php
@@ -5,8 +5,6 @@ namespace Http\HttplugBundle\ClientFactory;
 use Http\Mock\Client;
 
 /**
- * Class Http\HttplugBundle\ClientFactory\MockFactory
- *
  * @author Gary PEGEOT <garypegeot@gmail.com>
  */
 class MockFactory implements ClientFactory
--- a/ClientFactory/MockFactory.php
+++ b/ClientFactory/MockFactory.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Http\HttplugBundle\ClientFactory;
+
+use Http\Mock\Client;
+
+/**
+ * Class Http\HttplugBundle\ClientFactory\MockFactory
+ *
+ * @author Gary PEGEOT <garypegeot@gmail.com>
+ */
+class MockFactory implements ClientFactory
+{
+    /**
+     * @var Client
+     */
+    private $client;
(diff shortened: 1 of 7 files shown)


ClientFactory/MockFactory.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $this->client = $client;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createClient(array $config = [])
+    {
+        if (!class_exists('Http\Mock\Client')) {
```
dbu: please use the ::class pseudoconstant. we are on php 5.5 or newer.

DependencyInjection/HttplugExtension.php
```diff
@@ -106,6 +106,16 @@ private function configureClients(ContainerBuilder $container, array $config)
                 $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
             }
         }
+
+        $mockClass = 'Http\Mock\Client';
```
dbu: lets use the ::class pseudoconstant here. and make the use statement say `use Http\Mock\Client as MockClient;` to make things more readable.

ClientFactory/MockFactory.php
```diff
@@ -0,0 +1,36 @@
...
+            throw new \LogicException('To use the mock adapter you need to install the "php-http/mock-client" package.');
+        }
+
+        return $this->client ?: new Client();
```
dbu: this will create a new client on each call, if none has been registered. should we instead register the newly created client if none was set?

Resources/config/services.xml
```diff
@@ -68,5 +68,6 @@
         <service id="httplug.factory.socket" class="Http\HttplugBundle\ClientFactory\SocketFactory" public="false">
             <argument type="service" id="httplug.message_factory"/>
         </service>
+        <service id="httplug.factory.mock" class="Http\HttplugBundle\ClientFactory\MockFactory" public="false" />
```
dbu: i think this service should be removed from the container when the mock client is not available. that way, users notice problems at container build time instead of at runtime.

DependencyInjection/HttplugExtension.php
```diff
@@ -106,6 +107,19 @@ private function configureClients(ContainerBuilder $container, array $config)
                 $container->setAlias('httplug.client.default', 'httplug.client.'.$first);
             }
         }
+
+        $mockServiceId = 'httplug.client.mock';
+
+        if (!\class_exists(MockClient::class)) {
+            $container->removeDefinition('httplug.factory.mock');
```
fbourigault: Instead of removing a definition when the class does not exist, could you move the relevant service definition to a dedicated file and load this file if the class is available?

PR #247, opened 2018-01-08
Don't remove abstract plugins

### LINKED ISSUE #244: Cannot define cache pool on per-client basis without configuring dummy cache plugin
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes

I was happy to learn I can configure cache plugins on per-client basis!
However, it makes me configure default `cache` plugin anyway. This doesn't work:

```yaml
httplug:
  clients:
    app:
      plugins:
        - cache:
          cache_pool: 'app.cache.google'

```

It will result in error: `Service "httplug.client.app.plugin.cache": Parent definition "httplug.plugin.cache" does not exist.`

Ok let's add it:
```diff
httplug:
        cache: ~
```
Doesn't work either: `The child node "cache_pool" at path "httplug.plugins.cache" must be configured.`

At this point I have to configure dummy pool
```diff
-       cache: ~
+       cache:
+         cache_pool: 'app.cache.void'
```

Only now it works. What I am asking is to not force to configure default cache plugin.

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #244 
| License         | MIT

These abstract definitions are needed for client plugin definitions, even if plugins weren't configured at root level.

Allows to configure plugin this way

``` yaml
httplug:
    clients:
        test:
            plugins:
                - cache:
                    cache_pool: my_cache_pool

```
instead of forcing unneeded configuration at httplug.plugins.cache level. See linked issue.

### DIFF
--- a/Tests/Unit/DependencyInjection/ConfigurationTest.php
+++ b/Tests/Unit/DependencyInjection/ConfigurationTest.php
@@ -355,4 +355,24 @@ class ConfigurationTest extends AbstractExtensionConfigurationTestCase
         $file = __DIR__.'/../../Resources/Fixtures/config/bc/profiling_toolbar.yml';
         $this->assertProcessedConfigurationEquals([], [$file]);
     }
+
+    /**
+     * @expectedException \Symfony\Component\Config\Definition\Exception\InvalidConfigurationException
+     * @expectedExceptionMessage The child node "cache_pool" at path "httplug.clients.test.plugins.0.cache" must be configured.
+     */
+    public function testClientCacheConfigMustHavePool()
+    {
+        $file = __DIR__.'/../../Resources/Fixtures/config/client_cache_config_with_no_pool.yml';
+        $this->assertProcessedConfigurationEquals([], [$file]);
+    }
+
+    /**
+     * @expectedException \Symfony\Component\Config\Definition\Exception\InvalidConfigurationException
+     * @expectedExceptionMessage The child node "cache_pool" at path "httplug.plugins.cache" must be configured.
+     */
+    public function testCacheConfigMustHavePool()
+    {
+        $file = __DIR__.'/../../Resources/Fixtures/config/cache_config_with_no_pool.yml';
+        $this->assertProcessedConfigurationEquals([], [$file]);
+    }
 }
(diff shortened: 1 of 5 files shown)


Tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -355,4 +355,24 @@ public function testProfilingToolbarCollision()
         $file = __DIR__.'/../../Resources/Fixtures/config/bc/profiling_toolbar.yml';
         $this->assertProcessedConfigurationEquals([], [$file]);
     }
+
+    /**
+     * @expectedException \Symfony\Component\Config\Definition\Exception\InvalidConfigurationException
+     * @expectedExceptionMessage The child node "cache_pool" at path "httplug.clients.test.plugins.0.cache" must be configured.
```
dbu: please use $this->expectException and $this->expectExceptionMessage instead of the annotations. the annotations have been deprecated in favor of the method. they have the upsides of being able to use the ::class constant and (not relevant in this case here) of only accepting the exception at the place you actually expect it, not before that.

Tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -355,4 +355,24 @@ public function testProfilingToolbarCollision()
...
+     */
+    public function testClientCacheConfigMustHavePool()
+    {
+        $file = __DIR__.'/../../Resources/Fixtures/config/client_cache_config_with_no_pool.yml';
+        $this->assertProcessedConfigurationEquals([], [$file]);
+    }
+
+    /**
+     * @expectedException \Symfony\Component\Config\Definition\Exception\InvalidConfigurationException
+     * @expectedExceptionMessage The child node "cache_pool" at path "httplug.plugins.cache" must be configured.
```
dbu: same here

Tests/Unit/DependencyInjection/ConfigurationTest.php
ostrolucky: There is no instance in this code base where expectExceptionMessage is used. I've followed style of existing code. 

Furthermore, expectException is PHPUnit 6+ only.

If you decided to change the code style, please do it yourself with refactoring commit which replaces it in all places at once.

dbu: sorry, my bad. i am working in too many repositories that are at different states of upgrading those things... indeed, its better to be consistent with this. i created #248

dbu:
I think this is ready to merge. @Nyholm @fbourigault do you agree with this change?

fbourigault:
Thank you!

PR #252, opened 2018-02-04
Empty configuration should not contain any data.

Nyholm:
This will fix our broken master. 

Related to https://github.com/symfony/symfony/pull/25891

### DIFF
--- a/Tests/Resources/Fixtures/config/empty.yml
+++ b/Tests/Resources/Fixtures/config/empty.yml
@@ -1,1 +1,0 @@
-httplug:


dbu:
thanks!

xabbuh:
@Nyholm Can you explain what this fixture was meant to test? Would it maybe be worth to adapt the `Configuration` class to maintain the previous behaviour?

Nyholm:
This test is suppose to test if heaven falls down if you do not write any configuration. Im all for modifying `Configuration` but I was not sure how to do it

PR #254, opened 2018-02-20
Make main aliases public

stof:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

Symfony 3.4+ is making services private by default, so this requires configuring the visibility explicitly.

#### Why?

This allows keeping `$this->get('httplug.client')` working without deprecation.

For now, I haven't applied this change to the different client services, but it might make sense as well.

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -15,6 +15,7 @@ use Http\Message\Authentication\Wsse;
 use Http\Mock\Client as MockClient;
 use Psr\Http\Message\UriInterface;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\Alias;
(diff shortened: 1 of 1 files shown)


stof:
what about non-default clients ?

dbu:
the *should* be injected in DI. but i am not opposed to make the client services public.

stof:
well, if you consider that using DI for other clients is better, that's fine with me (I don't have a use case for that need anyway as I have only 1 client currently)

dbu:
i guess if you have multiple clients configured, you would want to inject them and not pull. if there is a use case for it, i am not opposed to a pull request however.

PR #267, opened 2018-06-27
Add QueryParam auth plugin support

### LINKED ISSUE #265: QueryParams auth not supported
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.10.0

#### Actual Behavior

This configuration:

```yaml
httplug:
    plugins:
        cache:
            # Should be the auto-wired service.
            # @see https://github.com/php-http/HttplugBundle/issues/263
            cache_pool: 'cache.app'
    clients:
        pagespeed:
            config:
                timeout: 5
            http_methods_client: true
            plugins:
                -
                    authentication:
                        pagespeed:
                            type: query_params
                            key: 'test'
                - 'httplug.plugin.cache'
```

Does not work.

#### Expected Behavior

The [`QueryParams`](http://docs.php-http.org/en/latest/message/authentication.html#query-params) should be managed.

Soullivaneuh:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #265 
| Documentation | https://github.com/php-http/documentation/pull/238
| License         | MIT

#### What's in this PR?

Add missing support for the `QueryParam` auth plugin.

#### Why?

Because there is no reason to not have it. :-D

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -522,6 +522,10 @@ class Configuration implements ConfigurationInterface
                                 $this->validateAuthenticationType(['username', 'password'], $config, 'wsse');
 
                                 break;
+                            case 'query_param':
+                                $this->validateAuthenticationType(['params'], $config, 'query_param');
+
+                                break;
                         }
 
                         return $config;
@@ -529,7 +533,7 @@ class Configuration implements ConfigurationInterface
                 ->end()
                 ->children()
                     ->enumNode('type')
-                        ->values(['basic', 'bearer', 'wsse', 'service'])
+                        ->values(['basic', 'bearer', 'wsse', 'service', 'query_param'])
                         ->isRequired()
                         ->cannotBeEmpty()
                     ->end()
@@ -537,6 +541,7 @@ class Configuration implements ConfigurationInterface
                     ->scalarNode('password')->end()
                     ->scalarNode('token')->end()
                     ->scalarNode('service')->end()
+                    ->arrayNode('params')->prototype('scalar')->end()
                     ->end()
                 ->end()
             ->end(); // End authentication plugin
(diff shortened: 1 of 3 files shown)


Soullivaneuh:
@Nyholm I have an issue with the always empty `params` array.

I'm investigating, but if you have a clue to give... ;-)

Nyholm:
> I have an issue with the always empty params array.

Which line?

Soullivaneuh:
@Nyholm https://github.com/php-http/HttplugBundle/pull/267/files#diff-850942b3ba24ab03a40aaa81b6152852R544

In a nutshell, `params` is always present as an empty array event if not filled. But `validateAuthenticationType` does not except `params` unless for `query_params`.

But I'm not sure we can remove it from the config if the array is empty. :thinking:

Nyholm:
Be more pragmatic =)

Write a special case for `query_params` in `validateAuthenticationType`.

Soullivaneuh:
@Nyholm The issue is `params` is **always** present as an empty array: https://github.com/php-http/HttplugBundle/pull/267/files#diff-850942b3ba24ab03a40aaa81b6152852R544

So yes, I can just unset `params` if the value is empty on `validateAuthenticationType`, but I consider that quite ugly and `params` will always be present on the final configuration.

Or maybe the idea you have is different?

Soullivaneuh:
> the risk with query parameters for credentials is that they show up in all kind of access logs from the server

I totally agree! But the targeted API maintainer does not seems to be agree. BTW, the maintainer is Google, f*ck yeah! :trollface: 

They is a risk, but I don't think we should log the emitted request on production.

I'll take a look for the documentation and the rebase.

Soullivaneuh:
Is anything missing to get this merged and released?

Nyholm:
Thank you for the ping. 
You are all good. Thank you for the work!

PR #277, opened 2018-09-23
remove abandoned styleci bridge

### LINKED ISSUE #266: PHP-CS-Fixer not working anymore
It seems you remove my abandoned styleci-config-bridge package. But the `.php_cs` file still use it.

Would you like me to configure [FlintCI](https://flintci.io/) service? It will allow you to have your common configuration on `.php_cs` only, this service will only run it without any extra configuration except the activation itself.

Regards.

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #266
| Documentation   | (not mentioned in the doc)
| License         | MIT

#### What's in this PR?

Remove configuration for the styleci php-cs-fixer bridge

#### Why?

The bridge is abandoned. styleci seems to not offer anything to run codestyle locally with the same configuration as their service.
Imho its enough to have the automated check.

#### Checklist


dbu:
if we agree that this is the correct move, i will remove the file from client-common, boilerplate, curl-client, plugins, message, discovery, guzzle6-adapter, mock-client, chache-client and react-adapter.

sagikazarmark:
I agree as well. PrettyCI looks cool, I used FlintCI before, which accepted plain phpcs config, not sure if PrettyCI does.

Can you check if you can disable StyleCI? Do I need to grant some privileges?

mnapoli:
Hello there! Just answering the question and I'll be out of here: if you want to use it then yes you can use a custom phpcs config, see https://prettyci.com/documentation

dbu:
see #278

dbu:
replaced by #278

PR #280, opened 2018-10-24
Make it possible to create public httplug clients

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      |no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Added `public: true/false` to the Configuration. It is used to make the created HTTPlug client public or private.

#### Why?

All HTTPlug clients that are created are private by default, which is a good thing. But sometimes, you just need a public HTTPlug client.

#### Checklist


Nyholm:
Hm. I agree to this PR. We discussed it before and we said that the user should always want a private service since that is the best practice. 

However, I know people tried to access them as they were public.. I don’t know if we ever talked about making it configurable, but I believe that would be awesome for all users. 

:+1:

DependencyInjection/Configuration.php
```diff
@@ -195,6 +195,10 @@ private function configureClients(ArrayNodeDefinition $root)
                         ->defaultNull()
                         ->info('The service id of the client to use.')
                     ->end()
+                    ->booleanNode('public')
+                        ->defaultFalse()
+                        ->info('Set true if you really cannot use dependency injection and need to make the client service public.')
```
xabbuh: Set to true if [...]

DependencyInjection/HttplugExtension.php
```diff
@@ -316,15 +316,15 @@ private function configureClient(ContainerBuilder $container, $clientName, array
             $container
                 ->register($serviceId.'.client', HttpClient::class)
                 ->setFactory([new Reference($arguments['factory']), 'createClient'])
-                ->addArgument($arguments['config'])
-                ->setPublic(false);
```
xabbuh: Shouldn't this definition still be private?

DependencyInjection/Configuration.php
ruudk: Fixed

DependencyInjection/HttplugExtension.php
ruudk: Sharp, undid this change.

Tests/Unit/DependencyInjection/HttplugExtensionTest.php
```diff
@@ -267,4 +267,30 @@ private function verifyProfilingDisabled()
             );
         }
     }
+
+    public function testClientShouldBePrivatByDefault()
```
xabbuh: Private

ruudk: Fixed

ruudk:
I made some changes to the test to check the Symfony version. I made the `public` configuration option `null` by default. That way, we know if the user actually tried to make the service public or not. It's always used in the decorated services (since they are private by default) and only used when it's a bool for the client.

Tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -119,6 +119,7 @@ public function testSupportsAllConfigFormats()
                     'factory' => 'httplug.factory.guzzle6',
                     'http_methods_client' => true,
                     'service' => null,
+                    'public' => false,
```
dbu: this should now be `null`, right?

Tests/Unit/DependencyInjection/HttplugExtensionTest.php
```diff
@@ -267,4 +268,124 @@ private function verifyProfilingDisabled()
             );
         }
     }
+
+    public function testClientShouldBePrivateByDefault()
```
dbu: can we call this `testClientShouldHaveDefaultVisibility()` '

ruudk: Changed

Tests/Unit/DependencyInjection/ConfigurationTest.php
ruudk: Changed it... Somehow it didn't fail when it was either `null` or `false`.

ruudk:
@xabbuh Tests are now fixed and green ✅ 🤗

Tests/Unit/DependencyInjection/HttplugExtensionTest.php
```diff
@@ -267,4 +268,145 @@ private function verifyProfilingDisabled()
...
+    {
+        $this->load([
+            'clients' => [
+                'acme' => [],
+            ],
+        ]);
+
+        $this->assertContainerBuilderHasService('httplug.client.acme');
+
+        if (version_compare(Kernel::VERSION, '3.4', '>=')) {
```
xabbuh: The version of the HttpKernel component does not necessarily have to be the same as the DependencyInjection component. If we want to fully be on the safe side, we could for example check for the `EnvVarProcessor` class to exist:

```php
use Symfony\Component\DependencyInjection\EnvVarProcessor;

// ...

if (class_exists(EnvVarProcessor::class)) {
    // ...
}
```

ruudk:
Thanks for merging @dbu. When will it get released?

dbu:
i am updating the changelog in #281, once that is good we can tag the release

dbu:
done: https://github.com/php-http/HttplugBundle/releases/tag/1.12.0

PR #283, opened 2018-10-31
use php 7.3 for special tests

dbu:
also use php 7.3 for test runs with special environment settings.

follow up to #282

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
     fast_finish: true
     include:
           # Minimum supported Symfony version and latest PHP version
-        - php: 7.2
+        - php: 7.3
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -36,15 +36,15 @@ matrix:
           env: COVERAGE=true TEST_COMMAND="composer test-ci"
 
           # Test LTS versions
-        - php: 7.2
+        - php: 7.3
           env: DEPENDENCIES="dunglas/symfony-lock:^2"
-        - php: 7.2
+        - php: 7.3
           env: DEPENDENCIES="dunglas/symfony-lock:^3"
-        - php: 7.2
```
Nyholm: Is this really needed? I know travis is quicker to start on 7.2.. Ah, anyhow... 

It is not needed but why not =)

dbu: we can also wait with merging this until 7.3 is released as final, rather than RC. i guess then travis will also be quick with it.

PR #287, opened 2018-11-19
Make underlying http client also public when configured

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | follow up of https://github.com/php-http/HttplugBundle/pull/280
 Documentation   | not needed
| License         | MIT

#### What's in this PR?

This makes the underlying http client also public when configured. 

#### Why?

Something I didn't add in #280 but need when using mocks. This is needed for https://github.com/php-http/HttplugBundle/pull/286.

#### Checklist

- [x] Updated CHANGELOG.md
- [ ] Documentation: no need to update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## Unreleased
+
+### Added
+
+- Make underlying http client (`httplug.client.acme.client`) also public when configured
+
 ## 1.13.0 - 2018-11-13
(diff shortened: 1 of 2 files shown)


dbu:
i closed #286 because i don't see the use case. similar here, i don't see the use in having this internal service be public. if you really need it to be public for a special test setup, i suggest that you add a compiler pass in your application that finds the corresponding services and changes them to public after this bundle has defined them as private.

PR #288, opened 2018-11-28
Fix wrong duration calculation for asynchronous calls

oliverde8:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #X, partially #Y, mentioned in #Z
| License         | MIT

#### What's in this PR?

When using asnyc request, duration is always 0 

This fixes that

#### Why?

Currenly for async request we don't see any duration. 
httpplug-async-bug-times-1

First query is synchronous, the other 2 are asynchronous, and are slow. This fixes that

httpplug-async-bug-times-2

Now we can see the duration of asynchronous calls as well

#### Checklist


dbu:
released as https://github.com/php-http/HttplugBundle/releases/tag/1.13.1

PR #291, opened 2018-12-13
support client-common 2.0

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/php-http/client-common/pull/123
| Documentation   | -
| License         | MIT

#### What's in this PR?

Handle version 2 of client-commons with a check for the Impl classes.

#### Why?

Those classes where moved to make room for the interface.

#### Checklist

#### To Do

Should we bump the version of the bundle as well? There are some BC breaks and people might rely on the bundle to bring in the whole httplug libraries...


Nyholm:
> Should we bump the version of the bundle as well?

I dont think so.. I rather make the number of libraries to bump as small as possible". 

Of course we should bump it if there are any strong reason for it

dbu:
@Nyholm could you please also review https://github.com/php-http/client-common/pull/123 ?

dbu:
build failure is the same as for [master](https://travis-ci.org/php-http/HttplugBundle/builds/468665919). some problem with phpunit component versions afaik.

dbu:
and we should not need a new major version for this, there are no code changes needed.

gmponos:
Current PR brings up an issue... If you keep supporting both versions of HTTP-Client you can not depend on the PSR. Is this the direction you want this to go?

Nyholm:
True. But we are not depending on the psr. Are we?

gmponos:
OK.. I should started this conversation here.. let's continue on the PR of yours here https://github.com/php-http/HttplugBundle/pull/295

PR #293, opened 2018-12-28
Support common client 2.x

Nyholm:
Follow up on #291 

This uses the forward compatibility layer introduced in php-http/client-common#132


Nyholm:
I've tested this and it is working. 

I also added
```
     "prefer-stable": true,
     "minimum-stability": "dev"
```

Which means that we dont need to wait for a release of 1.9 of client-common to actually merge this.

Nyholm:
We allow too many versions. That is why Travis is failing: https://travis-ci.org/php-http/HttplugBundle/jobs/473334912

I updated all symfony related versions. We only support the latest patch version for all supported minor versions.

sowbiba:
I cannot install latest version of httplug-bundle because

```
  Problem 1
    - Installation request for php-http/httplug-bundle 1.14.0 -> satisfiable by php-http/httplug-bundle[1.14.0].
    - php-http/httplug-bundle 1.14.0 requires php-http/cache-plugin ^1.6 -> satisfiable by php-http/cache-plugin[1.6.x-dev] but these conflict with your requirements or minimum-stability.
```

client-common 1.9 is released, will cache-plugin be released soon ? :pray: 

Thanks

PR #294, opened 2018-12-29
Support for more clients versions

Nyholm:

PR #295, opened 2018-12-29
Make sure we support any psr-18 client

### LINKED ISSUE #276: Make sure we can use this bundle with ANY PSR-18 client
Adding this issue as a reminder. Im not sure what changes that needs to be done atm.

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #276
| Documentation   | 
| License         | MIT

### DIFF
--- a/Collector/Formatter.php
+++ b/Collector/Formatter.php
@@ -7,6 +7,8 @@ use Http\Client\Exception\HttpException;
 use Http\Client\Exception\TransferException;
 use Http\Message\Formatter as MessageFormatter;
 use Http\Message\Formatter\CurlCommandFormatter;
+use Psr\Http\Client\NetworkExceptionInterface;
+use Psr\Http\Client\RequestExceptionInterface;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
@@ -49,11 +51,11 @@ class Formatter implements MessageFormatter
      */
     public function formatException(Exception $exception)
     {
-        if ($exception instanceof HttpException || $exception instanceof \Psr\Http\Client\RequestExceptionInterface) {
+        if ($exception instanceof HttpException || $exception instanceof RequestExceptionInterface) {
             return $this->formatter->formatResponse($exception->getResponse());
         }
 
-        if ($exception instanceof TransferException || $exception instanceof \Psr\Http\Client\NetworkExceptionInterface) {
+        if ($exception instanceof TransferException || $exception instanceof NetworkExceptionInterface) {
             return sprintf('Transfer error: %s', $exception->getMessage());
         }
 
--- a/Collector/Formatter.php
+++ b/Collector/Formatter.php
@@ -49,11 +49,11 @@ class Formatter implements MessageFormatter
      */
     public function formatException(Exception $exception)
     {
-        if ($exception instanceof HttpException) {
+        if ($exception instanceof HttpException || $exception instanceof \Psr\Http\Client\RequestExceptionInterface) {
             return $this->formatter->formatResponse($exception->getResponse());
         }
 
-        if ($exception instanceof TransferException) {
+        if ($exception instanceof TransferException || $exception instanceof \Psr\Http\Client\NetworkExceptionInterface) {
             return sprintf('Transfer error: %s', $exception->getMessage());
         }
 
--- a/Collector/PluginClientFactory.php
+++ b/Collector/PluginClientFactory.php
@@ -6,6 +6,7 @@ use Http\Client\Common\Plugin;
 use Http\Client\Common\PluginClient;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
+use Http\HttplugBundle\Tests\Unit\Collector\ClientInterface;
 use Symfony\Component\Stopwatch\Stopwatch;
 
 /**
@@ -46,9 +47,9 @@ final class PluginClientFactory
     }
 
     /**
-     * @param HttpClient|HttpAsyncClient $client
-     * @param Plugin[]                   $plugins
-     * @param array                      $options {
+     * @param HttpClient|ClientInterface|HttpAsyncClient $client
+     * @param Plugin[]                                   $plugins
+     * @param array                                      $options {
      *
      *     @var string $client_name to give client a name which may be used when displaying client information like in
      *         the HTTPlugBundle profiler.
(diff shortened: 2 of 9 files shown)


Collector/PluginClientFactory.php
```diff
@@ -6,6 +6,7 @@
 use Http\Client\Common\PluginClient;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
+use Http\HttplugBundle\Tests\Unit\Collector\ClientInterface;
```
gmponos: Wrong interface?

Nyholm: Oops. Thanks

Collector/ProfileClient.php
```diff
@@ -7,6 +7,7 @@
 use Http\Client\Exception\HttpException;
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
+use Http\HttplugBundle\Tests\Unit\Collector\ClientInterface;
```
Nyholm: Same here.

Collector/ProfileClientFactory.php
```diff
@@ -6,6 +6,7 @@
 use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Http\HttplugBundle\ClientFactory\ClientFactory;
+use Psr\Http\Client\ClientInterface;
```
gmponos: I have checked out at your branch. I run `composer update` but still this namespace is not discovered by PHPStorm.. and the package is not installed.

Could it be that exists localy on you because you have not run `composer update`?

Nyholm: You need to make sure you install php-http/client-common:2.0-dev. If not you will get httplug: 1.1

Nyholm: Oh, you are correct.

Nyholm:
Oh, you are correct. Since we have a bunch if httplug1.0 clients in require-dev we will never be able to install httplug2.0 by just running composer update. 

I tested this by

```
composer create-project symfony/skeleton acme
cd acme
composer req php-http/httplug-bundle:dev-master nyholm/psr7
```

I created my client like:
```php
namespace App\Service;

use Nyholm\Psr7\Response;
use Psr\Http\Client\ClientInterface;
use Psr\Http\Message\ResponseInterface;

class MyHttpClient implements ClientInterface
{
    public function sendRequest(RequestInterface $request): ResponseInterface
    {
        return new Response(250);
    }
}
```

And I registered it like: 
```yaml
httplug:
    main_alias:
        client: httplug.client.my_client
    discovery:
        client: 'auto'

    clients:
        my_client:
            service: 'App\Service\MyHttpClient'

            http_methods_client: true
            plugins:
                - 'httplug.plugin.content_length'
                - 'httplug.plugin.redirect'

```

gmponos:
> > Current PR brings up an issue... If you keep supporting both versions of HTTP-Client you can not depend on the PSR. Is this the direction you want this to go?

> True. But we are not depending on the psr. Are we?


Well.. on the current a PR you want to do this `if( $client instanceof ClientInterface)` although this code works.. even if I don't have version 2 installed it kinda feels weird... 

Secondly as a developer of this package... yes I do not depend on PSR client and I might not want to depend... As a consumer using this package.. is there a way that I feel confident that the client I am retrieving is a PSR one? Haven't discovered all the full aspects of the bundle so on this point I might be wrong...

Thirdly... [here](https://github.com/php-http/HttplugBundle/blob/master/Collector/Formatter.php#L52) wont this miss the exceptions?

gmponos:
@Nyholm I think that your example works out of luck.. although I know that this was by design it just happens that PSR client and PHP-HTTP to have the same signature `sendRequest`.. 

for instance the `ProfileClient` I think it will work even a custom client of mine that does not implement any interface and it just has the `sendRequest`.. it feels kinda loose in this way...

Nyholm:
> Well.. on the current a PR you want to do this if( $client instanceof ClientInterface) although this code works.. even if I don't have version 2 installed it kinda feels weird...

Yes, I agree. But it works =)

> Secondly as a developer of this package... yes I do not depend on PSR client and I might not want to depend... As a consumer using this package.. is there a way that I feel confident that the client I am retrieving is a PSR one? Haven't discovered all the full aspects of the bundle so on this point I might be wrong... I think there is some autodiscovery mechanism.. that's what I am talking about.

The auto discovery is here: https://github.com/php-http/discovery

And no, you cannot be sure unless you decide what you want. If you do `composer req php-http/httplug:^2.0`, then you will always get a PSR18 client. Same if you do `composer req psr/http-client`

> Thirdly... here wont this miss the exceptions?

That should be updated. Thank you!

> although I know that this was by design it just happens that PSR client and PHP-HTTP to have the same signature sendRequest..
> for instance the ProfileClient I think it will work even a custom client of mine that does not implement any interface and it just has the sendRequest.. it feels kinda loose in this way...

Technically, yes. But we have code that make sure you submit a client. Check out the [ProfileClient](https://github.com/php-http/HttplugBundle/blob/master/Collector/ProfileClient.php#L61-L62) and the [FlexibleClient](https://github.com/php-http/client-common/blob/master/src/FlexibleHttpClient.php#L25)

gmponos:
>> for instance the ProfileClient I think it will work even a custom client
> But we have code that make sure you submit a client. 

OK.. I missed the `!` there...

I will have a deeper look because then I don't quite fully get it how your code example posted above worked with master :/

`composer req php-http/httplug-bundle:dev-master nyholm/psr7`

Collector/Formatter.php
```diff
@@ -49,11 +49,11 @@ public function __construct(MessageFormatter $formatter, CurlCommandFormatter $c
      */
     public function formatException(Exception $exception)
     {
-        if ($exception instanceof HttpException) {
+        if ($exception instanceof HttpException || $exception instanceof \Psr\Http\Client\RequestExceptionInterface) {
```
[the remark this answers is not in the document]
Nyholm: Sure, thanks

dbu:
is this issue fixed with #294? can you rebase and maybe add some tests that make sure both httplug 1 (when installed) and a psr-18 client (when installed) work with this code?

Nyholm:
Correct. 
Sure!

Nyholm:
Why do you believe that?

Nyholm:
Continue in #296

PR #299, opened 2019-01-02
Profiler dot client

### LINKED ISSUE #298: Profiler does not show stack when client name contains dots
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.13.1

#### Actual Behavior

When the client name contains dots (example : application.ms.microservice), the profiler stack cannot be opened

#### Expected Behavior

Open Request/Response stack

#### Steps to Reproduce
```
httplug:
    plugins:
        logger:
            logger: 'logger'
            formatter: null
    clients:
        application.ms.microservice: ~
``` 

#### Possible Solutions

This issue happened because `data-toggle` contains `#httplug-application.ms.microservice-1-details`

In https://github.com/php-http/HttplugBundle/blob/master/Resources/views/script/httplug.js.twig#L7

`document.querySelectorAll(selector)` the dots are considered as class selector and the target is not found

sowbiba:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #298 
| License         | MIT

#### What's in this PR?

Add SlugifyExtension to remove non standard characters from client name to use as selector
https://www.w3.org/TR/CSS21/syndata.html#characters

#### Why?

https://github.com/php-http/HttplugBundle/issues/298

#### Checklist

- [X] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
- [X] Documentation pull request created (if not simply a bugfix)


Collector/Twig/SlugifyExtension.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @return array
+     */
+    public function getFilters()
+    {
+        return [
+            new \Twig_SimpleFilter('slugify', [$this, 'slugify'], ['is_safe' => ['html']]),
```
dbu: i guess there are quite some twig filters out there that provide a function with the same name.

i agree to not introduce a dependency on some twig thingy, and the actual slugify looks simple enough that we don't need a library for that. but i think we need a more specific name for this to avoid collisions. maybe `httplug_slugify` ?

Collector/Twig/SlugifyExtension.php
```diff
@@ -0,0 +1,34 @@
...
+        ];
+    }
+
+    /**
+     * @param string $string http message
```
dbu: why http message? i think we can do without the phpdoc, or just `@param string $string` until we are on php 7 and can use scalar type hints.

Collector/Twig/HttpPlugSlugifyExtension.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        return [
+            new \Twig_SimpleFilter('httplug_slugify', [$this, 'slugify'], ['is_safe' => ['html']]),
+        ];
+    }
+
+    /**
+     * @param $string
+     *
+     * @return string|string[]|null
```
dbu: can this ever return anything else than a string?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## 1.14.1
```
Nyholm: FYI, 1.14.0 is not released yet.

sowbiba:
@Nyholm Thanks for the suggestions. Changes made

Nyholm:
Excellent. Thank you @sowbiba

Issue #301, opened 2019-01-15
Cannot run tests on master because trait `...VersionBridgePlugin` is not found


ruudk:
I'm trying to run `vendor/bin/simple-phpunit` on master, but it fails:

```
$ vendor/bin/simple-phpunit
symfony/yaml is not required in your composer.json and has not been removed
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies
Package operations: 23 installs, 0 updates, 0 removals
  - Installing phpunit/php-timer (1.0.9): Loading from cache
  - Installing sebastian/global-state (2.0.0): Loading from cache
  - Installing sebastian/recursion-context (3.0.0): Loading from cache
  - Installing sebastian/object-reflector (1.1.1): Loading from cache
  - Installing sebastian/object-enumerator (3.0.3): Loading from cache
  - Installing sebastian/resource-operations (1.0.0): Loading from cache
  - Installing phar-io/version (1.0.1): Loading from cache
  - Installing phar-io/manifest (1.0.1): Loading from cache
  - Installing myclabs/deep-copy (1.7.0): Loading from cache
  - Installing theseer/tokenizer (1.1.0): Loading from cache
  - Installing sebastian/version (2.0.1): Loading from cache
  - Installing sebastian/environment (3.1.0): Loading from cache
  - Installing sebastian/code-unit-reverse-lookup (1.0.1): Loading from cache
  - Installing phpunit/php-token-stream (2.0.2): Loading from cache
  - Installing phpunit/php-text-template (1.2.1): Loading from cache
  - Installing phpunit/php-file-iterator (1.4.5): Loading from cache
  - Installing phpunit/php-code-coverage (5.3.2): Loading from cache
  - Installing sebastian/exporter (3.1.0): Loading from cache
  - Installing doctrine/instantiator (1.0.5): Loading from cache
  - Installing phpunit/phpunit-mock-objects (5.0.10): Loading from cache
  - Installing sebastian/diff (2.0.1): Loading from cache
  - Installing sebastian/comparator (2.1.3): Loading from cache
  - Installing symfony/phpunit-bridge (6.5.99): Symlinking from /Volumes/CS/www/HttplugBundle/vendor/symfony/phpunit-bridge
Writing lock file
Generating optimized autoload files
PHP Fatal error:  Trait 'Http\Client\Common\Plugin\VersionBridgePlugin' not found in /Volumes/CS/www/HttplugBundle/Tests/Functional/ProfilingTest.php on line 125

[...]

PR #302, opened 2019-01-15
Only require Twig when profiler is enabled

### LINKED ISSUE #300: Remove Twig dependency
I just discovered that HTTPlugBundle requires Twig. I guess the only need for this is because of the HTTP profiler extension. Wouldn't it be an idea to remove the Twig dependency, and only register the HTTP profiler when Twig is enabled?

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #300
| License         | MIT

#### What's in this PR?

Require Twig to be installed when the profiler is enabled. Move `twig/twig` dependency to suggest and `require-dev`

#### Why?

We don't have Twig installed in our project. HTTPlug requires it to be installed.

#### Checklist

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -76,6 +76,10 @@ class HttplugExtension extends Extension
                 ->getDefinition('httplug.formatter.full_http_message')
                 ->addArgument($config['profiling']['captured_body_length'])
             ;
+
+            if (!class_exists(\Twig_Environment::class)) {
+                $container->removeDefinition('httplug.collector.twig.http_message');
+            }
         }
 
         $this->configureClients($container, $config);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+## 1.15.0 [unreleased]
+
+### Changed
+
+- Removed `twig/twig` dependency
+
 ## 1.14.0
 
 ### Added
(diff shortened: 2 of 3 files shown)


ruudk:
I cannot test this (yet) because of https://github.com/php-http/HttplugBundle/issues/301

xabbuh:
As written in #300 I do not think we should forbid using the profiler tools when Twig is not installed. Using them in tests does not require Twig if I do not miss anything.

ruudk:
@xabbuh It's not only the Twig extension that requires Twig, right? Also by registering the profiler button in the web profiler it requires Twig?

Shouldn't we remove these services when Twig is not installed?
```xml
<service id="httplug.collector.collector" class="Http\HttplugBundle\Collector\Collector" public="false">
            <tag name="data_collector" template="@Httplug/webprofiler.html.twig" priority="200" id="httplug"/>
        </service>

        <service id="httplug.plugin.stack" class="Http\HttplugBundle\Collector\StackPlugin" public="false" abstract="true">
            <argument type="service" id="httplug.collector.collector"/>
            <argument type="service" id="httplug.collector.formatter"/>
        </service>
```

ruudk:
I changed the PR. It now removes the Twig extension when Twig is not installed. 

I think it's impossible to use the Symfony Web Toolbar without Twig, right?

xabbuh:
The `template` attribute of the `data_collector` tag is used when the `data_collector.templates` parameter is populated. But this parameter is only evaluated inside the `ProfilerController` which the web debug toolbar and the profiler panels. This controller is part of the WebProfilerBundle which already has a dependency on Twig.

DependencyInjection/HttplugExtension.php
```diff
@@ -81,6 +81,10 @@ public function load(array $configs, ContainerBuilder $container)
         $this->configureClients($container, $config);
         $this->configurePlugins($container, $config['plugins']); // must be after clients, as clients.X.plugins might use plugins as templates that will be removed
         $this->configureAutoDiscoveryClients($container, $config);
+
+        if (!class_exists(\Twig_Environment::class)) {
+            $container->removeDefinition('httplug.collector.twig.http_message');
```
dbu: this should be inside the `if ($profilingEnabled) {` i think, as the service is defined in data-collector.xml.

dbu:
can you please include the change to composer.json in this PR as well, moving `twig/twig` to require-dev?

and can you please add a  changelog entry? please use "1.15.0 [unreleased]" as heading.

DependencyInjection/HttplugExtension.php
ruudk: Good one, changed.

ruudk:
@dbu I applied the suggested changes.

PR #309, opened 2019-01-24
support autowire convenience clients

### LINKED ISSUE #261: AutoWiring and HttpMethodsClient
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 1.10.0

With the following configuration:

```yaml
httplug:
    clients:
        default:
            config:
                timeout: 5
```

The bundle provide autowiring for the default http client:

```
  Http\Client\HttpAsyncClient                                                                     
      alias to httplug.async_client.default                                                       
  Http\Client\HttpClient                                                                          
      alias to httplug.client.default                                                             
```

It's also possible to enable the http methods client:

```yaml
httplug:
    clients:
        default:
            config:
                timeout: 5
            http_methods_client: true
```

But the autowiring still rely on `HttpClient` interface:

```
  Http\Client\HttpAsyncClient                                                                     
      alias to httplug.async_client.default                                                       
  Http\Client\HttpClient                                                                          
      alias to httplug.client.default.http_methods                                                
```

So, we of course have a `HttpMethodsClient`, but the following code sample does not rely on anything:

```php
public function faviconRedirectAction(string $domain, int $minSize, HttpClient $client): Response
{
    dump($client);die;
```

And we have to check this with `instanceof` in order to not make some tools like PHPStan crying a lot.

It would be wonderful to be able to do something like this:

```php
public function faviconRedirectAction(string $domain, int $minSize, HttpMethodsClient $client): Response
{
    dump($client);die;
```

And directly have autowiring working.

What do you think?

The workaround for people interested like me is pretty simple, one line on your services definition:

```yaml
services:
    Http\Client\Common\HttpMethodsClient: '@httplug.client.default.http_methods'
```

dbu:
this only happens for the "default" client: the first configured client or the one called default.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #261 
| Documentation   | -
| License         | MIT

#### What's in this PR?

Autowire the convenience clients

#### Why?

This is the expected behaviour. We only autowire them if the default client has the decorators to avoid unexpected random autowiring.

#### Checklist

### DIFF
--- a/DependencyInjection/HttplugExtension.php
+++ b/DependencyInjection/HttplugExtension.php
@@ -3,8 +3,10 @@
 namespace Http\HttplugBundle\DependencyInjection;
 
 use Http\Client\Common\BatchClient;
+use Http\Client\Common\BatchClientInterface;
 use Http\Client\Common\FlexibleHttpClient;
 use Http\Client\Common\HttpMethodsClient;
+use Http\Client\Common\HttpMethodsClientInterface;
 use Http\Client\Common\Plugin\AuthenticationPlugin;
 use Http\Client\Common\PluginClient;
 use Http\Client\Common\PluginClientFactory;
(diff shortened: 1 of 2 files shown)


DependencyInjection/HttplugExtension.php
```diff
@@ -111,9 +113,30 @@ private function configureClients(ContainerBuilder $container, array $config)
...
+            } else {
+                $default = 'default';
+            }
+
+            // Autowiring alias for special clients, if they are enabled on the default client
+            if ($config['clients'][$default]['flexible_client']) {
+                $container->setAlias(FlexibleHttpClient::class, $serviceId.'.flexible');
+            }
+            if ($config['clients'][$default]['http_methods_client']) {
+                if (class_exists(HttpMethodsClientInterface::class)) {
```
Nyholm: We should use `interface_exists`

DependencyInjection/HttplugExtension.php
```diff
@@ -111,9 +113,30 @@ private function configureClients(ContainerBuilder $container, array $config)
...
+                    // support for client-common 1.9
+                    $container->setAlias(HttpMethodsClientInterface::class, $serviceId.'.http_methods');
+                }
+            }
+            if ($config['clients'][$default]['batch_client']) {
+                if (class_exists(BatchClientInterface::class)) {
```
Nyholm: Same here.

DependencyInjection/HttplugExtension.php
dbu: ups. i somehow keep forgetting that php distinguishes between interface and class for that thing. fixed!

PR #311, opened 2019-02-08
cache-plugin is not a hard dependency to use this bundle

### LINKED ISSUE #310: Why do we need the cache plugin?
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | no
| Version      |  dev-master

I did a quick search, Why do we need the cache plugin in our require-section?

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | not really
| Deprecations?   | no
| Related tickets | fixes #310 
| Documentation   | https://github.com/php-http/documentation/pull/254
| License         | MIT

#### What's in this PR?

Remove hard dependency on the cache plugin.

#### Why?

When no caching is configured, the bundle does not need that package.

#### Checklist

- [x] Documentation pull request created (if not simply a bugfix) => https://github.com/php-http/documentation/pull/254

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The change log describes what is "Added", "Removed", "Changed" or "Fixed" betwee
 ### Changed
 
 - Removed `twig/twig` dependency
+- Removed hard dependency on `php-http/cache-plugin`. If you want to use the
+  cache plugin, you need to require it in your project.
 
 ## 1.14.0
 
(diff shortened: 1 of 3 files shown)


DependencyInjection/HttplugExtension.php
```diff
@@ -175,6 +177,9 @@ private function configurePluginByName($name, Definition $definition, array $con
     {
         switch ($name) {
             case 'cache':
+                if (!class_exists(CachePlugin::class)) {
```
xabbuh: Should we do that in the `Configuration` class instead to provide a better error message?

dbu: oh, indeed. with a validate it looks much nicer!

PR #313, opened 2019-02-14
Support decorated clients

### LINKED ISSUE #285: Configure client with custom class
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? |yes
| Version      | 1.13.0

We use Symfony 4 auto-wiring. To make it easier to inject the right client inside the service it would be great if I could specify the class name per httplug client.

This feature is supported by [EightPointGuzzleBundle](https://github.com/8p/EightPointsGuzzleBundle) and allows you to do something like this:

```yaml
httplug:
    clients:
        acme:
            classes:
                client: 'My\Custom\HttpClient'
```

I looked at HTTPlug and I see that it uses factories to get this done. I could of course create a factory for every single httplug client, but that feels like a lot of work.

Wondering if there is a better way? I'm willing to create a PR.

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #285
| License         | MIT

#### What's in this PR?

To make auto wire easier I added a `decorate_with` option to the client configuration. You can specify the FQCN of the class you want to wrap around the HTTPlug client. The HTTPlug client will be injected in the constructor of the decorated service.

#### Why?

I have a project with 15 http clients (and counting). This makes it easier to automatically inject those on the right spot. 

#### Example Usage

```yaml
httplug:
    clients:
        my_awesome_service:
            public: true
            decorate_with: My\Namespace\MyAwesomeHttpClient
```

``` php
<?php

use Http\Client\HttpClient;
use Psr\Http\Message\ResponseInterface;

final class MyAwesomeHttpClient implements HttpClient
{
    private $client;

    public function __construct(HttpClient $client)
    {
        $this->client = $client;
    }

    public function sendRequest(RequestInterface $request) : ResponseInterface
    {
        return $this->client->sendRequest($request);
    }
}
```

#### Checklist


adri:
Great idea 👍

dbu:
i see what you want to do. but why not use https://symfony.com/doc/current/service_container/service_decoration.html for that?

ruudk:
Because this makes my configuration much less. I don't need to specify the decorated services for every HTTPlug client in a different file/location.

Tobion:
You only safe one line of config. I don't think that is worth to introduce another config instead of using a known, existing feature.

dbu:
thanks for the input tobion, i will respectfully decline this PR as the standard solution is a different one. i hope that works for you @ruudk? and don't let this deter you from proposing improvements, we want this bundle to work well for heavy users like you!

PR #314, opened 2019-02-14
Add configuration option to disable registration of default client

### LINKED ISSUE #312: Do not register default HttpClient to prevent auto wire issues
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | 1.14.0

#### Actual Behavior

We are heavily using Symfony's auto wire functionality. Sometimes, developer inject HttpClient in the constructor without actually specifying the exact `httplug.client.name`. No error is thrown, and HTTPlug just picks the `httplug.client.default` service (or the first client it can find)

#### Expected Behavior

Don't register a default `Http\Client\HttpClient` service and Symfony will automatically warn the developer that multiple services of `Http\Client\HttpClient` exist. Developer is then notified about this and will think :) 

#### Possible Solutions

I really propose to get rid of this behaviour in the next major version of this bundle.

#### Related code
https://github.com/php-http/HttplugBundle/blob/8b7cb00d58a4699bad47c0bed4c5b0455b9f5704/DependencyInjection/HttplugExtension.php#L115-L123

https://github.com/php-http/HttplugBundle/blob/a6466b46a68b330ce55107cbc6111213f95b5f8c/Resources/config/services.xml#L44-L46

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #312
| License         | MIT

#### What's in this PR?

This allows you to disable the registration of the default HTTPlug client.

#### Why?

We are heavily using Symfony's auto wire functionality. Sometimes, developer inject HttpClient in the constructor without actually specifying the exact `httplug.client.name`. No error is thrown, and HTTPlug just picks the `httplug.client.default` service (or the first client it can find)

#### Checklist

- [ ] Documentation pull request created


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The change log describes what is "Added", "Removed", "Changed" or "Fixed" betwee
 - Autowiring support for FlexibleClient, HttpMethodsClientInterface and
   BatchClientInterface if they are enabled on the default/first client.
   (Only available with Httplug 2)
+- Add configuration option to disable registration of default client
```
dbu: i find it useful when the changelog is specific so i can find the thing it talks about. can we make this "Configuration option `register_default_client` to allow disabling registering a default client"

DependencyInjection/HttplugExtension.php
```diff
@@ -87,6 +87,10 @@ public function load(array $configs, ContainerBuilder $container)
         $this->configureClients($container, $config);
         $this->configurePlugins($container, $config['plugins']); // must be after clients, as clients.X.plugins might use plugins as templates that will be removed
         $this->configureAutoDiscoveryClients($container, $config);
+
+        if (!$config['register_default_client']) {
+            $container->removeDefinition('httplug.client.default');
```
dbu: your goal is to prevent autowiring from happening, right? would it not be better to have an `autowire_client` flag and remove the alias with the client class name instead of the `default` service? with this, you would get an error that the alias is invalid, no?

CHANGELOG.md
dbu: this should also mention autowiring somehow

dbu:
@ruudk wdyt about the questions i asked?

ruudk:
Sorry for not responding. I will pick this up next week.

dbu:
please also rebase on master when you work on this. we moved the php files into src/ and tests/ 

i had no conflicts for other open merge requests, but github seems to not want to automatically solve this.

DependencyInjection/HttplugExtension.php
ruudk: I think the difference here is Symfony 3.4 and 4+. 

With Symfony 3.4: when you have a `default` service, it will just auto wire when you use its class name. 
I believe this changed in 4+ where you have to explicitly set the name of the service to the class name.

I faced this issue in a SF 3.4 application.

CHANGELOG.md
ruudk: Done.

ruudk:
Rebased the PR and didn't apply much changes. See comment above.

dbu:
looking at https://github.com/php-http/HttplugBundle/blob/master/src/Resources/config/services.xml#L44-L45 i am a bit confused. we set the default client to be loaded from discovery if there are no clients defined at all. and then we overwrite that with https://github.com/php-http/HttplugBundle/blob/d05392163155c27ee9a93e2c20d46719105cb897/src/DependencyInjection/HttplugExtension.php#L114-L123 if there are any clients configured. this seems odd.

so the configuration option would prevent both the factory service and aliasing the first client as the default client. 

@Nyholm do you remember why we have the factory based client service as well? so that things work if you configure no client at all?

and the alias in https://github.com/php-http/HttplugBundle/blob/master/src/Resources/config/services.xml#L47 is actually on a client with `default` in the name. did not find where we define that service.

Nyholm:
I like the feature but not the implementation (I think). 

David: At runtime we configure profiling to the auto discovered client if we don't got any. It is a bit messy to say the least. Me and @fbourigault spend some long hours trying to figure stuff out.

@ruudk What about only removing the aliases and not the actual service?

ruudk:
@Nyholm I updated the PR. It now only removes the aliases. Therefore, I think we might want to rename the configuration property to something like `register_default_client_for_autowiring`?

dbu:
agreed, and i think register_default_client_for_autowiring is a good name. it would be `canBeDisabled()` so that users can set enabled: false if they don't want autowiring.

can you please also rebase your pull request on master? i think we fixed the lowest build there.

Nyholm:
Awesome. This is a much simpler implementation :+1:

joelwurtz:
I prefer small value :), what about: `default_client_autowiring` ? (Or even `default_autowiring`)

Nyholm:
`client_interface_autowring`?

dbu:
i would not call it default_autowiring. we could do `autowiring` maybe? i guess there is nothing else that is autowired from this bundle?

joelwurtz:
If we want to use this value for disabling potential futur autowiring of other class, then i'm 👍 for `autowiring` only

ruudk:
> i would not call it default_autowiring. we could do autowire maybe? i guess there is nothing else that is autowired from this bundle?

but what if we add something else in the future that autowires? Then this configuration property becomes to generic and we're screwed.

ruudk:
I really like `default_client_autowiring` because it's clear what it does.

joelwurtz:
And even if we want to split we can do somehting like that in the futur 

```
autowiring:
    defaut_client: false
    other: true
```

 `autowiring`  is the best choice IMO

ruudk:
Oke, I will change it.

ruudk:
Shouldn't we also remove these aliases then?
```xml
        <service id="Http\Message\MessageFactory" alias="httplug.message_factory" public="false" />
        <service id="Http\Message\RequestFactory" alias="httplug.message_factory" public="false" />
        <service id="Http\Message\ResponseFactory" alias="httplug.message_factory" public="false" />
        <service id="Http\Message\StreamFactory" alias="httplug.stream_factory" public="false" />
        <service id="Http\Message\UriFactory" alias="httplug.uri_factory" public="false" />
```

ruudk:
Rebased PR and applied changes.

dbu:
oh, so we do autowire other things. i feel that it makes no sense to disable autowiring the factories. and therefore i would want to call it `default_client_autowiring`. if we have more autowiring that we want to allow to disable, we can still move to the nested structure. @joelwurtz can you live with that? i think clean name is more important than short name. and a nested structure seems too much, given that we probably don't want to un-autowire other things in the future.

joelwurtz:
I'm fine with that :) didn't know either there was autowiring for other classes sorry for the noise.

Thanks @ruudk for checking those things

ruudk:
Updated the PR one more time ;)

tests/Functional/DiscoveryTest.php
```diff
@@ -71,4 +73,35 @@ public function testNoDiscoveryFallbacks()
...
+        $this->load([
+            'default_client_autowiring' => false,
+        ]);
+
+        $this->assertContainerBuilderHasService('httplug.client.default');
+        $this->assertContainerBuilderHasService('httplug.async_client.default');
+
+        self::assertThat(
+            $this->container,
+            new LogicalNot(new ContainerBuilderHasAliasConstraint(HttpClient::class))
```
dbu: i am afraid that this does not work on php 5.x because we use phpunit 5.7 then which is missing the LogicalNot class.

however, i would suggest to simply skip this test if the php version is < 7 rather than do something super complicated to test with end-of-life versions of php.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The change log describes what is "Added", "Removed", "Changed" or "Fixed" betwee
   BatchClientInterface if they are enabled on the default/first client.
   (Only available with Httplug 2)
 - Configuration for the content_type plugin
+- Configuration option `default_client_autowiring` to prevent autowiring 
+  the HttpClient and HttpAsyncClient.
```
dbu: i'd say "Configuration option `default_client_autowiring` that you can set to `false` to prevent autowiring the HttpClient and HttpAsyncClient."

tests/Functional/DiscoveryTest.php
ruudk: Fixed.

CHANGELOG.md
ruudk: Fixed

dbu:
one last thing i noticed: you add a git submodule `documentation` in this pull request. can you please remove that change? otherwise this looks ready to merge to me, thanks!

ruudk:
Good catch. I accidentally staged that directory after all changing this PR 10 times ;) It's fixed now.

dbu:
thanks a lot and sorry for the back and forth!

dbu:
i documented this in https://github.com/php-http/documentation/pull/260

PR #315, opened 2019-02-23
Update MockFactory.php

### LINKED ISSUE #264: Mock client doesn't work
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | 1.10.0

#### Actual Behavior

Every client is decorated with Http\Client\Common\PluginClient, but \Http\HttplugBundle\ClientFactory\MockFactory::setClient() requires exactly \Http\Mock\Client, this throws an exception 
```
TypeError : Argument 1 passed to Http\HttplugBundle\ClientFactory\MockFactory::setClient() must be an instance of Http\Mock\Client, instance of Http\Client\Common\PluginClient given, called in .../getHttplug_Factory_MockService.php
```

#### Expected Behavior

No exception should be thrown.

#### Steps to Reproduce

Just use the standard config:
```
# config_test.yml
httplug:
    clients:
        mock:
            factory: 'httplug.factory.mock'
```
and run any code that uses a client.

#### Possible Solutions

Probably don't decorate mock client at all?

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #264 
| Documentation   | -
| License         | MIT

#### What's in this PR?

Allow setting any client in the mock factory, to not break when we decorate the client for the debug toolbar.

### DIFF
--- a/ClientFactory/MockFactory.php
+++ b/ClientFactory/MockFactory.php
@@ -2,22 +2,25 @@
 
 namespace Http\HttplugBundle\ClientFactory;
 
+use Http\Client\HttpClient;
 use Http\Mock\Client;
 
 /**
  * @author Gary PEGEOT <garypegeot@gmail.com>
  */
-class MockFactory implements ClientFactory
+final class MockFactory implements ClientFactory
 {
     /**
-     * @var Client
+     * @var HttpClient
      */
     private $client;
 
     /**
-     * @param Client $client
+     * Set the client instance that this factory should return.
+     *
+     * Note that this can be any client, not only a mock client.
      */
-    public function setClient(Client $client)
+    public function setClient(HttpClient $client)
     {
         $this->client = $client;
     }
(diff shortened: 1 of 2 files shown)


ClientFactory/MockFactory.php
```diff
@@ -10,14 +11,16 @@
...
     private $client;
 
     /**
-     * @param Client $client
+     * Set the client instance that this factory should return.
+     *
+     * Note that this can be any client, not only a mock client.
      */
-    public function setClient(Client $client)
+    public function setClient(HttpClient $client)
```
Nyholm: This is technically a BC break.

But if I understand it correctly. Using this mock factory has never worked, right?
If so, it would mean that this is a bugfix and I'm 👍

dbu:
yes, from my understanding this never actually worked with our bundle configuration. (people might have done something with their own compiler passes, theoretically).

also, its only a BC break if you extended the factory, which would be an odd thing to do. i wonder if i should make the factory `final` while we are at it. (would break the same way as this bug fix already does, if somebody extended the factory). wdyt?

joelwurtz:
Since it's a bug it should be ok, it's only a bc break if anyone extends this factory (since a Mock Client is a HttpClient usage is still BC) so it's rather limited here, we should be fine.

PR #322, opened 2019-03-05
Allow to disable body limitation size

jdecool:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | https://github.com/php-http/documentation/pull/259
| License         | MIT

#### What's in this PR?

This PR allow to set the `httpplug.profiling.captured_body_length` to `NULL` to avoid body truncate in the Symfony profiler.

#### Why?

I want to view all the body content of a `Request` or `Response`.

#### Checklist

#### To Do

- [x] If accepted, update documentation

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -147,7 +147,14 @@ class Configuration implements ConfigurationInterface
                                         return null;
                                     }
 
-                                    return (int) $maxLength;
+                                    if (!is_int($maxLength)) {
+                                        $invalidConfiguration = new InvalidConfigurationException('The child node "captured_body_length" at path "httplug.profiling" must be an integer or null.');
+                                        $invalidConfiguration->setPath('httplug.profiling');
+
+                                        throw $invalidConfiguration;
+                                    }
+
+                                    return $maxLength;
                                 })
                             ->end()
                             ->defaultValue(0)
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -142,7 +142,7 @@ class Configuration implements ConfigurationInterface
                         ->scalarNode('formatter')->defaultNull()->end()
                         ->scalarNode('captured_body_length')
                             ->beforeNormalization()
-                                ->always(function($maxLength) {
+                                ->always(function ($maxLength) {
                                     if (null === $maxLength) {
                                         return null;
                                     }
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -140,9 +140,18 @@ class Configuration implements ConfigurationInterface
                             ->defaultValue($this->debug)
                         ->end()
                         ->scalarNode('formatter')->defaultNull()->end()
-                        ->integerNode('captured_body_length')
+                        ->scalarNode('captured_body_length')
+                            ->beforeNormalization()
+                                ->always(function($maxLength) {
+                                    if (null === $maxLength) {
+                                        return null;
+                                    }
+
+                                    return (int) $maxLength;
+                                })
+                            ->end()
                             ->defaultValue(0)
-                            ->info('Limit long HTTP message bodies to x characters. If set to 0 we do not read the message body. Only available with the default formatter (FullHttpMessageFormatter).')
+                            ->info('Limit long HTTP message bodies to x characters. If set to 0 we do not read the message body. If null the body will not be truncated. Only available with the default formatter (FullHttpMessageFormatter).')
                         ->end()
                     ->end()
                 ->end()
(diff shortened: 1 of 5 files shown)


joelwurtz:
Thanks, LGTM

src/DependencyInjection/Configuration.php
```diff
@@ -140,9 +140,18 @@ public function getConfigTreeBuilder()
...
                         ->scalarNode('formatter')->defaultNull()->end()
-                        ->integerNode('captured_body_length')
+                        ->scalarNode('captured_body_length')
+                            ->beforeNormalization()
+                                ->always(function ($maxLength) {
+                                    if (null === $maxLength) {
+                                        return null;
+                                    }
+
+                                    return (int) $maxLength;
```
dbu: i'd prefer to throw an InvalidConfigurationOption at this point `if (!is_int($maxLength))`. otherwise one might set it to 0 when accidentally passing a string.

jdecool:
I've created the documentation PR, and update this PR description.

jdecool:
> i am not sure if this is a good idea. you could set the limit to 100k or whatever crazy high value. removing the limit could break in bad ways when you ever happen to load several megabyte of data. 

Sure, it could be dangerous. But when I use the Symfony profiler (in dev mode), I don't want that my body content will be truncate.

In a debug purpose, I want to have all information I need.

I'm going to open a new PR in the `message` component to update the formatter.

dbu:
> I'm going to open a new PR in the message component to update the formatter.

i just changed the code that you will adjust, please check the newest master: https://github.com/php-http/message/pull/111

jdecool:
> i just changed the code that you will adjust, please check the newest master: php-http/message#111

👍

joelwurtz:
@dbu I think there is some cases where it's useful, specially when uou know that the remote api will never throw a too big responsel. 

As an example when using a client for elasticsearch, i know that i will never get a too big output, and that i can always show the the full body, (but i often hit the default limit here).

tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -395,4 +395,22 @@ public function testCacheConfigMustHavePool()
...
+    {
+        $file = __DIR__.'/../../Resources/Fixtures/config/limitless_captured_body_length.yml';
+        $config = $this->emptyConfig;
+        $config['profiling']['captured_body_length'] = null;
+        $this->assertProcessedConfigurationEquals($config, [$file]);
+    }
+
+    /**
+     * @expectedException \Symfony\Component\Config\Definition\Exception\InvalidConfigurationException
+     * @expectedExceptionMessage The child node "captured_body_length" at path "httplug.profiling" must be an integer or null.
```
dbu: please do not use this deprecated form to declare expected exceptions anymore. use `$this->expectException()` and `$this->expectExceptionMessage()` instead (right above the line that is epxected to throw the exception).

i am aware that in this case, the difference is irrelevant. but with longer tests, having the expectations on the method means confusion if an exception gets thrown before it was actually supposed to be triggered. also, we can use the ::class constant when expecting in the code, which is nicer.

dbu:
apart from that i am okay to merge this as soon as we have the merge request against php-http/message to state that `null` is explicitly supported.

tests/Unit/DependencyInjection/ConfigurationTest.php
jdecool: This change breaks test because the assert `expectException` and `expectExceptionMessage` are not available in PHPUnit 4.8.36 which is load to test the bundle with PHP 5.5.

jdecool: see https://travis-ci.org/php-http/HttplugBundle/jobs/503055333

dbu: okay, sorry about that one. we really should bump the minimal version for this bundle :-/

but then its fine as is.

dbu:
thanks for this contribution (and your patience while we figured out how to exactly do this ;-) )

jdecool:
No problem 😄

src/DependencyInjection/Configuration.php
```diff
@@ -140,9 +140,25 @@ public function getConfigTreeBuilder()
...
-                        ->integerNode('captured_body_length')
+                        ->scalarNode('captured_body_length')
+                            ->beforeNormalization()
+                                ->always(function ($maxLength) {
+                                    if (null === $maxLength) {
+                                        return null;
+                                    }
+
+                                    if (!is_int($maxLength)) {
+                                        $invalidConfiguration = new InvalidConfigurationException('The child node "captured_body_length" at path "httplug.profiling" must be an integer or null.');
```
xabbuh: Sorry for being late, but wouldn't it be more straight-forward to use something like `ifTrue(...)->thenInvalid()` construct?

What I have in mind would be something like this:

```php
->scalarNode('captured_body_length')
    ->validate()
        ->ifTrue(function ($v) {
            return null !== $v && !is_int($v);
        })
        ->thenInvalid('The child node "captured_body_length" at path "httplug.profiling" must be an integer or null ("%s" given).')
    ->end()
->end()
```

dbu: agreed. i created #323 with your proposed change

PR #333, opened 2019-04-15
Add integration for VCR plugin

### LINKED ISSUE #330: [Plugin] Add configuration for VCR plugin
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes
| Version      | master

#### Actual Behavior

No easy way to configure the [VCR plugin](https://github.com/php-http/vcr-plugin)

#### Expected Behavior

Same configuration pattern as the other plugins

#### Possible Solutions
Either a kind of "all-in-one" implementation:
```yaml
// config.yml
httplug:
    clients:
        acme:
            # ...
            plugins:
                # ...
                - record_and_replay: # or vcr
                    mode: record # record | replay | replay_or_record
                    recorder: filesystem # optionnal
                    naming_strategy: service_id.of.naming_strategy # optionnal

```

Or a more explicit way, ie be defining a record plugin, and a replay plugin

```yaml
// config.yml
httplug:
    clients:
        acme:
            # ...
            plugins:
                # ...
                - replay:
                    throw_on_missing: false # Default to true
                - record: ~
    plugins:
        record_and_replay: # or vcr
            recorder: filesystem # optionnal
            naming_strategy: service_id.of.naming_strategy # optionnal

```
Taking the risk that people put record before replay.

Any design preference @dbu?

GaryPEGEOT:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #330 
| Documentation   | php-http/documentation#267
| License         | MIT

#### What's in this PR?

Add an easy way to configure the VCR plugin

#### Example Usage

```yaml
# config/packages/dev/httplug.yaml
httplug:
    # ...
    clients:
        # ...
        acme:
            # ...
            plugins:
                -
                    vcr:
                        mode: replay # record | replay | replay_or_record
                        recorder: filesystem # optional
                        fixtures_directory: '%kernel.project_dir%/fixtures/http' # mandatory for "filesystem" recorder
                        naming_strategy: service_id.of.naming_strategy # optional
                        naming_strategy_options: # optional
                            hash_headers: []
                            hash_body_methods: []
```

#### Checklist

- [X] Updated CHANGELOG.md
- [x] Documentation pull request created: php-http/documentation#267


GaryPEGEOT:
@dbu about the travis failure, should I try to lower VCR plugin requirements to Symfony 2.8?

src/DependencyInjection/HttplugExtension.php
```diff
@@ -506,13 +520,77 @@ private function configurePlugin(ContainerBuilder $container, $serviceId, $plugi
...
+            $recorderDefinition->replaceArgument(0, $config['fixtures_directory']);
+            $recorderDefinition->setPublic(false);
+            $recorderId = "$prefix.recorder";
+
+            $container->setDefinition($recorderId, $recorderDefinition);
+        }
+
+        if ('default' === $config['naming_strategy']) {
+            $namingStrategyId = "$prefix.naming_strategy";
+            $namingStrategy = $container->register("$prefix.naming_strategy", PathNamingStrategy::class)
```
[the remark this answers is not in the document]
GaryPEGEOT: Fixed, and the string concat as well

src/DependencyInjection/Configuration.php
```diff
@@ -437,6 +440,50 @@ private function createClientPluginNode()
...
+                        ->ifTrue(function ($config) {
+                            return 'filesystem' === $config['recorder'] && empty($config['fixtures_directory']);
+                        })
+                        ->thenInvalid('If you want to use the "filesystem" recorder you must also specify a "fixtures_directory".')
+                    ->end()
+                    ->children()
+                        ->enumNode('mode')
+                        ->info('What should be the behavior of the plugin?')
+                        ->values(['record', 'replay', 'replay_or_record'])
+                        ->defaultValue('replay_or_record')
```
dbu: the defensive value would be to default to "replay". i am inclined to make this have no default and be a required field to force people to actively decice which mode is right for them (if they enable the vcr at all)

src/DependencyInjection/HttplugExtension.php
```diff
@@ -506,13 +531,79 @@ private function configurePlugin(ContainerBuilder $container, $serviceId, $plugi
...
         $this->configurePluginByName($pluginName, $definition, $pluginConfig, $container, $pluginServiceId);
         $container->setDefinition($pluginServiceId, $definition);
 
         return $pluginServiceId;
     }
+
+    private function configureVcrPlugin(ContainerBuilder $container, array $config, $prefix)
+    {
+        $recorder = $config['recorder'];
+        $recorderId = $container->hasDefinition($recorder) ? $recorder : 'httplug.plugin.vcr.recorder.'.$recorder;
```
dbu: i'd say `in_array($recorder, ['filesystem', 'in_memory']) ? 'httplug.plugin.vcr.recorder.'.$recorder : $recorder;`

otherwise we break on somebody define filesystem / in_memory as a service in their application, and we end up with a confusing error if the service is not defined. also, hasDefinition depends on the order in which configuration is loaded, the $recorder service could be defined later than this bundle is loaded.

src/DependencyInjection/HttplugExtension.php
```diff
@@ -94,6 +103,14 @@ public function load(array $configs, ContainerBuilder $container)
             $container->removeAlias(HttpAsyncClient::class);
             $container->removeAlias(HttpClient::class);
         }
+
+        if ($this->useVcrPlugin) {
```
dbu: unless i miss something, we don't set this property to another value. it should be a check whether the 'vcr' configuration is enabled in the resolved config.

src/DependencyInjection/Configuration.php
GaryPEGEOT: I removed the default value

src/DependencyInjection/HttplugExtension.php
GaryPEGEOT: Fixed in `03ccd71`.

dbu: ah, and we need to do it here because the plugin can be configured for several clients, but we only want to do this thing once.

tests/Unit/DependencyInjection/HttplugExtensionTest.php
```diff
@@ -449,4 +450,67 @@ public function testBatchClientCanBePublic()
...
+     * @group vcr-plugin
+     */
+    public function testVcrPluginConfiguration(array $config, array $services, array $arguments = [])
+    {
+        $prefix = 'httplug.client.acme.vcr';
+        $this->load(['clients' => ['acme' => ['plugins' => [['vcr' => $config]]]]]);
+        $this->assertContainerBuilderHasService('httplug.plugin.vcr.recorder.in_memory', InMemoryRecorder::class);
+
+        foreach ($services as $service) {
+            $this->assertContainerBuilderHasService("$prefix.$service");
```
dbu: we do not use the inline string expansion thing. can you please rewrite this as `$prefix.'.'.$service`? (its a bit weird, but i'd like to keep a consistent code style)

src/DependencyInjection/HttplugExtension.php
```diff
@@ -508,13 +533,79 @@ private function configurePlugin(ContainerBuilder $container, $serviceId, $plugi
...
+
+        return $plugins;
+    }
+
+    /**
+     * @param string $parent the parent service id
+     *
+     * @return ChildDefinition
+     */
+    private function getChildDefinition($parent)
```
dbu: can we name this `createChildDefinition`? i was trying to figure out why we can access a definition before the service file defining the vcr definitions has been loaded... until i read this method and realized it is a factory method (for the BC thingy).

src/DependencyInjection/HttplugExtension.php
```diff
@@ -508,13 +533,79 @@ private function configurePlugin(ContainerBuilder $container, $serviceId, $plugi
...
+                break;
+        }
+
+        return $plugins;
+    }
+
+    /**
+     * @param string $parent the parent service id
+     *
+     * @return ChildDefinition
```
dbu: `|DefinitionDecorator`

can you please add a description? "BC for old Symfony versions. Remove this method and use `new ChildDefinition` directly when we drop support for Symfony 2."

dbu:
oh, and the build fails where we use php < 7.1 because vcr is only available for recent php versions. i think its time to raise the php version for the bundle. (afaik we said we can allow outdated versions until it becomes a problem - now it did become a problem)

src/DependencyInjection/HttplugExtension.php
GaryPEGEOT: Fixed! I also added the doc PR.

dbu:
thanks a lot for this contribution! and thanks for your patience with all my nitpicks ;-)

PR #334, opened 2019-04-18
Remove enabled key from content type plugin options

EmiiKhaos:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Removes the `enabled` key from the `$config` array passed to the `ContentTypePlugin` constructor options.

#### Why?

The `content_type` plugin configuration can be enabled and thus has an `enabled` option in it's configuration. If this is passed without removing the `enabled` key, the options resolver of the `ContentTypePlugin` constructor options throws an exception:

```
The option "enabled" does not exist. Defined options are: "size_limit", "skip_detection".
```

#### Checklist

- [x] Updated CHANGELOG.md to describe bugfix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+### Fixed
+
+- Fix to pass only allowed options to the `ContentTypePlugin`. 
+
 ## 1.15.1 - 2019-04-12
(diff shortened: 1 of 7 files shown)


dbu:
tagged 1.15.2 with this fix

PR #336, opened 2019-05-30
Fix: Drop support for unsupported PHP versions

localheinz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | #333 
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

This PR

* [x] drops support for unsupported PHP versions following the merge of #333

#### Why?

The build is currently broken as [`php-http/vcr-plugin:^1.0@dev`]() requires PHP 7.1.

#### Example Usage

n/a

#### Checklist

- [x] Updated `CHANGELOG.md` to describe BC breaks

#### To Do

n/a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,6 @@ matrix:
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak" SYMFONY_PHPUNIT_VERSION="5.7"
 
           # Test the latest stable release
-        - php: 5.5
-          env: SYMFONY_PHPUNIT_VERSION="5.7"
-        - php: 5.6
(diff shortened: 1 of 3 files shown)


localheinz:
Looks like this doesn't actually fix the build, but requires more work.

dbu:
thanks.

looking at https://travis-ci.org/php-http/HttplugBundle/jobs/539402830 i think its also time to ditch symfony 2 support, wdyt?

a lot of builds are red because of deprecations that are not actual bugs. we might still adjust to avoid the deprecation to be future proof.

dbu:
did all of the removing old stuff in #340 
thanks for getting the ball rolling here !

localheinz:
Thank you, @dbu!

PR #348, opened 2019-06-21
Enhancement: Tag clients

localheinz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #346
| Documentation   | https://github.com/php-http/documentation/pull/268
| License         | MIT

#### What's in this PR?

This PR 

* [x] adds the tag `'httplug.client'` to clients created by the `HttplugExtension`

#### Why?

This would allow fetching services by tag and replacing them, for example, with a concrete `Mock\Client`, without having to know the exact service identifier. 

#### Checklist

- [x] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
- [x] Documentation pull request created (if not simply a bugfix)

PR #349, opened 2019-06-21
Enhancement: Configure visibility_required fixer

localheinz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

This PR

* [x] fixes the indentation in `php_cs.dist`
* [x] configures the `visibility_required` fixer to also require visibility on constants

Somewhat related to #348.

#### Why?

The minimum required PHP version is PHP 7.1, so we might just as well make use of the features.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -8,7 +8,7 @@ $finder = PhpCsFixer\Finder::create()
 
 return PhpCsFixer\Config::create()
     ->setRules([
-         '@Symfony' => true,
-         'array_syntax' => ['syntax' => 'short'],
+        '@Symfony' => true,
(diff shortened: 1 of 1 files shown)


localheinz:
Thank you, @dbu!

dbu:
thanks! 
we should port such config to https://github.com/php-http/boilerplate/ for what we can, and apply it in all repos of this organization. that process is not automated, but it would be good to be consistent. do you want to do that?

localheinz:
@dbu 

Can do!

PR #350, opened 2019-06-21
Enhancement: Apply void_return fixer to tests

localheinz:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | n/a
| Documentation   | n/a
| License         | MIT

#### What's in this PR?

This PR

* [x] applies the `void_return` fixer to test code

I ran

```
$ php-cs-fixer fix --allow-risky=yes --diff --verbose --rules=void_return tests
```

#### Why?

We require at least PHP 7.1, so why not?

#### Todo

* [ ] Should we also apply this fixer to production code?

### DIFF
--- a/tests/Functional/DiscoveredClientsTest.php
+++ b/tests/Functional/DiscoveredClientsTest.php
@@ -15,7 +15,7 @@ use Symfony\Component\EventDispatcher\Event;
 
 class DiscoveredClientsTest extends WebTestCase
 {
-    public function testDiscoveredClient()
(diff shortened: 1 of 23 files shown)


localheinz:
Thank you, @dbu!

dbu:
i merged as this is an improvement. can you look if we can set up the rule in the config file and apply it for src too or if that leads to any changes that look risky?

PR #353, opened 2019-07-30
TypeError: Argument 1 passed to ProfileClient::collectExceptionInformations() must be an instance of Exception

spolischook:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Fix error 

#### Why?
```
TypeError: Argument 1 passed to Http\HttplugBundle\Collector\ProfileClient::collectExceptionInformations() must be an instance of Exception, instance of TypeError given, called in /home/spolischook/www/mobile-api/vendor/php-http/httplug-bundle/src/Collector/ProfileClient.php on line 144
```

#### Checklist

#### To Do

- [] If the PR is not complete but you want to discuss the approach, list what remains to be done here

### DIFF
--- a/src/Collector/Formatter.php
+++ b/src/Collector/Formatter.php
@@ -44,11 +44,11 @@ class Formatter implements MessageFormatter
     /**
      * Formats an exception.
      *
-     * @param Exception $exception
+     * @param \Throwable $exception
      *
      * @return string
      */
-    public function formatException(Exception $exception)
+    public function formatException(\Throwable $exception)
     {
         if ($exception instanceof HttpException) {
             return $this->formatter->formatResponse($exception->getResponse());
--- a/tests/Unit/Collector/ProfileClientTest.php
+++ b/tests/Unit/Collector/ProfileClientTest.php
@@ -151,7 +151,7 @@ class ProfileClientTest extends TestCase
         $this->client
             ->expects($this->once())
             ->method('sendRequest')
-            ->willThrowException(new \TypeError('You set string to int prop'));
+            ->willThrowException(new \Error('You set string to int prop'));
 
         $response = $this->subject->sendRequest($this->request);
     }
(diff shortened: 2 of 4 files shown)


spolischook:
@Nyholm @fbourigault @dbu could you please approve/merge it?

src/Collector/Formatter.php
```diff
@@ -44,11 +44,11 @@ public function __construct(MessageFormatter $formatter, CurlCommandFormatter $c
     /**
      * Formats an exception.
      *
-     * @param Exception $exception
+     * @param \Throwable $exception
      *
      * @return string
      */
-    public function formatException(Exception $exception)
+    public function formatException(\Throwable $exception)
```
Nyholm: This is not a BC break because the class is internal.

spolischook: even more, Throwable is wide variation of Exception, so no BC break will happen

Nyholm: Yes. But that reason alone is still a BC break because the class is not final.

tests/Unit/Collector/ProfileClientTest.php
```diff
@@ -34,7 +35,7 @@ class ProfileClientTest extends TestCase
     private $activeStack;
 
     /**
-     * @var HttpClient
+     * @var HttpClient|MockObject
```
dbu: right.

just a comment, no need to change anything now: i have seen the syntax `HttpClient&MockObject` which is semantically correct and needed from some phpstan level on so phpstan is sure about the type of the variable. however, e.g. phpstorm seems not to know about it...

spolischook: Yea, it's right, PhpStorm don't suggest any method of MockObject so I should to add this,

PR #354, opened 2019-08-12
Adds link to debug token.

dbrumann:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

If a response contains a header `X-Debug-Token-Link`, i.e. a link to a Symfony debug toolbar profiler page, a clickable link is added to easily navigate to it.

#### Why?

This is purely for a more convenient debugging when investigating Responses from a (different) Symfony application.

#### Checklist

### DIFF
--- a/src/Collector/ProfileClient.php
+++ b/src/Collector/ProfileClient.php
@@ -173,6 +173,10 @@ class ProfileClient implements HttpClient, HttpAsyncClient
         $stack->setDuration($event->getDuration());
         $stack->setResponseCode($response->getStatusCode());
         $stack->setClientResponse($this->formatter->formatResponse($response));
+        if ($response->hasHeader('X-Debug-Token-Link')) {
+            $stack->setDebugToken($response->getHeaderLine('X-Debug-Token'));
+            $stack->setDebugTokenLink($response->getHeaderLine('X-Debug-Token-Link'));
+        }
     }
 
     /**
--- a/src/Collector/Stack.php
+++ b/src/Collector/Stack.php
@@ -82,6 +82,16 @@ final class Stack
     private $responseCode;
 
     /**
+     * @var string|null
+     */
+    private $debugToken;
+
+    /**
+     * @var string|null
+     */
+    private $debugTokenLink;
+
+    /**
      * @var int
      */
     private $duration = 0;
@@ -275,6 +285,38 @@ final class Stack
     public function setResponseCode($responseCode)
     {
         $this->responseCode = $responseCode;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDebugToken()
+    {
+        return $this->debugToken;
+    }
+
+    /**
+     * @param string $debugToken
+     */
+    public function setDebugToken($debugToken)
+    {
+        $this->debugToken = $debugToken;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDebugTokenLink()
+    {
+        return $this->debugTokenLink;
+    }
+
+    /**
+     * @param string $debugTokenLink
+     */
+    public function setDebugTokenLink($debugTokenLink)
+    {
+        $this->debugTokenLink = $debugTokenLink;
     }
 
     /**
--- a/src/Resources/views/stack.html.twig
+++ b/src/Resources/views/stack.html.twig
@@ -38,7 +38,7 @@
             {{ stack.clientRequest|httplug_markup|nl2br }}
         </div>
         <div class="httplug-message card">
-            <h4>Response</h4>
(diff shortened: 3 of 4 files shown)


src/Resources/views/stack.html.twig
```diff
@@ -38,7 +38,7 @@
             {{ stack.clientRequest|httplug_markup|nl2br }}
         </div>
         <div class="httplug-message card">
-            <h4>Response</h4>
+            <h4>Response {% if stack.debugTokenLink %}<span class="label status-info"><a href="{{ stack.debugTokenLink }}">{{ stack.debugTokenLink }}</a></span>{% endif %}</h4>
```
dbrumann: Possibly instead of showing the full link it might make sense to only display the token. In that case I guess an icon or something else indicating what the token is for would be helpful. Maybe there are other suggestions?

src/Collector/Stack.php
```diff
@@ -81,6 +81,11 @@ final class Stack
      */
     private $responseCode;
 
+    /**
+     * @var string
```
xabbuh: We can use type declarations in the methods below, can't we?

dbu: yep. still should keep them on properties i think (at least that has been the best practice we followed so far)

src/Resources/views/stack.html.twig
dbu: i like the idea of showing e.g. the token symfony uses itself for the debug toolbar, along with a "external link" icon. showing the full url could be quite long, and shortening might not show the right fragment.

dbrumann:
Made a few small changes. Here is an example screenshot:
Screenshot 2019-08-12 at 21 39 42

src/Collector/ProfileClient.php
```diff
@@ -173,6 +173,10 @@ private function collectResponseInformations(ResponseInterface $response, Stopwa
         $stack->setDuration($event->getDuration());
         $stack->setResponseCode($response->getStatusCode());
         $stack->setClientResponse($this->formatter->formatResponse($response));
+        if ($response->hasHeader('X-Debug-Token-Link')) {
```
xabbuh: Should we check for both headers that we request below?

dbrumann: I don't think it's necessary. The link is mandatory, but if `X-Debug-Token` is not provided all that happens is the label is shortened to `Debug Token: ` which admittedly is not ideal but would still work

src/Resources/views/stack.html.twig
dbrumann: I replaced the URL with the Token (see screenshot). I decided against adding an icon for now. If there is an svg I could use where we know that licensing is not an issue we can add it later.

dbu: awesome! as there is no super obvious icon that all of us immediately think of, i prefer text over icon.

src/Collector/Stack.php
dbu: but this is `string|null` i think, its not always set.

src/Collector/Stack.php
```diff
@@ -277,6 +287,38 @@ public function setResponseCode($responseCode)
         $this->responseCode = $responseCode;
     }
 
+    /**
+     * @return string
+     */
+    public function getDebugToken()
```
dbu: can you please add `: ?string` instead of the `@return` annotation?

src/Collector/Stack.php
```diff
@@ -277,6 +287,38 @@ public function setResponseCode($responseCode)
...
+    {
+        return $this->debugToken;
+    }
+
+    /**
+     * @param string $debugToken
+     */
+    public function setDebugToken($debugToken)
```
dbu: please write this as `setDebugToken(string $debugToken): void` (and then we can remove the `@param` because it does not add any information anymore.

src/Collector/Stack.php
dbrumann: Yes, it is. There was no `null` specified on the other properties, so I wasn't sure whether they are always set or if `null`-able is just not marked. I will update the hint.

src/Collector/Stack.php
dbrumann: Right now this is in line with the other properties. How about we keep it as is and instead I will do a follow up PR that adds type hints and return types for all properties on `Stack` to not mix styles?

src/Collector/Stack.php
dbu: annotations seem to be a bit sloppy here. i'd make the new ones correct nonetheless. when you do the return type declaration, php would explode when that one is not correct, and having the property itself correct will help us to see what it needs to be.

dbu:
can you please add the changelog entry? i think otherwise we are good.

dbrumann:
Yes, I will update the PR later this afternoon, including the changelog entry.

dbu:
totally missed that you had updated this. thanks, nice feature!

Issue #245, opened 2018-01-06
Improve design of profiler and add support for formatting request/response body


ostrolucky:
| Q            | A
| ------------ | ---
| Bug?         | no
| New Feature? | yes

Please compare how does profiler look like between these bundles. csa/guzzle-bundle is A LOT clearer.

<table>
<tr><th>csa/guzzle-bundle</th><th>php-http/httplug-bundle</th></tr>
<tr><td><img src="https://user-images.githubusercontent.com/496233/34644966-286466c0-f342-11e7-9f4e-35a4156006b1.png" width="400" alt="csa/guzzle-bundle profiler" /></td>
<td valign="top"><img src="https://user-images.githubusercontent.com/496233/34644977-8c2fe486-f342-11e7-8032-e896f09cf022.png" width="400" alt="php-http/httplug-bundle profiler"  /></td>
</tr>
</table>

I suggest:
1. Syntax highlight json responses (even better, xml too)
1. Don't put request/response side-by-side
1. Use table for headers
  

edit: Syntax highlighting/formatting might look like a hard part. It isn't, look at referenced bundle. [Detection and formatting](https://github.com/csarrazi/CsaGuzzleBundle/blob/2.x/src/Twig/Extension/GuzzleExtension.php#L38-L65), syntax highlight is done by [prismjs](http://prismjs.com/)

dbu:
+1 for headers as table. syntax highlighting could indeed be cheap with some js library. the side-by-side could also be a responsive layout to switch below a certain screen size.

do you want to start with a pull request for headers in a table? that sounds the most quickwin to me.

for syntax highlighting, i am +1 even though it means bundling some js library to do that.

fbourigault:
Request and response are displayed side by side because when you toggle the plugin execution, you can read starting from the left going down each plugin change in the request and then go up on the right side to see how plugins modify the response. Also, syntax highlighting would be great but be careful as the content may be truncated depending on the body length captured while profiling.

dbu:
> Request and response are displayed side by side because when you toggle the plugin execution, you can read starting from the left ...

ah right. wdyt about having the csa bundle layout first, but once you activate the plugin display, it switches to our current side-by-side layout? maybe additionally responsive so that on really large screens, we always get side by side?

fbourigault:
Yes, we could have a different layout for the main display.

ostrolucky:
[...]

PR #355, opened 2019-11-17
[Profiler] Use tabular design for HTTP request/headers

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | partially #245
| Documentation   |
| License         | MIT

#### What's in this PR and why?

Improved profiler design. Check [Before](https://user-images.githubusercontent.com/496233/69001989-3605b500-08e8-11ea-90b2-9a1c7b6d0b3f.png) and [After](https://user-images.githubusercontent.com/496233/69001956-dc9d8600-08e7-11ea-91a0-f44dad85a106.png). Also check linked issue.

#### Checklist

#### Out of scope
- `<t(d|h)>` text trimming. I just want to make this simple first time. This can be added in future, at this time horizontal scrolling works perfect here.
- Syntax highlighting for body. Should be done in separate PR, I want to minimze scope here which will help review and merge this faster.

### DIFF
--- a/src/Resources/views/http_message.html.twig
+++ b/src/Resources/views/http_message.html.twig
@@ -1,0 +1,29 @@
+{% set hasReachedBody = false %}
+{% set content = '' %}
+{% set data = data|split("\n")|slice(1) %}
+{% set xdebugTokenLink = data|filter(v => 'x-debug-token-link:' in v|lower)|first|split(': ')|last %}
+
+<table>
+    <thead>
+    <tr>
+        <th scope="col" class="key" colspan="2">{{ header }}{% if xdebugTokenLink %} <span style="float:right"><a href="{{ xdebugTokenLink }}">Profile link</a></span>{% endif %}</th>
+    </tr>
+    </thead>
+    <tbody>
+    {% for row in data %}
+        {% if row is empty %}
+            {% set hasReachedBody = true %}
+        {% elseif hasReachedBody %}
+            {% set content = content ~ "\n" ~ row %}
+        {% else %}
+            {% set row = row|split(':') %}
+            <tr>
+                <th>{{ row[0] }}</th>
+                <td>{{ row|slice(1)|join(':')|trim }}</td>
+            </tr>
+        {% endif %}
+    {% endfor %}
+    </tbody>
+</table>
+
+<div class='httplug-http-body httplug-hidden'>{{ content|nl2br ?: '(This message has no captured body)' }}</div>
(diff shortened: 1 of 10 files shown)


dbu:
awesome, thanks! and completely agree to keep syntax highlighting or more complicated layouting tweak to separate pull requests to make it easier to review.

can you please rebase? there is a conflict (i guess because #354 adding a link to the debug token)

ostrolucky:
I've slightly changed the design of #354, because its label style didn't fit in my feature, hopefully that's fine. 

Screenshot 2019-11-18 at 10 32 54

dbu:
i like it! @dbrumann as you contributed the profile link, i ask you if you are happy with this too or want a change?

dbrumann:
These changes look great. :+1: I like it. I will have a look at the code changes later today, but consider the end result approved

ostrolucky:
It just accured to me that maybe we could just highlight any link in header table? Because if you notice, there is a x-debug-token-link header there already. That would be more universal

dbu:
as this is a very specific thing to symfony, i like the link at the top.

additionally detecting urls in headervalues and link them sounds like a good idea (for a separate pull request)

ostrolucky:
Allright. We can also reduce complexity further by getting token link from header to display in twig, then Profiler no longer needs to store it separately.

dbu:
indeed. i see you removed the token from the display and from the profiler data (i think thats good, the token code does not help and is annoying to copy-paste if its part of a link anyways, better copy-paste from the headers). want to do that in this PR or separately?

ostrolucky:
I can do it here

ostrolucky:
done

src/Resources/views/http_message.html.twig
```diff
@@ -0,0 +1,29 @@
+{% set hasReachedBody = false %}
+{% set content = '' %}
+{% set data = data|split("\n")|slice(1) %}
+{% set xdebugTokenLink = data|filter(v => 'x-debug-token-link:' in v|lower)|first|split(':')|slice(1)|join(':')|trim %}
```
dbu: could you instead split with a limit of 1, to not break up the url and then have to re-join it? would be more elegant imho. `data|filter(v => 'x-debug-token-link:' in v|lower)|first|split(':', 2)|slice(1)|trim`

ostrolucky: I came up with even better simplification. WDYT?

dbu:
nice!

PR #357, opened 2019-11-17
[WIP] Update composer.json

GrahamCampbell:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #356
| Documentation   | N/A
| License         | MIT

#### To Do

- [ ] Test on multiple symfony versions

PR #360, opened 2019-11-27
Symfony5

### LINKED ISSUE #358: Conflict with twig 3.0 / symfony 4.4
| Q            | A
| ------------ | ---
| Bug?         | yes
| New Feature? | no
| Version      | N/A

#### Actual Behavior

composer install stops

#### Expected Behavior

it should work just fine

#### Steps to Reproduce

- upgrade a project to symfony 4.4 (+symfony/twig-bundle)
- install httplug-bundle

#### Possible Solutions

- use twig namespaced file names
- or forbid installation of twig 3.0

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #357, fixes #358
| Documentation   | -
| License         | MIT

#### What's in this PR?

Continue #357 and make sure symfony 5 can actually be installed and does not error.

#### Checklist


.travis.yml
```diff
@@ -34,7 +33,7 @@ matrix:
         - php: 7.3
           env: DEPENDENCIES="dunglas/symfony-lock:^3"
         - php: 7.3
-          env: DEPENDENCIES="dunglas/symfony-lock:^4"
+          env: DEPENDENCIES="dunglas/symfony-lock:^4" SYMFONY_DEPRECATIONS_HELPER="max[direct]=0"
```
dbu: this prevents deprecations triggered by the symfony/twig-bundle 4.4 from breaking the build.

fbourigault: According to https://symfony.com/doc/current/components/phpunit_bridge.html#direct-and-indirect-deprecations, would `max[self]=0` be a saner value? Also, moving it to the `env.global` would be better, isn't it?

dbu: right, self seems right. right now it works, but if we trigger our own deprecations, it can stop working.

not sure if we want this to be global. i guess actually yes - then we don't need a config for it at all but can just unconditionally do it. we don't care if other things trigger deprecations, we only care if our own code triggers deprecations.

dbu: erm, i mean make it global. somehow mixed it up with another thing where i do a composer config for flex :-D

dbu: changed

XWB:
@dbu Looks good. Can you merge this PR into master so we can test it before releasing?

jrjohnson:
Possible to see this in a release? Seems like a fix for issue I'm having in Symfony 4.4 as well.

dbu:
soon. we are wrapping some things up and hope to tag later this week. :santa: :gift:

ruudk:
Anything I can do to help getting this released before Christmas? 🎄 🌟 ?

dbu:
hi ruudk, the problem is the cache plugin. we recently merged a contribution for blacklisting urls from being cached, and there is confusion on https://github.com/php-http/cache-plugin/pull/62 but the author is not responding. there i tend to reject the change and keep things as they are. if you can have a look at it and input your impression, that would help.

there are https://github.com/php-http/HttplugBundle/pull/361 and https://github.com/php-http/HttplugBundle/pull/362 that i'd want to merge to the bundle before tagging. if you could pick those branches, wrap things up and do a new PR with the whole thing, that would help indeed.

ruudk:
I checked all 3 PR's, and they are all from the same author, that doesn't respond to feedback anymore. Is it that important to hold the release? Can't it be done in the next version? The worst that could happen is that it requires another minor version?

dbu:
ftr: tobias tagged 1.17.0

PR #361, opened 2019-11-28
Adding Support for cache-plugins cache_listeners

nobYsDarling:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Wraps cache_listeners option of cache plugin, to enable its configuration.

#### Why?

Cache listeners could not have been configured using this bundle. (or - which is absolutely possible - I did not get the usage)

#### Example Usage
```
httplug: 
    cache:
        config:
            cache_listeners: ['\Http\Client\Common\Plugin\Cache\Listener\AddHeaderCacheListener']
```

#### Checklist

- [X] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix


nobYsDarling:
Not quite sure, how to handle the travis ci fails. It seems this is due to other dependencies. an composer install is not possible right now due to conflicts of SymfonyDependencyInjectionTest and phpunit.

src/DependencyInjection/Configuration.php
```diff
@@ -739,6 +741,27 @@ private function createCachePluginNode()
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('cache_listeners')
+                    ->info('An array of classes to act on the response based on the results of the cache check. Must implement '.CacheListener::class.'. Defaults to an empty array.')
```
dbu: the CachePlugin expects instances. this configuration should be a list of service names rather than class names. a listener might have a non-empty constructor for example.

src/DependencyInjection/Configuration.php
```diff
@@ -739,6 +741,27 @@ private function createCachePluginNode()
...
+                    ->beforeNormalization()->castToArray()->ifEmpty()->thenUnset()->end()
+                    ->defaultValue([])
+                    ->prototype('scalar')
+                        ->validate()
+                            ->ifTrue(function ($v) {
+                                $vs = is_array($v) ? $v : (is_null($v) ? [] : [$v]);
+
+                                return empty($vs) || array_reduce($vs, function ($r, $e) {
+                                    return empty($e) || !class_exists($e) || !(new ReflectionClass($e))->implementsInterface(CacheListener::class);
```
dbu: when these are service names, the checks would need to move to a compiler extension.

src/DependencyInjection/HttplugExtension.php
```diff
@@ -203,10 +203,22 @@ private function configurePluginByName($name, Definition $definition, array $con
...
                 $options = $config['config'];
+
                 if (!empty($options['cache_key_generator'])) {
                     $options['cache_key_generator'] = new Reference($options['cache_key_generator']);
                 }
 
+                if (!empty($options['cache_listeners'])) {
+                    foreach ($options['cache_listeners'] as $i => $listener) {
+                        if (!empty($listener)) {
+                            $options['cache_listeners'][$i] = new Definition($listener);
```
dbu: this is assuming that the listener classes have empty constructors. i think we should define them to be service names, and then we can create new `Reference` here instead.

because of bundle loading order, we can't assume that the service already exists at this spot. the user will get an error later when a referenced service does not exist.

theoretically we could do a compiler pass (in a step where the definitions are supposed to all exist) to check the services for implementing the interface, but i don't think we need to go to that effort.

src/DependencyInjection/Configuration.php
nobYsDarling: This has been exactly my pov. I replaced it with a list of classes since this lines up with the documentation of the cache plugin.

dbu: which documentation says it should be class names? the phpdoc on the cache plugin class says its a list of cache listener instances:

https://github.com/php-http/cache-plugin/blob/15883781f0b62b2da6b399675bab9f994a5dd1c7/src/CachePlugin.php#L68

src/DependencyInjection/HttplugExtension.php
nobYsDarling: Switched form list of classes to list of service ids.

Removed the checks for correct implementation of given interface.

src/DependencyInjection/Configuration.php
nobYsDarling: Mhmm... didn't find it again... I might have been mistaken. You are right :-)

src/DependencyInjection/Configuration.php
nobYsDarling: done

nobYsDarling:
Travis CI failed in one case due to timeout of 300ms

src/DependencyInjection/Configuration.php
```diff
@@ -739,6 +740,12 @@ private function createCachePluginNode()
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('cache_listeners')
+                    ->info('A list of service ids to act on the response based on the results of the cache check. Must implement '.CacheListener::class.'. Defaults to an empty array.')
+                    ->beforeNormalization()->ifNull()->thenEmptyArray()->castToArray()->ifEmpty()->thenUnset()->end()
```
dbu: i would rather not unset it but have an empty array as default. that would make the extension class easier.

tests/Resources/Fixtures/config/full.xml
```diff
@@ -54,7 +54,7 @@
                 <my_service type="service" service="my_auth_service"/>
             </authentication>
             <cache cache-pool="my_cache_pool" stream-factory="my_other_stream_factory">
-                <config default-ttl="42" cache-lifetime="2592000" hash-algo="sha1" cache-key-generator="null">
+                <config default-ttl="42" cache-lifetime="2592000" hash-algo="sha1" cache-key-generator="null" cache-listeners="">
```
dbu: that won't work. cache-listener needs to be a xml list, as in 
```
<config ...>
    <cache-listener>service-name</cache-listener>
    <cache-listener>other-service-name</cache-listener>
    ....
</config>
```

i'd prefer to have the test with 2 listeners so that we see it works as expected.

src/DependencyInjection/HttplugExtension.php
```diff
@@ -203,10 +203,22 @@ private function configurePluginByName($name, Definition $definition, array $con
...
             case 'cache':
                 $options = $config['config'];
+
                 if (!empty($options['cache_key_generator'])) {
                     $options['cache_key_generator'] = new Reference($options['cache_key_generator']);
                 }
 
+                if (!empty($options['cache_listeners'])) {
+                    foreach ($options['cache_listeners'] as $i => $listener) {
+                        if (!empty($listener)) {
```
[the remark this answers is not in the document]
nobYsDarling: nice.

src/DependencyInjection/Configuration.php
nobYsDarling: done.

tests/Resources/Fixtures/config/full.xml
nobYsDarling: ack

src/DependencyInjection/Configuration.php
dbu: did you forget to push? there are 2 entries for cache_listeners in the configuration, and both do `ifEmpty()->thenUnset()`.

tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -266,6 +267,7 @@ public function testSupportsAllConfigFormats(): void
                         'cache_key_generator' => null,
                         'respect_response_cache_directives' => ['X-Foo'],
                         'blacklisted_paths' => ['@/path/not-to-be/cached@'],
+                        'cache_listeners' => [],
```
dbu: can we do the full example with 2 cache listeners configured?

tests/Resources/Fixtures/config/full.xml
```diff
@@ -62,6 +62,7 @@
             </authentication>
             <cache cache-pool="my_cache_pool" stream-factory="my_other_stream_factory">
                 <config default-ttl="42" cache-lifetime="2592000" hash-algo="sha1" cache-key-generator="null">
+                    <cache-listeners></cache-listeners>
```
dbu: afaik this should be <cache-listener>, symfony configuration handles pluralization for xml. but lets do the full test with 2 listeners instead of 0.

tests/Unit/DependencyInjection/ConfigurationTest.php
nobYsDarling: y0... I am kind of eager, if this is, what you expected. wörks for me so far.

src/DependencyInjection/Configuration.php
nobYsDarling: fixed

src/DependencyInjection/Configuration.php
```diff
@@ -768,6 +769,12 @@ private function createCachePluginNode()
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('cache_listeners')
```
dbu: at the top of this configuration tree, below the line `fixXmlConfig('respect_response_cache_directive')` we need a fixXmlConfig for 'cache_listeners'.

tests/Resources/Fixtures/config/full.xml
```diff
@@ -62,6 +62,8 @@
             </authentication>
             <cache cache-pool="my_cache_pool" stream-factory="my_other_stream_factory">
                 <config default-ttl="42" cache-lifetime="2592000" hash-algo="sha1" cache-key-generator="null">
+                    <cache-listeners>my_cache_listener_0</cache-listeners>
+                    <cache-listeners>my_cache_listener_1</cache-listeners>
```
dbu: when you do the fixXmlConfig, this will be `<cache-listener>` (which is the way how proper xml should look, according to symfony)

src/DependencyInjection/Configuration.php
```diff
@@ -768,6 +769,12 @@ private function createCachePluginNode()
...
+                    ->info('A list of service ids to act on the response based on the results of the cache check. Must implement '.CacheListener::class.'. Defaults to an empty array.')
+                    ->beforeNormalization()->ifNull()->thenEmptyArray()->castToArray()->ifEmpty()->thenUnset()->end()
```
dbu: i am still confused about `ifEmpty()->thenUnset()`. could it be that this will break in the extension if i configure the cache plugin but no listeners? although the "empty" configuration test seems to see the array, but then i feel like that unset does not do anything ever?

src/DependencyInjection/Configuration.php
nobYsDarling: done.

tests/Resources/Fixtures/config/full.xml
nobYsDarling: done.

src/DependencyInjection/Configuration.php
nobYsDarling: uhmmm... yea, this line is kind of f***ed up. ended up into just casting it to array, which should be fine.

dbu:
thank you!

PR #364, opened 2019-11-28
Test on PHP74

Nyholm:
Let's see if Travis has updated their list of runtimes =)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,9 @@ matrix:
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak"
 
           # Test the latest stable release
-        - php: 7.4
         - php: 7.2
         - php: 7.3
+        - php: 7.4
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ matrix:
           env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak"
 
           # Test the latest stable release
+        - php: 7.4
```
dbu: should be sorted in increasing order, and 7.4 the thing with coverage and whatnot.

dbu:
@Nyholm oh look, it works, and no bad surprises. can you adjust the .travis.yml thing?

dbu:
hm, why does coverage time out with 7.4? maybe we should run the coverage build with 7.3 after all - i guess its very unlikely that this is an issue in our code.

dbu:
the dev build should also use 7.4 btw

dbu:
fixed in #369

PR #365, opened 2019-12-05
Update composer.json

dbu:
increase symfony version requirements because older versions have security warnings.


dbu:
@php-http/httplug should we do these? if we don't there are security warnings on the repo because we allow vulnerable versions of symfony components. not sure if it is our job to do this - on the other hand if people realize thanks to this that they should be upgrading their vulnerable components, it has value.

dbu:
thanks!

PR #366, opened 2019-12-08
URLify header value links

ostrolucky:
Reference: https://github.com/php-http/HttplugBundle/pull/355#issuecomment-555062832

### DIFF
--- a/src/Resources/views/http_message.html.twig
+++ b/src/Resources/views/http_message.html.twig
@@ -17,9 +17,16 @@
             {% set content = content ~ "\n" ~ row %}
         {% else %}
             {% set row = row|split(':') %}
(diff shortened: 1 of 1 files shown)


ostrolucky:
Damit actually I forgot to test this with https and mistakenly used ?s instead of s?

dbu:
oops, also overlooked that one. at least it works if you do httpXs :grin: 

i guess adding automated tests for this is quite complicated, not sure its worth it.

dbu:
are you going to do a follow-up pull request to correct it?

ostrolucky:
Yes, in couple days perhaps

dbu:
i made #368 - that is all we need, or am i missing something?

PR #368, opened 2019-12-10
Update http_message.html.twig

dbu:
fix minor bug with #366 so that we actually match https links


ostrolucky:
Yeah this is all we need, but perhaps we don't even need regex group, but that's minor

dbu:
yep thanks, you are right the `()` is not necessary, checked with a regex checker to be sure ;-)

PR #372, opened 2019-12-26
Allow to register PSR17 and PSR18 classes as services

Nyholm:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

We need a way to support PSR17 factories. Our CurlClient 2.0 does not use the deprecated php-http/message-factory.


src/Resources/config/services.xml
```diff
@@ -55,8 +80,8 @@
             <argument type="service" id="httplug.message_factory"/>
         </service>
         <service id="httplug.factory.curl" class="Http\HttplugBundle\ClientFactory\CurlFactory" public="false">
-            <argument type="service" id="httplug.message_factory"/>
-            <argument type="service" id="httplug.stream_factory"/>
+            <argument type="service" id="httplug.psr17_response_factory"/>
+            <argument type="service" id="httplug.psr17_stream_factory"/>
```
dbu: won't this lead to regressions for users that have no psr17 implementation available?

the factories throw exception when no implementation is present. should we have a compiler pass that checks if the factories find anything and removes the services if they don't?

Nyholm: That is very true. I have to fix that.

Nyholm: I just tested in Symfony 5. 

It is not true, we are only using the factory if we are actually trying to use the service. So this is fine.

dbu: but are we only trying to use the service when a psr-17 implementation is available?

Nyholm: Ie, the curl client is using psr17. It is currently broken. 

So if you use the httplug.factory.curl, then we will use these psr17 factory services. 
There is nothing that is using these services out of the box.

PR #375, opened 2019-12-27
Declare strict types

gmponos:
I could also add it as a php-cs rule but checking travis I saw that you don't check the code against php-cs


gmponos:
Ah!.. you use pretty ci... ok.. so shall I add it? if I do I will need to enable risky rules.

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.phpunit.result.cache
```
gmponos: Added this also on this PR.

gmponos:
Done.

src/DependencyInjection/Configuration.php
```diff
@@ -270,222 +264,219 @@ private function createClientPluginNode()
...
-                            ->end()
-                        ->end()
-                    ->end() // End naming_strategy_options
-                    ->scalarNode('fixtures_directory')
-                        ->info('Where the responses will be stored and replay from when using the filesystem recorder. Should be accessible to your VCS.')
-                    ->end()
-                ->end()
-            ->end()
-        ->end();
+            ->arrayNode('reference')
```
Nyholm: Oh,, I do not like these changes. I liked the tree structure

gmponos: Wait.. I see more changes that I didn't do... :/

gmponos: OK. these changes are made from php-cs-fixer... :/ 

Maybe allowing strict rules enables some rules of symfony that were not previously enabled.. :/

I would suggest not to add the rule on php-cs

gmponos:
> What are the benefits of having declare(strict_types=1)?

I guess today none.. but why not having it?

src/DependencyInjection/Configuration.php
dbu: can you revert the whitespace problem, push and see if pretty-ci wants to destroy it again?

Nyholm:
> but why not having it?

Because if we made one small mistake, everything will break horribly. =)

src/DependencyInjection/Configuration.php
gmponos: Pretty ci passes.. another reason that I am thinking is that ci is using older php-cs version.

would be nice if the php-cs could be included as a phar in the project for these kind of reasons.. developers will be able to contribute easier.

gmponos:
> Because if we made one small mistake, everything will break horribly. =)

Ah.. I got to admit that you have a point and better safe than sorry.. then may I suggest in order not to get all the work lost.. to leave it for tests and interfaces? or you want to disable except if you want to have a php-cs rule for having it removed.

dbu:
i would be for having the strict types. if we do have mistakes, people will tell us, and we can fix them. more likely, other people have mistakes and can notice and fix them. non-strictness essentially hides issues. i would see this in a new minor version, not as a patch release.

most code in the bundle is not interacted with by the user, but is about the collector and factories and such.

src/Collector/Collector.php
```diff
@@ -1,10 +1,13 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Http\HttplugBundle\Collector;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+use Symfony\Component\HttpKernel\Kernel;
```
dbu: this line should not be added, its an unused import

dbu:
@Nyholm are you ok if we merge this? afaik it should only cause BC breaks if we have programming errors in our code, and if we do we rather know and fix them than ignoring the problem.

Nyholm:
Fair. @gmponos, can you make sure the ci is green

dbu:
also needs a rebase on master again, to solve conflicts :-/

PR #376, opened 2019-12-27
Added an sf4 example for enabling the bundle

gmponos:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Documentation   | Yes
| License         | MIT

#### What's in this PR?

Update the README file on how to register/enable the bundle.

TBH it might be preferable to point to an external resource.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,19 @@ HTTPlug works with any HTTP client implementation that provides a HTTPlug adapte
 
 Enable the bundle in your kernel:
 
+**Symfony 4 example:**
+
+``` php
+// config/bundles.php
+<?php
+return [
+    ...
+    Http\HttplugBundle\HttplugBundle::class => ['all' => true],
+];
+```
+
+**Symfony 3 example:**
+
 ``` php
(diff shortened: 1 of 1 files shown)


dbu:
thanks. this is better than explaining the symfony 3 way.

we have https://php-http.readthedocs.io/en/latest/integrations/symfony-bundle.html which seems like the right external source. can you change the readme to only explain the flex way, and link to the documentation for the installation without flex? and do this PR against the documentation to explain the symfony 4/5 way of things?

dbu:
see #380 and https://github.com/php-http/documentation/pull/275

dbu:
thanks for bringing this up @gmponos

gmponos:
arggg.. my mind always forgets things like documentations and stuff.. totally forgot to update this PR.

Anyway glad it helped in anyway...

dbu:
no worries, was quick to wrap it up, thanks for having realized it in the first place!

PR #379, opened 2020-01-01
replace the include tag with the include() function

xabbuh:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   |
| License         | MIT

#### What's in this PR?

Replaces usages of the Twig `include` tag with the `include()` function.

#### Why?

Twig best practice is to use the function instead of the tag:

> It is recommended to use the include function instead as it provides the same features with a bit more flexibility [...]

see https://twig.symfony.com/doc/3.x/tags/include.html

### DIFF
--- a/src/Resources/views/stack.html.twig
+++ b/src/Resources/views/stack.html.twig
@@ -34,10 +34,10 @@
     </div>
     <div class="httplug-messages">
         <div class="httplug-message card">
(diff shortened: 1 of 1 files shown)


dbu:
thanks!

PR #388, opened 2020-10-14
Split listener from ConfiguredClientsStrategy

wouterj:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | technically speaking yes (but I would assume this class should be considered internal?)
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

This splits the listener from the `ConfiguredClientsStrategy`.

#### Why?

The listener only executes a static call. This avoids initializing the dependencies (including HttpClient) on each request/command call.

#### Checklist

Not sure if any of these checks should be done for this PR?

### DIFF
--- a/src/Discovery/ConfiguredClientsStrategy.php
+++ b/src/Discovery/ConfiguredClientsStrategy.php
@@ -8,16 +8,6 @@ use Http\Client\HttpAsyncClient;
 use Http\Client\HttpClient;
 use Http\Discovery\HttpClientDiscovery;
 use Http\Discovery\Strategy\DiscoveryStrategy;
-use Symfony\Component\EventDispatcher\Event as LegacyEvent;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpKernel\Kernel;
-use Symfony\Contracts\EventDispatcher\Event;
-
-if (Kernel::MAJOR_VERSION >= 5) {
-    \class_alias(Event::class, 'Http\HttplugBundle\Discovery\ConfiguredClientsStrategyEventClass');
-} else {
-    \class_alias(LegacyEvent::class, 'Http\HttplugBundle\Discovery\ConfiguredClientsStrategyEventClass');
-}
 
 /**
  * A strategy that provide clients configured with HTTPlug bundle. With help from this strategy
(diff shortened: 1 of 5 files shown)


src/Discovery/ConfiguredClientsStrategy.php
```diff
@@ -8,24 +8,14 @@
...
-use Symfony\Component\EventDispatcher\Event as LegacyEvent;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpKernel\Kernel;
-use Symfony\Contracts\EventDispatcher\Event;
-
-if (Kernel::MAJOR_VERSION >= 5) {
-    \class_alias(Event::class, 'Http\HttplugBundle\Discovery\ConfiguredClientsStrategyEventClass');
-} else {
-    \class_alias(LegacyEvent::class, 'Http\HttplugBundle\Discovery\ConfiguredClientsStrategyEventClass');
-}
```
dbu, line 20: this was unused, correct?

dbu (approved):
i think this makes sense, and people should not be using these classes.

i would still do a changelog entry to mention why we do this (avoid instantiating the strategy on every request). the changelog will also help if somebody hacked weird things in the service definition or such and wonders why things stopped working.

src/Discovery/ConfiguredClientsStrategy.php
wouterj: Yes, I've removed the `ConfiguredClientsStrategyEventClass $e` argument from the `onEvent()` method. The event was not used, so I figured the easiest method is to remove this "bc hack".

wouterj:
CHANGELOG entry added.

dbu:
thanks a lot!

dbu:
thanks, tagged 1.19.0

wouterj:
Thanks for the merge & quick release!

dbu:
hi @blackarcanis , do you redefine the services of this bundle somehow? are you sure that the container has been rebuilt for the env in which you encounter the error? if this is with env=prod, can you try `bin/console --env=prod cache:clear` ?

blackarcanis:
@dbu Hum yes and only got this issue in our production. Seems to be due to our version of Sentry v2.3.0... 
I'm looking at this.

Ziyann:
@blackarcanis, I had to manually remove the old cache on production to fix that issue:
```rm -rf var/cache/prod && bin/console cache:clear```

dbu:
do you deploy to a new directory and build the cache, or simply overwrite the live code and then run `bin/console cache:clear` on an already warmed up cache? if the later, then i think Ziyann's suggestion is a good solution. (and if you are concious about uptime of your site, i'd recommend looking into deployment strategies with symlinks and stuff, but that is kinda off topic ;-) )

PR #389, opened 2020-11-27
Allow PHP 8

ddeboer:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

- Allow HttplugBundle to be installed on PHP 8.
- Remove support for guzzle6-adapter; support guzzle7-adapter instead.

#### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,11 +45,11 @@ matrix:
 
           # Test with httplug 1.x clients
         - php: 7.3
-          env: DEPENDENCIES="php-http/buzz-adapter:^1.0 php-http/guzzle6-adapter:^1.1.1 php-http/react-adapter:^0.2.1 php-http/socket-client:^1.0"
+          env: DEPENDENCIES="php-http/buzz-adapter:^1.0 php-http/react-adapter:^0.2.1 php-http/socket-client:^1.0"
 
           # Test with httplug 2.x clients
         - php: 7.3
-          env: DEPENDENCIES="php-http/guzzle6-adapter:^2.0.1 php-http/curl-client:^2.0.0 php-http/vcr-plugin:^1.0@dev"
+          env: DEPENDENCIES="php-http/guzzle7-adapter php-http/curl-client:^2.0.0 php-http/vcr-plugin:^1.0@dev"
 
           # Latest commit to master
         - php: 7.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
         - php: 7.3
           env: COVERAGE=true TEST_COMMAND="composer test-ci" DEPENDENCIES="php-http/vcr-plugin:^1.0@dev"
         - php: 7.4
-        - php: 8.0snapshot
+        - php: 8.0
 
           # Test LTS versions
         - php: 7.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,6 @@ before_install:
 
 install:
     # To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
-    - if [[ "$COMPOSER_FLAGS" == *"--prefer-lowest"* ]]; then composer update --prefer-dist --no-interaction --prefer-stable --quiet; fi
     - COMPOSER_MEMORY_LIMIT=-1 composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
     - vendor/bin/simple-phpunit install
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,8 @@ matrix:
           # Latest commit to master
         - php: 7.4
           env: STABILITY="dev" DEPENDENCIES="php-http/vcr-plugin:^1.0@dev"
+        - php: 8.0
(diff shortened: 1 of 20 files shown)


dbu:
do we need to build with an RC of php 8? 8.0 is not found in the ci build :-/

dbu:
and thanks for looking into this! i am excited for php 8!

ddeboer:
Whoa, suddenly we have all new kinds of failures. Due to Symfony 5.2?

dbu:
i think that is a symfony upgrade, yes:

```
Failed asserting that exception message 'The child config "cache_pool" under "httplug.clients.test.plugins.0.cache" must be configured: This must be a service id to a service implementing Psr\Cache\CacheItemPoolInterface' contains 'The child node "cache_pool" at path "httplug.clients.test.plugins.0.cache" must be configured.'.
```

i think the wording changed and we have assertions on that. can you adjust the assertion? i guess we should just look for `httplug.clients.test.plugins.0.cache` or something, otherwise the lowest version build will fail...

ddeboer:
Depends on https://github.com/php-http/stopwatch-plugin/pull/16.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+# 1.20.0 - to be released
+
+- Support PHP 8.
+- Support Guzzle 7.
+- BC BREAK: Remove support for Guzzle 6.
```
ddeboer, line 9: Let me know if you agree with this.

dbu: looking at the configuration files, we seem to still support guzzle 5 - if that still works, i suggest we don't replace guzzle 6 with 7, but just add 7. or is that not possible?

ddeboer: Yes, that’s definitely possible. Alle Guzzles >5 allow PHP 8 (they have `php: >=5.4.0` or something similar).

While it’s easiest if PHP 8 users just switch to the latest Guzzle’s HTTPlug adapter, I agree that it may ease maintenance if we don’t introduce a BC break here. So let me revert this change and see if this bundle can continue to support Guzzle 6.

dbu: i would expect guzzle 6 to not work with php 8, but it should continue to work on older versions. our guzzle6-adapter is not installable in php 8 anyways.

ddeboer:
Most jobs are green now. The remaining ones have to do with testing against guzzle6-adapter (by adding it to `DEPENDENCIES`), which tries to install Guzzle 6, which obviously conflicts with Guzzle 7 required by the default guzzle7-adapter.

dbu:
i suggest we switch all tests to guzzle7-adapter then. we don't test the other integrations anyways. and guzzle7 should work on all php versions we allow.

ddeboer:
Unfortunately guzzle7-adapter requires HTTPlug ^2.0, so now the test against HTTPlug 1.0 is [failing](https://travis-ci.com/github/php-http/HttplugBundle/jobs/452494797). How important is it to keep that test?

dbu:
it would be nice to know it still works. can we put the guzzle7-adapter version into a variable? maybe a global variable that says guzzle7 and then we overwrite it in the build where we need guzzle6-adapter?

ddeboer:
@dbu Only two failing test cases left. Do you have any idea what‘s going on there?

.travis.yml
```diff
@@ -22,14 +23,14 @@ matrix:
     fast_finish: true
     include:
           # Minimum supported Symfony version and lowest PHP version
-        - php: 7.2
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak"
+        - php: 7.3
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest" SYMFONY_DEPRECATIONS_HELPER="weak" DEPENDENCIES="php-http/guzzle7-adapter:^0.1.1"
```
dbu, line 27: i suggest to use guzzle6-adapter for the lowest version build. i think we only added guzzle7 to discovery later - but we should not raise the minimum discovery component requirement, it does work together (just not with guzzle7)

dbu:
almost, almost.
is it possible to install guzzle6 adapter along with guzzle7-adapter? if not, we should skip the Http\HttplugBundle\Tests\Functional\DiscoveredClientsTest::testForcedDiscovery in lowest version build (check if guzzle7-adapter is available and if not, skip)

.travis.yml
ddeboer: Ah, thanks for the hint!

ddeboer:
The weird thing is that the failing builds are installing guzzle7-adapter, which is not in `composer.json` nor a transitive dependency.

ddeboer:
All green, thanks for your help!

dbu (approved):
great, thanks!

dbu:
released as https://github.com/php-http/HttplugBundle/releases/tag/1.20.0

PR #393, opened 2021-07-05
Register client alias to support Symfony 5.3 Target attribute

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| Documentation   | if this is a new feature, link to pull request in https://github.com/php-http/documentation that adds relevant documentation
| License         | MIT

#### What's in this PR?

When you are using Symfony 5.3 and you have multiple clients defined, you can use the `Target`
attribute to select the client that you want.

#### Why?

Improve Developer Experience.

#### Example Usage

Example:
```php
final class MyService
{
    public function __construct(
        #[Target('my_client_alias')] 
        HttpClient $client
    ) {}
}
```

It will now automatically inject `httplug.client.my_client_alias`.

For more information:
https://symfony.com/blog/new-in-symfony-5-3-service-autowiring-with-attributes

#### Checklist

- [x] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
- [ ] Documentation pull request created (if not simply a bugfix)

PR #394, opened 2021-07-12
Use `ClientInterface` instead of `HttpClient` interface when registering aliases

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   |yes
| Related tickets | 
| License         | MIT

#### What's in this PR?

After https://github.com/php-http/HttplugBundle/pull/393 was merged I started using it in my project and soon discovered that `HttpClient` is deprecated and that the `ClientInterface` from Psr should be used instead.

#### Why?

This will deprecate the alias registration for HttpClient when the ClientInterface is found. 

It will also register the alias for the ClientInterface.

### DIFF
--- a/src/DependencyInjection/HttplugExtension.php
+++ b/src/DependencyInjection/HttplugExtension.php
@@ -383,8 +383,10 @@ class HttplugExtension extends Extension
         $serviceId = 'httplug.client.'.$clientName;
 
         if (method_exists($container, 'registerAliasForArgument')) {
-            $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
+            $alias = $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
             if (interface_exists(ClientInterface::class)) {
+                $alias->setDeprecated('php-http/httplug-bundle', '1.22', 'The "%alias_id%" alias is deprecated, use "Psr\Http\Client\ClientInterface" instead.');
+
                 $container->registerAliasForArgument($serviceId, ClientInterface::class, $clientName);
             }
         }
--- a/src/DependencyInjection/HttplugExtension.php
+++ b/src/DependencyInjection/HttplugExtension.php
@@ -384,7 +384,9 @@ class HttplugExtension extends Extension
 
         if (method_exists($container, 'registerAliasForArgument')) {
             $alias = $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
-            if (interface_exists(ClientInterface::class)) {
+
+            $interfaces = class_implements(HttpClient::class) ?? [];
+            if (isset($interfaces[ClientInterface::class])) {
                 $alias->setDeprecated('php-http/httplug-bundle', '1.22', 'The "%alias_id%" alias is deprecated, use "Psr\Http\Client\ClientInterface" instead.');
 
                 $container->registerAliasForArgument($serviceId, ClientInterface::class, $clientName);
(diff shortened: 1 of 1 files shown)


src/DependencyInjection/HttplugExtension.php
```diff
@@ -382,7 +383,12 @@ private function configureClient(ContainerBuilder $container, $clientName, array
         $serviceId = 'httplug.client.'.$clientName;
 
         if (method_exists($container, 'registerAliasForArgument')) {
-            $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
+            $alias = $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
+            if (interface_exists(ClientInterface::class)) {
```
dbu, line 387: i think the existence of the psr interface is not the correct way to know which version of httplug we are using. we should additionally check if the httplug client implements the psr interface or something else that tells us if we are installed with a httplug version that implements the interface.

dbu (changes requested):
thanks for the update and sorry for the late reply - i was on holidays last week.

i agree we should do this. however, i am a bit worried about creating false deprecations and wiring the psr interface if somebody manages to install this change along with the psr client but an old version of httplug.

src/DependencyInjection/HttplugExtension.php
ruudk: Does https://github.com/php-http/HttplugBundle/pull/394/commits/0006ec2294224a8fa4ec34b113ab93ddb0bf42fe solve it?

dbu (approved):
thanks, this looks robust to me!

dbu:
released as  https://github.com/php-http/HttplugBundle/releases/tag/1.22.0

src/DependencyInjection/HttplugExtension.php
```diff
@@ -382,7 +383,14 @@ private function configureClient(ContainerBuilder $container, $clientName, array
         $serviceId = 'httplug.client.'.$clientName;
 
         if (method_exists($container, 'registerAliasForArgument')) {
-            $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
+            $alias = $container->registerAliasForArgument($serviceId, HttpClient::class, $clientName);
+
+            $interfaces = class_implements(HttpClient::class) ?? [];
+            if (isset($interfaces[ClientInterface::class])) {
+                $alias->setDeprecated('php-http/httplug-bundle', '1.22', 'The "%alias_id%" alias is deprecated, use "Psr\Http\Client\ClientInterface" instead.');
```
dbu, line 390: @ruudk somebody reported that symfony complains about the deprecation warning, saying it needs an `%alias_id%` (though i see that here)... https://twitter.com/PP_Vortex/status/1419599426084032519

can you please check if the deprecation warning works correctly for you and there is no symfony warning about alias_id?

franziskahahn: I did some deep dive into vendors and found out the following fact:

the method `setDeprecated` in Alias.php changed in symfony 5.3 compared to symfony 4.4:
https://github.com/symfony/dependency-injection/blob/5.3/Alias.php
https://github.com/symfony/dependency-injection/blob/4.4/Alias.php


In symfony 4.4, the second parameter is the template. So when using http plug bundle with symfony 4.4 executing _assets:install_ interpretes this line as "deprecation template is '1.22'" and this string obviously doesn't contain %alias_id%. 


**My conclusion: Currently, php-http/httplug-bundle 1.22 is not compatible with symfony 4.4 but seems to be compatible with symfony 5.3** 

What would be the next steps here? Would you like to be compatible with both versions?

ruudk: I'll check it out and make a PR.

ruudk: See https://github.com/php-http/HttplugBundle/pull/396

dbu: thanks for reporting the issue and for the analysis @franziskahahn. the deprecation is removed in https://github.com/php-http/HttplugBundle/releases/tag/1.22.1

PR #396, opened 2021-07-26
Do not deprecate alias as it changed a lot in different Symfony versions

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | see #394
| Documentation   | 
| License         | MIT

#### What's in this PR?

Turns out that `Alias::setDeprecated` changed a lot between Symfony versions. I couldn't find a way to distinguish between them so decided to just remove the deprecation for now.

### DIFF
--- a/src/DependencyInjection/HttplugExtension.php
+++ b/src/DependencyInjection/HttplugExtension.php
@@ -387,8 +387,6 @@ class HttplugExtension extends Extension
 
             $interfaces = class_implements(HttpClient::class) ?? [];
             if (isset($interfaces[ClientInterface::class])) {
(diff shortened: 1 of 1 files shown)


dbu (approved):
yeah that makes sense to me. not worth it to do a complicated thing to support all various symfony versions.

we will eventually deprecate the httplug interface itself i think, and that should be enough warning. while the interface exists, we will want to keep autowiring it working anyways.

dbu:
thanks for the quick fix. i tagged https://github.com/php-http/HttplugBundle/releases/tag/1.22.1

PR #397, opened 2021-08-27
[Profiler] Utilize symfony/var-dumper for dumping response body 

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | finishes #245
| Documentation   | 
| License         | MIT

#### What's in this PR?

Similarly like `symfony/http-client` and other bundles, this utilizes `symfony/var-dumper` to format request/response body messages.

Json detection was partially taken over from [csa/guzzle-bundle](https://github.com/csarrazi/CsaGuzzleBundle/blob/master/src/Twig/Extension/GuzzleExtension.php#L52)

This results in much improved design of body formatting, replacing plain nl2br that was used till now.

image

#### Checklist

- [x] CI https://app.travis-ci.com/github/ostrolucky/HttplugBundle/builds/236415878

### DIFF
--- a/src/Resources/views/http_message.html.twig
+++ b/src/Resources/views/http_message.html.twig
@@ -14,7 +14,7 @@
         {% if row is empty %}
             {% set hasReachedBody = true %}
         {% elseif hasReachedBody %}
-            {% set content = content ~ "\n" ~ row %}
+            {% set content = content ~ row %}
         {% else %}
             {% set row = row|split(':') %}
             {% set value = row|slice(1)|join(':')|trim %}
@@ -33,4 +33,4 @@
     </tbody>
 </table>
 
-<div class='httplug-http-body httplug-hidden'>{{ content|nl2br ?: '(This message has no captured body)' }}</div>
+<div class='httplug-http-body httplug-hidden'>{{ content|httplug_markup_body }}</div>
--- a/src/Resources/views/http_message.html.twig
+++ b/src/Resources/views/http_message.html.twig
@@ -33,4 +33,4 @@
     </tbody>
 </table>
 
-<div class='httplug-http-body httplug-hidden'>{{ content|httplug_markup_body }}</div>
+<div class='httplug-http-body httplug-hidden'>{{ content ? content|httplug_markup_body : '(This message has no captured body)' }}</div>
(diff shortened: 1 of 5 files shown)


src/Resources/views/http_message.html.twig
```diff
@@ -33,4 +33,4 @@
     </tbody>
 </table>
 
-<div class='httplug-http-body httplug-hidden'>{{ content|nl2br ?: '(This message has no captured body)' }}</div>
+<div class='httplug-http-body httplug-hidden'>{{ content|httplug_markup_body }}</div>
```
dbu, line 36: should we keep the alternate text for when there is no body around?

dbu (commented):
very nice, thanks!

src/Resources/views/http_message.html.twig
ostrolucky: If there is no body, vardumper will show this, which is also representative enough, isn't it?
image

But in case you think it's better to have old placeholder back let me know.

dbu: i think depending on configuration, we do not record the body. it would make sense to not lead people to think that there was an empty body when actually the body was just not recorded. but i notice that we currently don't distinguish between "we recorded that there was an empty body" and "we did not record the body". i would prefer to bring back the message we had as that did not claim its one or the other. we could further improve things to distinguish between empty body and not recorded body, but thats out of scope for this pull request.

ostrolucky: Ok, done!

dbu:
thanks! 

do you plan to work on fixing the ci setup of the bundle? (i guess we should port it to github actions like we did for other php-http repositories...) while this looks good to me, i'd prefer to see the CI running through before merging ;-)

ostrolucky:
CI is running, see https://app.travis-ci.com/github/ostrolucky/HttplugBundle/builds. But yes, I plan to move it to GH actions.

dbu:
oh, github showed travis-ci as outstanding. looks good then, i merged the change.

i let you decide when to tag a release ;-)

PR #398, opened 2021-08-30
[Profiler] Always show badge with request count

ostrolucky:
Count here is similarly important like number of queries that doctrine-bundle shows, so it should be also displayed at all times

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Always shows badge counter here instead of only when there are errors

image

#### Why?

This info is always important, similarly like # of queries in ORM - people should try to keep it low

#### Checklist

### DIFF
--- a/src/Resources/views/webprofiler.html.twig
+++ b/src/Resources/views/webprofiler.html.twig
@@ -62,11 +62,9 @@
             {{ include('@Httplug/Icon/httplug.svg') }}
         </span>
         <strong>Httplug</strong>
-        {% if collector.failedStacks|length %}
-            <span class="count">
-                <span>{{ collector.failedStacks|length }}</span>
-            </span>
-        {% endif %}
+        <span class="count">
+            <span>{{ (collector.stacks|length - collector.failedStacks|length) ~ '/' ~ collector.stacks|length}}</span>
+        </span>
     </span>
 {% endblock %}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
-# 1.23.0 - ?
+# 1.23.0 - 2021-08-30
 
 - Changed the way request/response body is displayed in profiler. symfony/var-dumper is used now.
+- Changed badge counter of # of requests on side menu to be always visible
 
 # 1.22.1 - 2021-07-26
 


src/Resources/views/webprofiler.html.twig
```diff
@@ -62,11 +62,9 @@
             {{ include('@Httplug/Icon/httplug.svg') }}
         </span>
         <strong>Httplug</strong>
-        {% if collector.failedStacks|length %}
-            <span class="count">
-                <span>{{ collector.failedStacks|length }}</span>
-            </span>
-        {% endif %}
+        <span class="count">
+            <span>{{ collector.failedStacks|length ?:  collector.stacks|length}}</span>
```
dbu, line 66: isn't this misleading when we show the stacks count or the failed count?

i guess `stacks` implicitly is the successful stacks? if that is correct, i'd show the sum. if it is all, i'd always show the stacks length. or we could do `{{ stacks|length }}{% if failedStacks|length %} (failedStacks|length){%endif%}`

dbu (approved):
good idea!

one feedback on the number we show.

src/Resources/views/webprofiler.html.twig
ostrolucky: No, stacks is all of them, including unsuccessful. Reason I change failedStacks/stacks is because when there is failedStacks, background changes to red - at that point it would be misleading to highlight number representing all of the requests as red, that's why at that point only number of failed requests is made red.

dbu: ah. how about still outputting the total count along with the error count? `2 (6)` or something like that?

dbu: i checked what the doctrine profiler does, but it seems to not mark queries as failed in the pill on the toolbar, so it is different.

ostrolucky: Yeah because it's not acceptible that database query fails, that would be an exception. Most similar to this would be Logs tab.

How about {successful}/{total} number? Eg. 3/3 in case all are successful, or 2/3 in case one failed.

dbu: yeah, i like that one, should be pretty self-explanatory to read

dbu (approved):

dbu:
nice one!

PR #399, opened 2021-08-30
[CI] Migrate to GH actions

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

No coverage for now, can come later


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,100 @@
...
+    env:
+      SYMFONY_REQUIRE: ${{matrix.symfony-require}}
+      SYMFONY_DEPRECATIONS_HELPER: ${{matrix.symfony-deprecations-helper}}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "7.1"
+          - "7.2"
```
dbu, line 23: we only support 7.3 and newer. can you please add 8, however?

dbu: ah sorry, php 8.0 is handled below

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,100 @@
...
+          # Test with httplug 1.x clients
+          - dependencies: "php-http/buzz-adapter:^1.0 php-http/guzzle6-adapter:^1.1.1 php-http/react-adapter:^0.2.1 php-http/socket-client:^1.0"
+            php-version: "7.3"
+          # Test with httplug 2.x clients
+          - dependencies: "php-http/guzzle7-adapter php-http/curl-client:^2.0.0 php-http/vcr-plugin:^1.0@dev"
+            php-version: "7.3"
+
+          # Latest commit to master
+          - dependencies: "php-http/guzzle7-adapter php-http/vcr-plugin:^1.0@dev"
+            php-version: "7.0"
```
dbu, line 58: 8.0?

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,99 @@
...
+            stability: "dev"
+
+          # Test maintained versions of Symfony
+          - symfony-require: "3.4.*"
+            php-version: "7.3"
+          - dependencies: "php-http/guzzle7-adapter symfony/http-client:^4.4"
+            symfony-require: "4.4.*"
+            php-version: "7.3"
+          - dependencies: "php-http/guzzle7-adapter symfony/http-client:^5.0"
+            symfony-require: "5.0.*"
```
dbu, line 69: i guess this comes from the previous config, but current would be 5.3.* ... 

but lets port the setup first, without introducing additional changes.

dbu (commented):
thanks!

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,19 @@
+
```
sagikazarmark, line 1:
```suggestion
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,99 @@
+
```
sagikazarmark, line 1:
```suggestion
```

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,19 @@
...
+            - master
+    pull_request:
+
+jobs:
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: "Checkout"
```
sagikazarmark, line 16: I'd suggest using two spaces everywhere. I usually add a .editorconfig file to the .github directory to achieve that. Here is an example: https://github.com/twirphp/twirp/blob/master/.github/.editorconfig

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,19 @@
+
+name: "Coding Standards"
```
sagikazarmark, line 2: I'd rather call this workflow Static analysis as we could stuff in some other tools here later. Example: https://github.com/twirphp/twirp/blob/master/.github/workflows/static.yaml

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,99 @@
...
+name: "Continuous Integration"
```
sagikazarmark, line 2:
```suggestion
name: "CI"
```

The matrix names will be long enough already. Rename the file accordingly as well.

sagikazarmark (changes requested):

sagikazarmark:
Thanks for the PR. Please see my suggestions.

ostrolucky:
Thanks. I would say subjective suggestions, but I've implemented them. Just FYI, I took most of the structure over from https://github.com/doctrine/DoctrineBundle/tree/2.4.x/.github/workflows

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+      - name: "PHP-CS-Fixer"
+        uses: "docker://oskarstark/php-cs-fixer-ga:2.19.0"
```
dbu, line 18: looking at https://github.com/php-http/HttplugBundle/pull/399/checks?check_run_id=3461567716 it seems like something is off. it says that it "fixed" a file, but i guess it should error and report what needs to be fixed rather than silently fix during the run, and then lose the change?

dbu:
we should adjust the README.md as well, to have the badge and link from github actions rather than travis.

the badge url is https://github.com/php-http/HttplugBundle/actions/workflows/continuous-integration.yml/badge.svg and i would link that to https://github.com/php-http/HttplugBundle/actions/

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,98 @@
+name: "CI"
+
+on:
+    push:
+        branches:
+            - master
+    pull_request:
```
sagikazarmark, line 7: These are still four spaces.

sagikazarmark (changes requested):

sagikazarmark:
Thanks for making the changes @ostrolucky ! Yes, I guess they are subjective, but that's how things are implemented in the projects I maintain (Twirp, Guzzle, MoneyPHP) and they worked well for the last couple years, so I prefer keeping them consistent (unless there are reasons not to of course).

ostrolucky:
All good?

dbu (approved):
looks good to me, thanks

dbu:
should we put an editorconfig based on https://github.com/twirphp/twirp/blob/master/.github/.editorconfig into the .github folder? we have those at top level in our company projects and they are useful to help with consistent whitespace (and explicitly define the rules)

ostrolucky:
Rather top level. But separate changeset please

sagikazarmark:
Having a top level editorconfig file is always a good idea and we should already have those in all php-http repos. Having one in .github helps configuring how editors should behave within that folder. I'd say 4 spaces should be the default everywhere, but GitHub workflows are indentated with 2 spaces everywhere (docs, examples), so limiting 2 spaces to .github makes sense.

PR #401, opened 2021-10-04
Fix dark mode design compatibility

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | 
| License         | MIT

#### What's in this PR?

Before:
<table>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>
After:
<table>
<tr>
<td>
image
</td>
<td>
image
</td>
</tr>
</table>

#### Checklist

#### To Do

- Not compatible with Symfony < 4.2. Such Symfony versions are not maintained though. Worth adding conflict rule?

### DIFF
--- a/src/Resources/views/style/httplug.css.twig
+++ b/src/Resources/views/style/httplug.css.twig
@@ -46,7 +46,7 @@
     border-radius: 2px 0px 0px 2px;
     border: 0;
     line-height: inherit;
-    background-color: #eee;
+    background-color: var(--metric-value-background, #eee);
     opacity: 1;
     font-size: 14px;
     flex: 1;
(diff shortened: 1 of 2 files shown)


src/Resources/views/style/httplug.css.twig
```diff
@@ -46,7 +46,7 @@
     border-radius: 2px 0px 0px 2px;
     border: 0;
     line-height: inherit;
-    background-color: #eee;
+    background-color: var(--metric-value-background);
```
dbu, line 49: i am no CSS expert, but if i understand correctly, this would be compatible with Symfony 3.4 if you provide the old value as default:

```suggestion
    background-color: var(--metric-value-background, #eee);
```

dbu (commented):
very nice, thanks!

i would prefer to add the old values as fallback over dropping Symfony 3.4 support for just this reason. if there is a more compelling reason to drop support, i am fine with it, but having the fallback values seems easy enough to me to not need to drop it.

PR #403, opened 2021-10-14
Changed stopwatch category from default to "httplug"

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

I was irked that httplug queries in execution timeline section were grouped with other "default" stuff. It deserves it's own group so it's more prominent.

Maybe not the best example, as http queries here were quick, but you can still see some change

<table><tr><td>
image
</td><td>image</td></tr></table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+# 1.23.2 - unreleased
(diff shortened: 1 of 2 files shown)


dbu (commented):
nice, i agree this makes sense

PR #406, opened 2021-10-21
Fix Twig 3.x compatibility

VirgileZol:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Fix : Unknown "spaceless" tag.

#### Why?

Error 500 in httplug panel

#### Checklist

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         include:
           # Minimum supported Symfony version and lowest PHP version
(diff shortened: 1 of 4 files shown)


dbu:
looking at https://twig.symfony.com/doc/2.x/filters/spaceless.html , this seems to be different for different twig versions :see_no_evil: 

@ostrolucky you did a lot of upgrades on the rendering, do you have an idea how we should solve this? should we extract this into a sub-template and include different versions based on the twig version?

PR #407, opened 2021-11-10
Fix PHP 8.1 deprecation notice

stephanvierkant:
| Q               | A
| --------------- | ---
| Bug fix?        | no|**yes**
| New feature?    | **no**|yes
| BC breaks?      | **no**|yes
| Deprecations?   | **no**|yes
| License         | MIT

#### What's in this PR?

Fixes this deprecation notice in PHP 8.1:

> Deprecated: Optional parameter $cloner declared before required parameter $dumper is implicitly treated as a required parameter in {project}/vendor/php-http/httplug-bundle/src/Collector/Twig/HttpMessageMarkupExtension.php on line 29

#### Checklist

- [x] Updated CHANGELOG.md to describe BC breaks / deprecations | new feature | bugfix
- [x] Documentation pull request created (if not simply a bugfix)

PR #408, opened 2021-11-13
Add PHP 8.1 to CI

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #407
| Documentation   | 
| License         | MIT

#### What's in this PR?

Adds PHP 8.1 to CI. Currently installs RC6.

#### Why?

So that we can verify that it works properly on PHP 8.1.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -46,11 +46,11 @@ jobs:
 
           # Latest commit to master
           - dependencies: "php-http/guzzle7-adapter php-http/vcr-plugin:^1.0@dev"
-            php-version: "8.0"
+            php-version: "8.1"
             stability: "dev"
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -53,6 +53,14 @@ jobs:
             php-version: "8.0"
```
ostrolucky, line 53: Change existing experimental jobs instead, please. We don't need both PHP 8.0 @ dev and PHP 8.1 @ dev

ruudk: Like this?

ostrolucky (approved):

PR #415, opened 2022-04-22
fix: prevent null respect_response_cache_directives

qkdreyer:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

It prevents `respect_response_cache_directives` configuration key to be null, in order to prevent PHP error `array_intersect(): Argument #1 ($array) must be of type array, null given`

#### Why?

image

#### Example Usage

N/A

#### Checklist

#### To Do

- [ ] If the PR is not complete but you want to discuss the approach, list what remains to be done here

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -779,7 +779,7 @@ class Configuration implements ConfigurationInterface
                     ->info('A list of cache directives to respect when caching responses')
                     ->validate()
                         ->always(function ($v) {
-                            if (is_null($v) || is_array($v)) {
+                            if (is_array($v)) {
                                 return $v;
                             }
 


ostrolucky:
Isn't having null there meant as "respect any cache header directives", which is clearly different from "respect no cache header directives"?

dbu:
thanks @qkdreyer for this MR. there is indeed something wrong when we get php warnings.

looking at the [cache plugin code](https://github.com/php-http/cache-plugin/blob/63bc3f7242825c9a817db8f78e4c9703b0c471e2/src/CachePlugin.php#L361) i think we never intended the config to be null. i think having null in the configuration should still be allowed, but lead to the extension not setting the array key at all. then the plugin will take its default behaviour. (inside the plugin it makes sense that the default is a list of caching instructions, but when passing parameters, we should not have a default list, as that would not be future proof).

can you please instead of this change add a null-check to the [cache plugin configuration in the extension](https://github.com/php-http/HttplugBundle/blob/master/src/DependencyInjection/HttplugExtension.php#L199)?

ostrolucky:
As explained, better to update plugin instead. Closing

dbu:
https://github.com/php-http/HttplugBundle/releases/tag/1.26.1

PR #417, opened 2022-05-31
cache_lifetime must be an integer or null

VirgileZol:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Fix
Invalid type for path "httplug.clients.xxx.plugins.xxx.cache.config.cache_lifetime". Expected "int", but got "null".  

cache_lifetime must be an integer or null

#### Example Usage

``` yaml
cache_lifetime: ~
```

#### Checklist

- [x] Updated CHANGELOG.md to describe bugfix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ The change log describes what is "Added", "Removed", "Changed" or "Fixed" betwee
 
 # 1.xx.yy- YYYY-MM-JJ
 
-- Fixed you can now configure the cache plugin cache_lifetime with `null`.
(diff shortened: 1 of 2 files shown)


dbu (commented):
thanks for the contribution, indeed `null` is an accepted value in the plugin.

i realize that the `respect_response_cache_directives` null is somehow inconsistent with the other options - for those where the plugin accepts null as option, we forward the null, which is different from the default value. according to https://github.com/php-http/cache-plugin/blob/master/src/CachePlugin.php#L233 we treat `null` as `0` for the cache_lifetime option.

PR #418, opened 2022-06-21
chore: remove php-http/socket-client<2.0 support

qkdreyer:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

https://github.com/php-http/socket-client/blob/master/src/Client.php#L51

### DIFF
--- a/src/ClientFactory/SocketFactory.php
+++ b/src/ClientFactory/SocketFactory.php
@@ -31,6 +31,6 @@ class SocketFactory implements ClientFactory
             throw new \LogicException('To use the Socket client you need to install the "php-http/socket-client" package.');
         }
 
-        return new Client($this->messageFactory, $config);
+        return new Client($config);
     }
 }
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -37,11 +37,11 @@ jobs:
             symfony-deprecations-helper: "weak"
 
           # Test with httplug 1.x clients
-          - dependencies: "php-http/buzz-adapter:^1.0 php-http/guzzle6-adapter:^1.1.1 php-http/react-adapter:^0.2.1 php-http/socket-client:^1.0"
+          - dependencies: "php-http/buzz-adapter:^1.0 php-http/guzzle6-adapter:^1.1.1 php-http/react-adapter:^0.2.1"
             php-version: "7.3"
             symfony-deprecations-helper: "weak"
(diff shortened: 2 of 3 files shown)


ostrolucky:
This is incomplete. We can pass array there only for >=2.0, so we need to add conflict section

qkdreyer:
fixed the CI

qkdreyer:
I’m not sure how you would like to implement that… using try/catch or reflection ?

src/ClientFactory/SocketFactory.php
```diff
@@ -31,6 +31,6 @@ public function createClient(array $config = [])
             throw new \LogicException('To use the Socket client you need to install the "php-http/socket-client" package.');
         }
 
-        return new Client($this->messageFactory, $config);
+        return new Client($this->messageFactory, $config, $config);
```
dbu, line 34: thanks for looking into this.

i think the actual bug here is in the socket client. the aim was to be BC with version 1, but deprecate passing that factory. the signature of version 1 was correct for how we instantiated the client: https://github.com/php-http/socket-client/blob/v1.4.1/src/Client.php#L52

version 2 should at https://github.com/php-http/socket-client/blob/47e68cade6b76e7fcaaf648048845d9dde0accf9/src/Client.php#L61 should use $config2, not $config which i don't see how it could be useful...

i suggest we fix the socket client to work as intended. we could additionally add a version constant into the socket client to allow the bundle to call the constructor in a future proof way and avoid the deprecation warning.

dbu: @joelwurtz what do you think, would it make sense that way?

dbu:
> I’m not sure how you would like to implement that… using try/catch or reflection ?

you could mark a conflict with socket-client < 2.0 in the composer section. then new versions of this bundle can only be installed with a socket-client version 2, but not version 1.

i think that is the best solution for the bundle. we should also fix the BC logic of the socket client to actually work as intended, but thats a different repo.

dbu:
@qkdreyer do you have time to update this?

qkdreyer:
https://github.com/php-http/socket-client/pull/62

dbu (approved):
thanks a lot!

dbu:
the socket client is now also fixed in version 2.1.1 to have the intended BC with version 1.

PR #419, opened 2022-07-21
Make error plugin configurable

dbrumann:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | https://github.com/php-http/documentation/pull/297
| License         | MIT

#### What's in this PR?

Makes the error plugin configurable in HttplugBundle.

#### Why?

As far as I can tell, enabling the ErrorPlugin via  configuration is currently not easily possible. For applications it can make sense to violate the PSR-18 spec and have this configuration enabled, so adding it to the bundle seems reasonable to me.

#### Example Usage

``` yaml
# config/packages/httplug.yaml
httplug:
    clients:
        my_client:
            factory: 'httplug.factory.guzzle7'
            config:
                connect_timeout: 1
                timeout: 10
                force_ip_resolve: 'v4'
            plugins:
                - error: ~
```

#### Checklist

#### To Do

- [x] Fix failing tests, if it is confirmed this change makes sense

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+# 1.27.0 -
+
+- Added support for configuring options of error plugin via configuration
+
 # 1.26.2 - 2022-06-01
 
 - Fixed: You can now configure the cache plugin option `cache_lifetime` to `null` (which makes the plugin not add to the maxAge).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
-# 1.27.0 -
+# 1.27.0 - 2022-07-25
 
-- Added support for configuring options of error plugin via configuration
+- Added support for configuring the error plugin via configuration
 
 # 1.26.2 - 2022-06-01
 
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -590,6 +590,15 @@ class Configuration implements ConfigurationInterface
             ->end()
         ->end();
         // End stopwatch plugin
+
+        $error = $children->arrayNode('error')
+            ->canBeEnabled()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->scalarNode('only_server_exception')->defaultFalse()->end()
+            ->end()
+        ->end();
+        // End error plugin
     }
 
     /**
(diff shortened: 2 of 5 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -590,6 +590,15 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
             ->end()
         ->end();
         // End stopwatch plugin
+
+        $error = $children->arrayNode('error');
+        $disableAll ? $error->canBeEnabled() : $error->canBeDisabled();
```
dbu, line 595: to avoid behaviour BC breaks, this should always `canBeEnabled`. otherwise people will suddenly unexpectedly get exceptions

dbu (commented):
hi denis, thanks for the merge request. i agree it makes sense to allow to configure the plugin.

the build worked a few days ago, but then i don't see how this change would break the console... do you have time to check what happened? i just re-ran the build on master and its still green: https://github.com/php-http/HttplugBundle/actions/runs/2723047937

and can you please add doc to https://docs.php-http.org/en/latest/integrations/symfony-bundle.html#plugins ?

ostrolucky:
I've tested this change and it doesn't seem to work unfortunately. My changes at https://github.com/php-http/HttplugBundle/compare/master...add-error-plugin-configuration

dbrumann:
@ostrolucky Can you check again? I am not sure, if I changed something in between, but your config and mine looks pretty much identical (other than the check for `canBeEnabled()`/`canBeDisabled()`. Just to be safe we can also close this and use your branch instead.

ostrolucky:
Indeed. Replacing `$disableAll ? $error->canBeEnabled() : $error->canBeDisabled();` with `$error->canBeEnabled();` makes this working

dbu (approved):
thanks, looks good to me. can you please update the CHANGELOG.md to add a 1.27.0 section with a note about this change? release date should be 2022-07-26

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+# 1.27.0 -
```
dbu, line 5:
```suggestion
# 1.27.0 - 2022-07-25
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+
+- Added support for configuring options of error plugin via configuration
```
dbu, line 7:
```suggestion
- Added support for configuring the error plugin via configuration
```

(before you could not even enable the plugin with default options...)

dbu (approved):
thanks! i will tag a release now

dbu:
https://github.com/php-http/HttplugBundle/releases/tag/1.27.0

dbrumann:
Thank you both for your help! :+1:

PR #422, opened 2022-09-16
Profiler: Add slash if there is none between host and path

ostrolucky:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

See 
image
image

Looks like in case target doesn't have leading slash, profiler shows strange URL

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The change log describes what is "Added", "Removed", "Changed" or "Fixed" betwee
 
 - If captured_body_length is set to 0 (default value), show a special message rather than the
   generic message `This message has no captured body`.
+- Fixed: Add slash in profiler if there is none between host and path
 
 # 1.27.0 - 2022-07-25
 
--- a/src/Resources/views/stack.html.twig
+++ b/src/Resources/views/stack.html.twig
@@ -14,7 +14,7 @@
             {% if stack.requestPort not in [null, 80, 443] %}
                 <span class="httplug-port">:{{ stack.requestPort }}</span>
             {% endif %}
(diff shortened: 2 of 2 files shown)


dbu:
that is confusing. isn't that a bug in the getRequestTarget implementation, or the underlying Uri implementation? looking at https://github.com/Nyholm/psr7/blob/f734364e38a876a23be4d906a2a089e1315be18a/src/RequestTrait.php#L27 , i don't see why the leading / would get lost.

i am a bit reluctant to merge this, as i don't understand how it can happen without a (more severe) bug in the underlying request / uri implemenation. (or if it gets lost in the collector / stack, we should fix it there, but i had a look at those classes and don't see them changing the request target, they seem just data container).

also, the trim would mis-edit some weird urls like `//path` with double slash: https://3v4l.org/BR9Lp

ostrolucky:
It's when path of URI is without leading slash. RequestTrait snippet you posted adds leading slash only when path is empty. And we use Guzzle btw, that also behaves this way.

ostrolucky:
> also, the trim would mis-edit some weird urls like //path with double slash: [3v4l.org/BR9Lp](https://3v4l.org/BR9Lp)

I've adjusted code for this now

dbu:
but how can the path be without leading slash?

ah, or does that happen when the uri is created `withPath('no-leading/slash')`, for example when the domain is configured as base on the client?

ostrolucky:
correct

PR #430, opened 2023-05-17
Use ClientInterface in MockFactory

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | ?
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

HttpClient is deprecated in favor of ClientInterface. By accepting ClientInterface in the MockFactory, developers can transition to the ClientInterface.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -44,10 +44,6 @@ jobs:
             php-version: "8.2"
             symfony-deprecations-helper: "weak"
 
-          # Test with httplug 1.x clients
-          - dependencies: "php-http/buzz-adapter:^1.0 php-http/guzzle6-adapter:^1.1.1 php-http/react-adapter:^0.2.1"
-            php-version: "7.3"
-            symfony-deprecations-helper: "weak"
           # Test with httplug 2.x clients
           - dependencies: "php-http/guzzle7-adapter php-http/curl-client:^2.0.0 php-http/vcr-plugin:^1.0@dev php-http/socket-client:^2.0"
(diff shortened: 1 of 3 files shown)


ruudk:
@dbu is this the way to do it?

dbu (approved):
thanks. this should be BC because the httplug client implements the psr interface.

ruudk:
What about the failing test? Do I need to fix more?

dbu:
it mentions buzz. is that the problem? the buzz adapter is quite outdated and i want to deprecate it.

can you try to raise the minimum requirement on php-http/httplug ? if i understand the problem correctly, we can then no longer install with buzz-adapter (because that is pulling us down to such a low version of httplug that we don't implement the psr interface yet). we should raise that minimum requirement anyways to avoid invalid situations (check what version has httplug client interface start extend the psr interface).

if this is indeed the problem, i suggest we remove buzz from that matrix entry (or if thats the only difference to other builds just remove the whole matrix entry)

ruudk:
@dbu done, raised version and removed buzz-adapter from the matrix.

.github/workflows/continuous-integration.yml
```diff
@@ -44,10 +44,6 @@ jobs:
             php-version: "8.2"
             symfony-deprecations-helper: "weak"
 
-          # Test with httplug 1.x clients
```
dbu, line 47: ah, httplug 1 was the problem in general. agree to drop that.

PR #434, opened 2023-08-25
Only support supported Symfony versions

ruudk:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | | Documentation   | if this is a new feature, link to pull request in https://github.com/php-http/documentation that adds relevant documentation
| License         | MIT

#### What's in this PR?

Only support Symfony versions that are not EOL.

https://symfony.com/releases


ostrolucky:
What are we gaining by this change? I wouldn't drop older versions only until we can at least drop some BC layers.

ruudk:
You mean I should remove BC compatibility for those older versions, if any, in order for this to be considered useful?

ostrolucky:
If there are some BC layers for 4.4 then removing them and dropping 4.4 support could make sense. But I don't think we have any BC layers, that's why there isn't any benefit for just changing requirements. Having 4.4 there doesn't hold us back and raising requirement would be artificial, since project has no issues running under 4.4.

dbu:
thanks for looking at this and the discussion. i agree with not dropping legacy only for the sake of dropping it. if there is anything hindering us, or if there is legacy support code e.g. in the DI system, then lets do it. but when there is additional code, its nice for people stuck on outdated symfony versions to still support them.

PR #435, opened 2023-08-31
chore: Add ClientInterface return typehint for php-http/httplug >= 2.4

demyan112rv:
| Q               | A
| --------------- | ---
| Deprecations?   | yes

#### What's in this PR?

Add ClientInterface return typehint for php-http/httplug >= 2.4

#### Why?

Return type of method ClientFactory::createClient() has typehint with deprecated interface Http\Client\HttpClient: 
since version 2.4, use Psr\Http\Client\ClientInterface instead

### DIFF
--- a/src/ClientFactory/ClientFactory.php
+++ b/src/ClientFactory/ClientFactory.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Http\HttplugBundle\ClientFactory;
 
 use Http\Client\HttpClient;
+use Psr\Http\Client\ClientInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


dbu (approved):
correct. actually with httplug 2, the httplug HttpClient interface extends the psr ClientInterface, but if somebody relies on still getting an old httplug instance the old annotation still makes sense.

dbu:
https://github.com/php-http/HttplugBundle/releases/tag/1.30.0

dbu:
thanks a lot for the contribution!

PR #438, opened 2023-10-31
feat: header authentication service configuration

### LINKED ISSUE #437: Header authentication service configuration
**Description**

The `Header` authentication service, introduced in https://github.com/php-http/message/commit/144d13ba024d0c597830636907144bc7c23f50dd, is present since v1.9.0 of its related package.

However, it is currently not possible to configure it through the bundle configuration: https://github.com/php-http/HttplugBundle/blob/57284fb032bcafc3e978a1d7e4a2b591045f7e15/src/DependencyInjection/Configuration.php#L645-L666

**Example**

```yaml
httplug:
  clients:
    sample:
      # ...
      plugins:
        - base_uri:
            uri: "%env(SAMPLE_API_URL)%"
        - authentication:
            header:
              type: header
              name: "ApiKey"
              token: "%env(SAMPLE_API_TOKEN)%"
```

soullivaneuh:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #437 
| Documentation   | https://github.com/php-http/documentation/pull/313
| License         | MIT

#### What's in this PR?

Exposes configuration for the header authentication plugin.

#### Why?

Allows the direct usage of this plugin through the bundle configuration.

#### Example Usage

``` yaml
  - authentication:
      header:
          type: header
          key: "ApiKey"
          value: "%env(SAMPLE_API_TOKEN)%"
```

#### Checklist

#### To Do

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 The change log describes what is "Added", "Removed", "Changed" or "Fixed" between each release.
 
+- Added configuration for the `header` authentication plugin (#437).
+
 # 1.30.1 - 2023-09-07
 
 - Added alias to allow autowiring the `AsyncHttpClient` interface (#436).
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "php-http/discovery": "^1.14",
         "php-http/httplug": "^2.0",
         "php-http/logger-plugin": "^1.1",
-        "php-http/message": "^1.4",
+        "php-http/message": "^1.9",
         "php-http/message-factory": "^1.0.2",
         "php-http/stopwatch-plugin": "^1.2",
         "psr/http-message": "^1.0 || ^2.0",
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -663,6 +663,10 @@ class Configuration implements ConfigurationInterface
                                 $this->validateAuthenticationType(['params'], $config, 'query_param');
 
                                 break;
+                            case 'header':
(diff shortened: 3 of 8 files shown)


soullivaneuh:
I am not sure of what I did wrong to make some tests crying. May you give me a hint @dbu?

soullivaneuh:
I found the reason. Apparently, it is because we are declaring the `name` scalar value that make the configuration process being confused as `name` is also used as a key attribute: https://github.com/pocivaneuh/HttplugBundle/blob/32caa9a2c95a4da393a0a91205ea904fe3c53960/src/DependencyInjection/Configuration.php#L640

I changed to `key` instead: 32caa9a

soullivaneuh:
All green! :+1: I also updated the documentation PR according to the configuration key name change.

dbu:
ah right, i think that is for xml where you can't have a custom xml element and we use name as identifier attribute. in yaml its the `my_auth:` part, and does not need an explicit name.

should we call the thing `header_name` instead of `key`? key is a bit generic and could be confusing in the context of security. it boils down to what header name to set in the request.

soullivaneuh:
Oh that case, `header_name` and `header_value` to make things consistent?

dbu:
yes, i like header_name and header_value. that makes it really clear and leaves no room for confusion (i hope :-) )

soullivaneuh:
@dbu keys named was changed in 04414b2.

Squashed and rebased, ready for review! :+1:

dbu (approved):
thanks!

PR #439, opened 2023-11-16
Psr17: replace usage of deprecated PHP-HTTP Factories with PSR-17 factories

### LINKED ISSUE #433: Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead
**PHP version**: x.y.z (hint: `php --version`) 8.1.9

**Description**

At the moment `HttplugBundle` depends on `php-http/message-factory`

https://github.com/php-http/HttplugBundle/blob/7c638710c3d4d4fe14de9e28e9a98c073a289cea/composer.json#L35

which [was deprecated](https://github.com/php-http/message-factory).

**How to reproduce**

**Possible Solution**

**Additional context**

glaubinix:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | yes
| Related tickets | fixes https://github.com/php-http/HttplugBundle/issues/433
| License         | MIT

#### What's in this PR?

Replaces usage of PHP-HTTP factories with PSR-17 factories

#### Why?

PHP-HTTP factories are deprecated

#### Checklist

#### To Do

- [x] Remove remaining references to the PHP-HTTP factories
- [x] Remove `php-http/message-factory` from composer.json

#### Questions
- [x] guzzle5-adapter doesn't support psr-17 and seems abandoned. Is it ok to drop support for it?


glaubinix:
Considering that the [guzzle5-adapter is basically abandoned](https://github.com/php-http/guzzle5-adapter/issues/37), I would suggest removing support for it here too.

dbu:
thanks a lot for this work! i agree we should do this, and do BC breaks. and i also agree with dropping guzzle 5 support. we could even also drop guzzle 6 i think and very old php versions. 

because of the major BC breaks, this seems the moment to bump the bundle to version 2. i just created the 2.x branch - can you please rebase and then add to the changelog the necessary informations of what changed / what users need to adjust in their configuration?

do you see other cleanups or changes that we should do before tagging 2.x? it would be the opportunity to get rid of legacy things and make necesary BC breaks.

glaubinix:
Rebased the PR onto 2.x and added changelog entries.

I see you already have a PR for marking classes as final and resolving some more deprecations. The other big deprecation I saw is the `HttpClient` type hint that needs to be replaced with `ClientInterface` and `AutoDiscoveryFactory` to use the new discovery factory. Happy to send a PR for that once this PR is done.

Optional changes should probably be based on how much easier it will make your life as a maintainer:
* guzzle6 (still works without much change, so up to you)
* drop support for old PHP versions, could then also drop Symfony 4.4 which isn't supported anymore either

dbu:
awesome, thanks a lot! i merged, as this is a nice package of changes.

if you are motivated to do it, i think raising to PHP 8 would be really neat so we can modernize the code as well. dropping support for legacy symfony versions would also be good - there is some ugly stuff in src/Collector/PluginClientFactoryListener.php for example to support legacy versions.

PR #442, opened 2023-11-21
Use ClientInterface instead of HttpClient

glaubinix:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | addresses one more item of https://github.com/php-http/HttplugBundle/pull/439#issuecomment-1820427669
| License         | MIT

#### What's in this PR?

Instead of using the deprecated `HttpClient` interface, this now uses `ClientInterface` everywhere

#### Why?

The old interface is deprecated

#### Checklist

#### To Do

- [ ] I am not sure if keeping around both `httplug.client.default` and `httplug.psr18_client.default` is OK or if this will cause any issues? If we can keep them around, then it might be easier for people to update the library without running into issues.


tests/Unit/Collector/ProfileCustomClientTest.php
```diff
@@ -240,6 +240,6 @@ public function testOnRejected(): void
     }
 }
 
-interface ClientInterface extends HttpClient, HttpAsyncClient
+interface CustomClientInterface extends ClientInterface, HttpAsyncClient
```
dbu:
```suggestion
interface CombinedClientInterface extends ClientInterface, HttpAsyncClient
```

dbu (approved):
thanks a lot!

i have one small nitpick on the naming in the test.

dbu (approved):
thanks!

we should update a few things on https://docs.php-http.org/en/latest/integrations/symfony-bundle.html and https://docs.php-http.org/en/latest/integrations/symfony-full-configuration.html (sources are at https://github.com/php-http/documentation/tree/main/integrations)

glaubinix:
Yes, that was on my list. How do you generally handle cases that are different in versions, e.g. removed config keys?

dbu:
generally it would be best if we had a separate documentation for each major version, but the doc is for all the various components together so no meaningful versioning is possible.

i would just remove obsolete things, and have `..versionadded:: 2.0` notes for new things (see for example components/client-common.rst)

mostly people will read the doc when setting things up. for BC breaks the changelog of the corresponding repository is the right place.

PR #446, opened 2024-01-12
adjust composer install steps

dbu:


ostrolucky:
This solution would be better https://github.com/doctrine/DoctrineBundle/commit/42f5834825eaa967af833722c336d5dd05cbd4f3

Basically, `shivammathur/setup-php` can install flex for you

dbu:
oh nice! need to change that in several repositories i maintain :-)

PR #451, opened 2024-04-24
feat: add factory for NoPrivateNetworkHttpClient

### LINKED ISSUE #450: Add Factory for NoPrivateNetworkHttpClient
**Description**
A new factory for the `NoPrivateNetworkHttpClient` (SF 5.1 and up) which blocks calls to private network addresses would be nice to have.

**Example**
I'm currently implementing outgoing webhooks where the user can set the address, and to stop SSRF attacks a factory like that would be neat 

**Additional context**
I'll make a PR for it soon :)

KDederichs:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #450 
| License         | MIT

#### What's in this PR?

Adds a factory for Symfony's `NoPrivateNetworkHttpClient`

#### Why?

While you can just wrap it yourself, it gets annoying when you're trying to inject other non Symfony clients in unit tests (mock client)

#### Example Usage

``` yaml
        webhook:
            http_methods_client: false
            factory: 'httplug.factory.symfony_no_private_network'
```

#### Checklist

#### To Do

- [ ] Please tell me if you're actually OK with something like this, if so I'll add it to the docs and changelog

### DIFF
--- a/src/ClientFactory/SymfonyNoPrivateNetworkFactory.php
+++ b/src/ClientFactory/SymfonyNoPrivateNetworkFactory.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Http\HttplugBundle\ClientFactory;
+
+use Psr\Http\Message\ResponseFactoryInterface;
+use Psr\Http\Message\StreamFactoryInterface;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Component\HttpClient\HttplugClient;
+use Symfony\Component\HttpClient\NoPrivateNetworkHttpClient;
+
+class SymfonyNoPrivateNetworkFactory
+{
+    /**
+     * @var ResponseFactoryInterface
+     */
+    private $responseFactory;
+
+    /**
+     * @var StreamFactoryInterface
+     */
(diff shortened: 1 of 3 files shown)


ostrolucky:
This should be a plugin instead IMO. We can't add factory for every "client" that symfony/http-client has, because they are using client implementations as a way of extending the functionality, for which httplug has concept of plugins instead

KDederichs:
Sure, plugin also sounds good.
I'll write one for myself then and if you think there's demand for that I'll PR it into `php-http/client-common`

dbu:
thanks for wanting to contribute. i agree that this would fit better in a plugin than as a specific client.

if you write a plugin that replicates the functionality, that would be interesting to have in php-http/client-common. it can not depend on the symfony client though.

if you write a plugin that leverages the symfony code, you could publish it in a separate repository and we can mention it in the plugin list.

btw, note that httplug bundle also allows you to specify a custom client service as http client if you want to instantiate a specific client.

PR #456, opened 2024-06-11
Introduce Phpstan

Prometee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| Documentation   | 
| License         | MIT

#### What's in this PR?

Add PHPStan analyse.

#### Why?

Avoid issue with non typed variables.

#### To Do

- [x] Add PHPStan baseline.
- [x] Add PHPStan to the static analysis.
- [x] Try to fix some of the baseline errors.


composer.json
```diff
@@ -43,12 +43,11 @@
     },
     "conflict": {
         "php-http/guzzle6-adapter": "<1.1",
-        "php-http/cache-plugin": "<1.7.0",
+        "php-http/cache-plugin": "<1.7",
         "php-http/curl-client": "<2.0",
         "php-http/socket-client": "<2.0",
         "kriswallsmith/buzz": "<0.17",
-        "php-http/react-adapter": "<3.0",
-        "php-http/cache-plugin": "<1.7"
```
dbu, line 51: are you sure we can drop this?

.github/workflows/continuous-integration.yml
```diff
@@ -104,5 +104,8 @@ jobs:
           dependency-versions: "${{ matrix.dependency-versions }}"
           composer-options: "${{ matrix.composer-options }}"
 
+      - name: "Run PHPStan"
+        run: "vendor/bin/phpstan analyse --no-progress"
```
dbu: i would prefer a separate job in static-analysis.yml, similar to https://github.com/php-http/httplug/blob/2.x/.github/workflows/static.yml

for local development, people can install phpstan.phar into their system.
and we probably need to enable the symfony extension of phpstan to help avoid some of the complaints.

phpstan.neon.dist
```diff
@@ -0,0 +1,9 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    reportUnmatchedIgnoredErrors: false
+
+    level: max
```
dbu: i'd prefer to start with level 1 to see the important things. if we start with max right away, the baseline is huge and priorities are not clear.

dbu:
thanks a lot for picking this up!

composer.json
Prometee: Two lines with the same conflict were here :

```
"php-http/cache-plugin": "<1.7.0",
"php-http/cache-plugin": "<1.7"
```

I simply merged them into one.

Then yes I'm sure 😉

dbu: oops, haha. yeah ok, you convinced me :joy:

.github/workflows/static-analysis.yml
```diff
@@ -16,6 +16,12 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v3"
+      - name: PHPStan
+        uses: "docker://oskarstark/phpstan-ga"
+        env:
+          REQUIRE_DEV: true
+        with:
+          args: analyze --no-progress
```
dbu: this makes it part of the coding-standards job.

if we define it as separate job, it runs in parallel with the cs fixer and also fails separately so its immediatly clear which things are good and which not. same as in https://github.com/php-http/httplug/blob/2.x/.github/workflows/static.yml (we could rename the `coding-standards` job to `php-cs-fixer` to make it fully clear which is which.

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Call to static method create\\(\\) on an unknown class Symfony\\\\Component\\\\HttpClient\\\\HttpClient\\.$#"
+			count: 1
+			path: src/ClientFactory/SymfonyFactory.php
```
dbu, line 6: i guess this means we don't test the factory, otherwise we would depend on the symfony http client.

we should probably add that, but in a separate merge request. and its no blocker for releasing 2.0

.github/workflows/static-analysis.yml
Prometee: Totally agree!

phpstan-baseline.neon
```diff
@@ -0,0 +1,31 @@
...
+
+		-
+			message: "#^Class Http\\\\Client\\\\Plugin\\\\Vcr\\\\NamingStrategy\\\\NamingStrategyInterface not found\\.$#"
+			count: 1
+			path: src/DependencyInjection/Configuration.php
```
dbu, line 11: does the code properly handle a missing vcr plugin? if it does its okay i guess.

phpstan.neon.dist
```diff
@@ -0,0 +1,9 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    reportUnmatchedIgnoredErrors: false
```
dbu: i would not disable this report, otherwise the baseline could accumulate unneeded exceptions over time.

Prometee: I normally test PHPStan under different PHP and Sf version and keep a `phpstan.neon` file on my filesystem (not comited) to get check if ignored error are still needed, but since the CI is going to test it with only one PHP version and a single related Symfony version it's useless here.

src/DependencyInjection/HttplugExtension.php
```diff
@@ -549,7 +549,10 @@ private function configureAutoDiscoveryClients(ContainerBuilder $container, arra
      */
     public function getConfiguration(array $config, ContainerBuilder $container): ?ConfigurationInterface
     {
-        return new Configuration($container->getParameter('kernel.debug'));
+        /** @var bool $debug */
```
dbu: can you try configuring the phpstan symfony extension as per https://github.com/phpstan/phpstan-symfony?tab=readme-ov-file#configuration ? it should make phpstan aware of the symfony specific things and know the types of container->get and getParameter.

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  push:
+    branches:
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.x'
+  pull_request:
+
+jobs:
+  coding-standards:
```
dbu:
```suggestion
  phpstan:
```

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,24 @@
...
+    name: "Coding Standards"
```
dbu:
```suggestion
    name: "PHPStan"
```

Prometee:
@dbu what about raising the PHP version up for `2.0`, I propose to start with `^8.1` and keep only 2 version of Symfony : `^6.4|^7.1`. It's amazing that this bundle can support four versions but it's normally not possible for Symfony to be able to ensure such a compatibility across versions. WDYT ?

dbu:
> @dbu what about raising the PHP version up for `2.0`, I propose to start with `^8.1` and keep only 2 version of Symfony : `^6.4|^7.1`. It's amazing that this bundle can support four versions but it's normally not possible for Symfony to be able to ensure such a compatibility across versions. WDYT ?

yes, good point. its good to support the legacy versions for legacy projects, but in the new major, we should raise to only support the maintained versions of symfony. we can potentially also drop some BC hacks when we drop old symfony versions.

(but lets not do that in the same pull request ;-) )

src/DependencyInjection/HttplugExtension.php
Prometee: Tricky since no `var` folder is not generated (only PHPUnit will create those files). I added what I can to the `phpstan.neon.dist`.

composer.json
```diff
@@ -58,6 +57,7 @@
         "php-http/cache-plugin": "^1.7",
         "php-http/mock-client": "^1.2",
         "php-http/promise": "^1.0",
+        "phpstan/phpstan-symfony": "^1.2",
```
dbu: the phpstan-symfony extension is already provided by the github-action docker image.

Prometee: You are right, I did not read the Dockerfile : https://github.com/OskarStark/phpstan-ga/blob/master/Dockerfile#L24.

src/DependencyInjection/HttplugExtension.php
Prometee: Also I can't include files like `vendor/phpstan/*` since they will be inside the global composer folder : `~/.composer/vendor/phpstan*`

phpstan.neon.dist
```diff
@@ -0,0 +1,10 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    scanFiles:
+        - vendor/symfony/dependency-injection/Loader/Configurator/ContainerConfigurator.php
```
dbu, line 8: oh, neat! so this is enough to allow phpstan to see the type of the kernel.debug flag.

Prometee: No unfortunately, it needs the var cache folder to be generated.

phpstan-baseline.neon
dbu: we could `composer require` the vcr plugin in a first step of the phpstan job, to have it validate correctly. the php tests would discover if we messed up with validation.

some nice to haves would be to have the configuration class error when configuring without vcr without having the plugin installed, and https://github.com/php-http/HttplugBundle/blob/b7ad2c1951edec36ef0bd9897e309b0038c59a54/src/DependencyInjection/HttplugExtension.php should throw an exception at the beginning of the method if vcr is not available.

but probably this still not enough to make phpstan happy, so composer require step for this could be added. that means that phpstan locally would complain about those though :thinking: 

we can also merge the current state and continue in separate pull requests.

phpstan.neon.dist
dbu: but you don't have the `@var` phpdoc anymore, and the build is green...

src/Collector/PluginClientFactory.php
```diff
@@ -47,7 +47,7 @@ public function __construct(Collector $collector, Formatter $formatter, Stopwatc
      * @param Plugin[]                        $plugins
      * @param array                           $options {
      *
-     *     @var string $client_name to give client a name which may be used when displaying client information like in
+     *     @type string $client_name to give client a name which may be used when displaying client information like in
```
dbu: cs fixer vs phpstan: fight round 1 :facepalm: 

the phpstan way to do this would be an array shape. 

`@param array{string: client_name} $options`

and explain the meaning of the options as a list below the `@param` annotations.

phpstan.neon.dist
Prometee: yes it's because I droped the commit because it doesn't concern the current PHPStan level, it will reappear when I'll reach level 3 or 4.

src/Collector/PluginClientFactory.php
Prometee: Thanks you, I was about to do it 😉

phpstan.neon.dist
dbu: ah, i see. 

i'd say lets get level 2 working and then merge what we have.

i can give it a try to have the container xml built without running the whole test suite.

Prometee: Build is green now, then I guess you can make a final review and do what you want with this PR 😄

dbu (approved):
awesome, thanks a lot for the contribution!

dbu:
created follow-up tickets #458 #457 #459

PR #460, opened 2024-06-14
add ThrottlePlugin

Big-Shark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

#### What's in this PR?

Add config settings for ThrottlePlugin

#### Why?

Adding the plugin without using configurations causes inconvenience.

### Example

        default:
            factory: 'httplug.factory.curl'
            config:
                CURLOPT_TIMEOUT: 10
            plugins:
                - throttle:
                      name: default
                      key: default
                      tokens: 1
                      max_time: 1

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -630,8 +630,9 @@ class Configuration implements ConfigurationInterface
             ->addDefaultsIfNotSet()
             ->children()
                 ->scalarNode('name')->end()
-                ->scalarNode('tokens')->defaultValue(1)->end()
-                ->scalarNode('max_time')->defaultNull()->end()
+                ->scalarNode('key')->defaultNull()->end()
+                ->integerNode('tokens')->defaultValue(1)->end()
+                ->floatNode('max_time')->defaultNull()->end()
             ->end()
         ->end();
         // End throttle plugin
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -624,6 +624,17 @@ class Configuration implements ConfigurationInterface
             ->end()
         ->end();
         // End error plugin
+
+        $throttle =  $children->arrayNode('throttle')
+            ->canBeEnabled()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->scalarNode('name')->end()
+                ->scalarNode('tokens')->defaultValue(1)->end()
+                ->scalarNode('max_time')->defaultNull()->end()
+            ->end()
+        ->end();
+        // End throttle plugin
     }
 
     /**
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -625,7 +625,7 @@ class Configuration implements ConfigurationInterface
         ->end();
         // End error plugin
 
-        $throttle =  $children->arrayNode('throttle')
+        $throttle = $children->arrayNode('throttle')
             ->canBeEnabled()
             ->addDefaultsIfNotSet()
             ->children()
--- a/src/DependencyInjection/HttplugExtension.php
+++ b/src/DependencyInjection/HttplugExtension.php
@@ -10,6 +10,7 @@ use Http\Client\Common\FlexibleHttpClient;
 use Http\Client\Common\HttpMethodsClient;
 use Http\Client\Common\HttpMethodsClientInterface;
 use Http\Client\Common\Plugin\AuthenticationPlugin;
+use Http\Client\Common\Plugin\ThrottlePlugin;
 use Http\Client\Common\PluginClient;
 use Http\Client\Common\PluginClientFactory;
 use Http\Client\HttpAsyncClient;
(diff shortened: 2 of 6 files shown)


Big-Shark:
@dbu Please, can you check my PR. If you have some question, just ask me.

src/DependencyInjection/HttplugExtension.php
```diff
@@ -292,6 +294,28 @@ private function configurePluginByName($name, Definition $definition, array $con
 
                 break;
 
+            case 'throttle':
+                if (!\interface_exists(LimiterInterface::class)) {
```
dbu: this check is redundant. the throttle-plugin repository depends on symfony/rate-limiter. checking for the ThrottlePlugin is enough.

dbu (commented):
thanks, nice work! what happens if somebody would install the old version of throttle plugin with these changes? with the named parameters, we would get an error from symfony about the new parameters, i think? if that is correct, we should add a `conflict` in composer.json to avoid installing this with a too old version.

can you please  add to the HttplugExtensionTest::testClientPlugins to enable the throttle plugin and verify the service is created (the test is generic, just add the config and the expected service name)

to keep the documentation up to date, can you please also add to https://github.com/php-http/documentation/blob/main/integrations/symfony-bundle.rst#list-of-services and https://github.com/php-http/documentation/blob/main/integrations/symfony-full-configuration.rst ?

Big-Shark:
@dbu 
> thanks, nice work! what happens if somebody would install the old version of throttle plugin with these changes? with the named parameters, we would get an error from symfony about the new parameters, i think? if that is correct, we should add a conflict in composer.json to avoid installing this with a too old version.

yes, you are right, thanks
image

> can you please add to the HttplugExtensionTest::testClientPlugins to enable the throttle plugin and verify the service is created (the test is generic, just add the config and the expected service name)

I tried, but minimal supported version for bundle, it's 7.3, but throttle plugin works only with 7.4 and upper, and I can not run this test in 7.3. I think we can add this only in v2, if you want.
https://github.com/php-http/HttplugBundle/actions/runs/9521365832/job/26248616912

src/DependencyInjection/HttplugExtension.php
Big-Shark: Fixed

ostrolucky:
I think you can make this work if you use positional arguments instead of named

Big-Shark:
@ostrolucky Possible, but is it necessary, because there are no backwards compatible changes?

ostrolucky:
Well I believe it will be backwards compatible with positional arguments instead of named ones.

Big-Shark:
@ostrolucky It will work, but the options won't work as expected, which can also be confusing.

dbu (approved):
regarding the test with the plugin: i see. then lets add that in the 2.x branch. i will merge this now and then merge 1.x to 2.x

Big-Shark:
@dbu thanks, please, create new release

dbu:
on it, on it ;-) 

https://github.com/php-http/HttplugBundle/releases/tag/1.34.0

and merging to 2.x in https://github.com/php-http/HttplugBundle/pull/461 (i added the test right away)

src/DependencyInjection/Configuration.php
```diff
@@ -624,6 +624,18 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
             ->end()
         ->end();
         // End error plugin
+
+        $throttle = $children->arrayNode('throttle')
+            ->canBeEnabled()
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->scalarNode('name')->end()
```
dbu, line 632: looking at the test failure in https://github.com/php-http/HttplugBundle/pull/461: what is the name here? and what is key?

the name is not made required, but the configuration class does not handle a missing name field.

src/DependencyInjection/HttplugExtension.php
```diff
@@ -292,6 +294,24 @@ private function configurePluginByName($name, Definition $definition, array $con
...
 
+            case 'throttle':
+                if (!\class_exists(ThrottlePlugin::class)) {
+                    throw new InvalidConfigurationException('You need to require the Throttle Plugin to be able to use it: "composer require php-http/throttle-plugin".');
+                }
+
+                $key = $config['name'] ? '.'.$config['name'] : '';
+                $container
+                    ->register($serviceId.$key, LimiterInterface::class)
+                    ->setFactory([new Reference('limiter.'.$config['name']), 'create'])
```
dbu, line 305: this does not check if the name is set or not. and if the name is empty, it will use `limiter.`, is that the intention?

Big-Shark: @dbu No, it was my mistake, I assume that if the default value is not specified, then it is required by default

src/DependencyInjection/Configuration.php
Big-Shark: @dbu You need to set name, name this is required option

dbu: then we are missing `isRequired()` 

can you please also add an `info()` as well with an explanation what this name is? the name of what thing do we specify here?

src/DependencyInjection/HttplugExtension.php
```diff
@@ -292,6 +294,24 @@ private function configurePluginByName($name, Definition $definition, array $con
 
                 break;
 
+            case 'throttle':
+                if (!\class_exists(ThrottlePlugin::class)) {
+                    throw new InvalidConfigurationException('You need to require the Throttle Plugin to be able to use it: "composer require php-http/throttle-plugin".');
+                }
+
+                $key = $config['name'] ? '.'.$config['name'] : '';
```
dbu, line 302: if name is required, this check is for the special case where name is set to `''`? below it would still do strange things when name is `''`

dbu: and i think symfony configuration will not allow `''` on something that is set to be required

src/DependencyInjection/Configuration.php
Big-Shark: Done, check #462

src/DependencyInjection/HttplugExtension.php
Big-Shark: Done, check #462

PR #462, opened 2024-06-17
Throttle plugin small fixes

Big-Shark:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #460 
| License         | MIT

#### What's in this PR?

Did code more clearly

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -630,7 +630,7 @@ class Configuration implements ConfigurationInterface
             ->addDefaultsIfNotSet()
             ->children()
                 ->scalarNode('name')->isRequired()->info('Rate limiter configuration name from rate_limiter.yaml')->end()
-                ->scalarNode('key')->defaultNull()->info('A unique key for using one rate limiter name for different clients')->end()
+                ->scalarNode('key')->defaultNull()->info('Key to avoid sharing this rate limiter with other clients or other services. You can use the name of the client for example.')->end()
                 ->integerNode('tokens')->defaultValue(1)->info('How many tokens spending per request')->end()
                 ->floatNode('max_time')->defaultNull()->info('Maximum accepted waiting time in seconds')->end()
             ->end()
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -629,7 +629,7 @@ class Configuration implements ConfigurationInterface
             ->canBeEnabled()
             ->addDefaultsIfNotSet()
             ->children()
-                ->scalarNode('name')->isRequired()->info('Rate limiter configuration name from rate_limiter.yaml')->end()
+                ->scalarNode('name')->isRequired()->info('Rate limiter factory service name, e.g. limiter.http_client for a rate limiter called http_client')->end()
                 ->scalarNode('key')->defaultNull()->info('Key to avoid sharing this rate limiter with other clients or other services. You can use the name of the client for example.')->end()
                 ->integerNode('tokens')->defaultValue(1)->info('How many tokens spending per request')->end()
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -629,10 +629,10 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
             ->canBeEnabled()
             ->addDefaultsIfNotSet()
             ->children()
-                ->scalarNode('name')->end()
-                ->scalarNode('key')->defaultNull()->end()
-                ->integerNode('tokens')->defaultValue(1)->end()
-                ->floatNode('max_time')->defaultNull()->end()
+                ->scalarNode('name')->isRequired()->info('Rate limiter configuration name from rate_limiter.yaml')->end()
+                ->scalarNode('key')->defaultNull()->info('A unique key for using one rate limiter name for different clients')->end()
```
dbu: i just read up on rate limiter, now i see what it does. we could also default this to the client name instead of null, not sure which way is more expected. like this we err on the side of limiting too much if the user does not understand the documentation, so i think its the safe default.

i suggest adding a bit more to the info text:

```suggestion
                ->scalarNode('key')->defaultNull()->info('Key to avoid sharing this rate limiter with other clients or other services. You can use the name of the client for example.')->end()
```

src/DependencyInjection/Configuration.php
```diff
@@ -629,10 +629,10 @@ private function addSharedPluginNodes(ArrayNodeDefinition $pluginNode, $disableA
             ->canBeEnabled()
             ->addDefaultsIfNotSet()
             ->children()
-                ->scalarNode('name')->end()
-                ->scalarNode('key')->defaultNull()->end()
-                ->integerNode('tokens')->defaultValue(1)->end()
-                ->floatNode('max_time')->defaultNull()->end()
+                ->scalarNode('name')->isRequired()->info('Rate limiter configuration name from rate_limiter.yaml')->end()
```
dbu: this will tricky when a user configures their own service for a rate limiter factory and does not call it `limiter.{name}`. 

in other places, we ask for a full service name for things. i think i would prefer to do the same here:
```suggestion
                ->scalarNode('name')->isRequired()->info('Rate limiter factory service name, e.g. limiter.http_client for a rate limiter called http_client')->end()
```

wdyt? if we change it here, we also need to adjust the extension class.

and now that i understand what it is, it should be called more specific `limiter_name`. or while we do change it to be the service,  `limiter_service`.

Big-Shark: @dbu Symfony automatic create factory in container, I just have this config
image

I think using just "name" more understandable than whole name in container, like "limiter.http_client"

src/DependencyInjection/Configuration.php
Big-Shark: > client name instead of null

You mean key instead of null?
I think 99% does not need it, and "null" it's default value in RateLimiterFactory. We can use something else, but we need it only if we have the same rate limiters for different services. And we do not have plan for creating different rate limiters without any reason.

dbu: what i meant is we could use the name of the http client where the plugin is attached as key for the rate limiter, so the limit applies per client instead of globally for the limiter.

but that was thinking out loud, lets keep as is and only adjust the documentation with my suggestion if you agree

src/DependencyInjection/Configuration.php
dbu: yes, when one configures in framework bundle, that creates the services with this naming pattern.

the thing is that a user might configure their own service rather than use the framework configuration, e.g. when they have a custom implementation of the rate limiter. and for that i think its more consistent with the rest of the bundle to have the full service name.

dbu:
wrapped up in #465, thanks!

PR #464, opened 2024-09-01
throttle plugin requires a name

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | (no) if the name option for the throttle plugin was not specified, 1.34.0 did not report invalid configuration, but the extension class was broken and if php ignored the undefined array key, the plugin can not be instantiated because the service does not exist.
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Throttle plugin references a symfony/rate-limiter service with the `name`. This is not optional.

### DIFF
--- a/src/DependencyInjection/HttplugExtension.php
+++ b/src/DependencyInjection/HttplugExtension.php
@@ -299,14 +299,13 @@ class HttplugExtension extends Extension
                     throw new InvalidConfigurationException('You need to require the Throttle Plugin to be able to use it: "composer require php-http/throttle-plugin".');
                 }
 
-                $key = $config['name'] ? '.'.$config['name'] : '';
                 $container
-                    ->register($serviceId.$key, LimiterInterface::class)
+                    ->register($serviceId.$config['name'], LimiterInterface::class)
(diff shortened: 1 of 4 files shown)


src/DependencyInjection/HttplugExtension.php
```diff
@@ -299,14 +299,13 @@ private function configurePluginByName($name, Definition $definition, array $con
                     throw new InvalidConfigurationException('You need to require the Throttle Plugin to be able to use it: "composer require php-http/throttle-plugin".');
                 }
 
-                $key = $config['name'] ? '.'.$config['name'] : '';
                 $container
-                    ->register($serviceId.$key, LimiterInterface::class)
+                    ->register($serviceId.$config['name'], LimiterInterface::class)
                     ->setFactory([new Reference('limiter.'.$config['name']), 'create'])
```
dbu, line 304: this is asserting that the name must exist.

and if we would use $key, we would set a reference to `limiter.` which would not be found.

PR #472, opened 2024-09-05
remove all deprecated code

dbu:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | -
| Documentation   | -
| License         | MIT

#### What's in this PR?

Remove things that are marked as deprecated.

#### Why?

Cleanup for major version.

#### Checklist


src/Collector/Twig/HttpMessageMarkupExtension.php
```diff
@@ -34,40 +34,14 @@ public function __construct(?ClonerInterface $cloner = null, ?DataDumperInterfac
...
-        $headers = preg_replace("|\n(.*?): |si", "\n<b>$1</b>: ", $parts[0]);
-
-        return sprintf("%s\n\n<div class='httplug-http-body httplug-hidden'>%s</div>", $headers, $parts[1]);
-    }
-
     public function markupBody(string $body): ?string
     {
-        if (in_array(substr($body, 0, 1), ['{', '['], true)) {
-            $json = json_decode($body, true);
+        if ('' !== $body && in_array($body[0], ['{', '['], true)) {
```
ostrolucky, line 41: Since you are changing this within extra scope, my suggestion here would be to use `str_starts_with` instead

src/Collector/Twig/HttpMessageMarkupExtension.php
```diff
@@ -34,40 +34,14 @@ public function __construct(?ClonerInterface $cloner = null, ?DataDumperInterfac
...
+            $json = json_decode($body, true, 512, JSON_THROW_ON_ERROR);
```
ostrolucky: This is behavioural change. Check below accounted for this, but now exception will be thrown here and check below won't trigger.

ostrolucky (changes requested):

src/Collector/Twig/HttpMessageMarkupExtension.php
dbu: uh, thanks for the heads up. i just accpted intellij suggesting to improve the code... added the catch now to ignore the exception.

src/Collector/Twig/HttpMessageMarkupExtension.php
dbu: str_starts_with does not accept an array. i am not changing the scope, i only verify that the body is not empty, because then we would try to access an array out of its bounds. (which probably has been prevented somewhere else, or we would already have seen issues). but in the interest of defensive code i think we should not risk accessing the string outside its bounds.

ostrolucky (approved):

PR #481, opened 2025-12-06
fix: update routing configuration to use PHP files for WebProfilerBundle

ip512:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
relevant documentation
| License         | MIT

#### What's in this PR?

Use php config files from web-profiler-bundle instead of deprecated XML ones

#### Why?

When running tests locally, I get the following error :

`Cannot load resource "@WebProfilerBundle/Resources/config/routing/wdt.xml". Make sure the "WebProfilerBundle" bundle is correctly registered and loaded in the application kernel class. If the bundle is registered, make sure the bundle path "@WebProfilerBundle/Resources/config/routing/wdt.xml" is not empty.`

Although XML files exist, they can't be loaded in version 7.4 of the bundle

### DIFF
--- a/tests/Resources/app/AppKernel.php
+++ b/tests/Resources/app/AppKernel.php
@@ -66,8 +66,9 @@ class AppKernel extends Kernel
         $collection->add('/', new Route('/', ['_controller' => 'kernel::indexAction']));
 
         $routes = new RoutingConfigurator($collection, $kernelLoader, $file, $file);
-        $routes->import('@WebProfilerBundle/Resources/config/routing/wdt.xml')->prefix('_wdt');
-        $routes->import('@WebProfilerBundle/Resources/config/routing/profiler.xml')->prefix('_profiler');
+        $extension = self::MAJOR_VERSION >= 7 ? 'php' : 'xml';
+        $routes->import('@WebProfilerBundle/Resources/config/routing/wdt.'.$extension)->prefix('_wdt');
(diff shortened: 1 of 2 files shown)


tests/Resources/app/AppKernel.php
```diff
@@ -66,8 +66,9 @@ public function loadRoutes(LoaderInterface $loader): RouteCollection
         $collection->add('/', new Route('/', ['_controller' => 'kernel::indexAction']));
 
         $routes = new RoutingConfigurator($collection, $kernelLoader, $file, $file);
-        $routes->import('@WebProfilerBundle/Resources/config/routing/wdt.xml')->prefix('_wdt');
-        $routes->import('@WebProfilerBundle/Resources/config/routing/profiler.xml')->prefix('_profiler');
+        $extension = self::MAJOR_VERSION >= 7 ? 'php' : 'xml';
```
dbu, line 70: strictly speaking, this is about the version of the WebProfilerBundle which might diverge from the HttpKernel component which provides the symfony version. but this is only for tests, so imho good enough as is.

```suggestion
        // TODO hardcode xml when we support only Symfony 7.3 or higher.
        $extension = self::MAJOR_VERSION >= 7 ? 'php' : 'xml';
```

dbu (approved):
i think this happens because we end up with a mix of symfony 7 and 8 components because we don't always specify the symfony version for flex in the ci

as we want to support symfony 8 anyways, this is a welcome fix, thanks!

